PR #68, opened 2021-04-27
add main module flag to emcc build

sinkingsugar:


sinkingsugar:
Hmm first of all rust side is not working, we probably need to manually link them...
Second pthread probably does not work, limit in emscripten...

sinkingsugar:
https://doc.rust-lang.org/rustc/codegen-options/index.html might have to deal with those options, but pthread is still a issue

PR #73, opened 2021-06-14
Add rust formatting config

Kryptos-FR:
Format existing code using `cargo fmt` according to that config.

### DIFF
--- a/rust/rustfmt.toml
+++ b/rust/rustfmt.toml
@@ -1,0 +1,2 @@
+newline_style = "Native"
+tab_spaces = 2
--- a/rust/src/block.rs
+++ b/rust/src/block.rs
@@ -1,7 +1,7 @@
-use crate::chainblocksc::CBOptionalString;
 use crate::chainblocksc::CBContext;
 use crate::chainblocksc::CBExposedTypesInfo;
 use crate::chainblocksc::CBInstanceData;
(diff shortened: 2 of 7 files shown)


sinkingsugar:
Nice,
How does that work? do we have to run manually `cargo fmt`? cos I generally do it via vscode already usually (manually)

Also btw, you should remove `rust/src/chainblocksc.rs`
That file needs some mention on the readme maybe. Basically ti should not be there but it's there just in order to let the Windows 32bit build pass, cos I push that specific version of the bindings, otherwise they cannot be generated in CI.

rust/rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+newline_style = "Native"
```
sinkingsugar, line 1: what are the other styles?

sinkingsugar (changes requested):

rust/rustfmt.toml
Kryptos-FR: Possible values: "Auto", "Native", "Unix", "Windows"

Reference: https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#newline_style

Kryptos-FR:
> Also btw, you should remove `rust/src/chainblocksc.rs`

Ok.

sinkingsugar (approved):

PR #74, opened 2021-06-19
Docs for a few basic blocs

Kryptos-FR:
Add documentation for a few basic blocks: `Cond`, `If`, `When`, `WhenNot`.

The inline code currently doesn't have syntax highlighting. There are apparently some plugins available but their doc is cryptic and I'm not sure how to add them (nor activate them).

In the future, it should also include visuals of our UX as an alternative to the clojure syntax. Not sure how that will be presented: either on the same documentation page or separated (in which case this doc will become developer-oriented only).

Worth noting that given the template used here we could envision an automatic generation from the source given proper annotations, at least partially.


sinkingsugar (changes requested):
Btw we can auto generate most of this,
Consider that inside the code Parameters are also all documented.

sinkingsugar:
Have a look at https://github.com/sinkingsugar/chainblocks/blob/devel/src/tests/infos.clj , there is already reflection in the script language.

sinkingsugar (approved):

PR #75, opened 2021-06-27
Blocks documentation generated from code

Kryptos-FR:
Add a script (`src/tests/infos_docs.edn`) to extract the documentation from the block's info and generate the corresponding markdown files.

Not perfect especially around the list of types but good enough to at least serve as a draft that can be edited if needed.
For instance, the `Definition` and `Examples` sections are mostly empty.


docs/docs/blocks/!/All.md
```diff
@@ -0,0 +1,36 @@
+# All
+
+```clojure
+(All
+  :Value [(Any)]
+)
```
sinkingsugar, line 6: Is there any way to print those in a clojure like formatting way?
```
(All
  :X [(None)]
  :Value [(Any)])

Comment (sinkingsugar) [reply docs/docs/blocks/!/All.md:6]: I know might be unnatural but it's kinda standard for clojure/lisp users

File: docs/docs/blocks/!/AnyMore.md [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+)
+```
+
+## Definition
+
+
+## Parameters
+| Name | Type | Default | Description |
+|------|------|---------|-------------|
+| Value | `[(Any)]` | `None` | The value to test against for equality. |
```
sinkingsugar, line 15: I think should create a issue to track the `Type` further parsing. But for now we can skip it.

docs/docs/blocks/!/ChainRunner.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+## Definition
+
+
+## Parameters
+| Name | Type | Default | Description |
+|------|------|---------|-------------|
+| Chain | `[(ContextVar [(Chain)])]` | `None` | The chain variable to compose and run. |
+| Mode | `[(Enum)]` | `Enum: 0 vendor: 0x73696e6b type: 0x72756e43` | The way to run the chain. Inline: will run the sub chain inline within the root chain, a pause in the child chain will pause the root too; Detached: will run the chain separately in the same node, a pause in this chain will not pause the root; Stepped: the chain will run as a child, the root will tick the chain every activation of this block and so a child pause won't pause the root. |
```
sinkingsugar, line 17: Hmm, maybe we should print those `vendor` and `type` as chars as they are multichar `'frag'` for example.

sinkingsugar: What do you think? I can help with the implementation

docs/docs/blocks/!/Ref.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Output
+| Type | Description |
+|------|-------------|
+| `[(Any)]` |  |
+
+
+## Examples
+
+```clojure
```
sinkingsugar, line 38: Btw I noticed mkdocs has no clojure syntax highlight, we should investigate how to include that (pretty sure it's probably a flag or some modules to add)

docs/docs/blocks/!/TryMany.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Definition
+
+
+## Parameters
+| Name | Type | Default | Description |
+|------|------|---------|-------------|
+| Chain | `[(Chain) (String) (None) (ContextVar [(Chain)])]` | `None` | The chain to spawn and try to run many times concurrently. |
+| Policy | `[(Enum)]` | `Enum: 1 vendor: 0x73696e6b type: 0x7472794d` | The execution policy in terms of chains success. |
```
sinkingsugar, line 19: Another thing about Enum.. lisp side we should actually instead process them and print the real enum that is behind it.
I don't do it in the stringify cos it's expensive but maybe we should somehow.

docs/docs/blocks/If.md
```diff
@@ -1,39 +0,0 @@
-# If
```
sinkingsugar, line 1: How come this is gone?

src/tests/infos_docs.edn
```diff
@@ -0,0 +1,142 @@
+; run with ./build/cbl.exe ./src/tests/infos_docs.edn
+(defnode main)
```
sinkingsugar, line 2: I would use clojure like formatting 😄

sinkingsugar (changes requested):

docs/docs/blocks/!/Ref.md
Kryptos-FR: There are plugins to add syntax coloring. It just that the documentation of such plugins is very hard to understand: they expect you to know how mkdocs works internally and provide no easy step to integrate it.

src/tests/infos_docs.edn
Kryptos-FR: What do you mean?

docs/docs/blocks/If.md
Kryptos-FR: Looks like a mistake

docs/docs/blocks/!/All.md
Kryptos-FR: What is a "clojure like formatting way"?

docs/docs/blocks/!/AnyMore.md
Kryptos-FR: Need some utilities to manipulate strings (skip, substr, remove, concat...).

docs/docs/blocks/!/All.md
sinkingsugar: https://github.com/bbatsov/clojure-style-guide

docs/docs/blocks/!/AnyMore.md
sinkingsugar: Ok, that's also good issues material

docs/docs/blocks/!/Ref.md
sinkingsugar: Made an issue

src/tests/infos_docs.edn
sinkingsugar: https://github.com/bbatsov/clojure-style-guide

docs/docs/blocks/If.md
sinkingsugar: it's in a subfolder `!`, is that intentional?

docs/docs/blocks/!/All.md
Kryptos-FR: Lots of their rules are contradictory, so I don't like that.

docs/docs/blocks/If.md
Kryptos-FR: Yes. It is a hack for now.

If they stayed at the root, then all the folders would appear at the end in the navigation side-panel. I tried other characters (for instance `_` but that one is rendered as a white space so not helpful).

Try to move some files top the root and generate/serve the website, you will understand what I mean.

sinkingsugar: Sounds good

src/tests/infos_docs.edn
Kryptos-FR: AFAIK according to the guide (which I don't like because some rules contradict each other), it is already in the right form (single argument is fine on the same line).

docs/docs/blocks/!/All.md
sinkingsugar: Well, I'm open to better proposals but I'd rather follow the clojure/edn or even scheme community? How about scheme for example, what's their standard?

src/tests/infos_docs.edn
sinkingsugar: it is, I think just the way to close the block is not, that's it.

sinkingsugar: but we can keep this way for now anyway

docs/docs/blocks/!/TryMany.md
sinkingsugar: Maybe we can just postpone this?

docs/docs/blocks/!/All.md
Kryptos-FR: I don't know what scheme standard is. But what I don't like in the clojure guide is that sometimes they say one space indentation is not good, and in the next section they use a single space and in the next one they use yet again different alignment rules. It should be much simpler: 2 (or 4 spaces) alignment in all cases. It's all about readability and ease on the eyes/brain and should be context-free.

Anyway, that's a topic for another (low-priority) issue. We can skip those concerns in the current PR.

docs/docs/blocks/!/ChainRunner.md
Kryptos-FR: I'm not sure what those represent actually. But if they are texts, then it's better than showing hexadecimals.

docs/docs/blocks/!/TryMany.md
Kryptos-FR: Yeah, we can have another issue/PR for that. The purpose of the current PR is to have the foundation for doc generation.

docs/docs/blocks/!/All.md
sinkingsugar: Cool agreed

sinkingsugar (approved):

PR #82, opened 2021-07-16
add gltf joint processing - progress

sinkingsugar:


sinkingsugar:
Merging it for now as it has no impact, animation system/tracks to follow, as well as various refactors

PR #90, opened 2021-08-10
CMake cleanup

guusw:
Should be good now, only test failing on the branch is docker because doesn't have the credentials on the branch.


.gitmodules
```diff
@@ -79,3 +82,15 @@
...
+	url = https://github.com/chainblocks/bgfx.cmake.git
+[submodule "deps/brotli"]
+	path = deps/brotli
+	url = https://github.com/chainblocks/brotli.git
+[submodule "deps/snappy"]
+	path = deps/snappy
+	url = https://github.com/chainblocks/snappy.git
+[submodule "cmake/corrosion"]
+	path = cmake/corrosion
+	url = https://github.com/AndrewGaspar/corrosion.git
```
sinkingsugar, line 96: Please fork under "chainblocks" org

deps/sandbox.cpp
```diff
@@ -0,0 +1,3 @@
+int main() {
```
sinkingsugar, line 1: What's this?

include/chainblocks.hpp
```diff
@@ -59,7 +59,7 @@ class InvalidVarTypeError : public CBException {
   explicit InvalidVarTypeError(std::string_view msg) : CBException(msg) {}
 };
 
-CBlock *createBlock(std::string_view name);
+CHAINBLOCKS_API CBlock *createBlock(std::string_view name);
```
sinkingsugar, line 62: We actually want to export ONLY chainblocksInterface and consume from `CBCore` for many reasons - we could talk about in a issue or so. But in a nutshell it's for versioning and coherence, so those are not necessary

rust/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "chainblocks"
+name = "chainblocks-rust"
```
sinkingsugar, line 2: leave this as "chainblocks" we already have "chainblocks" on crates.io

rust/Cargo.toml
```diff
@@ -1,5 +1,5 @@
...
 description = "rust bindings and utility in order to use chainblocks"
```
sinkingsugar, line 3: This is wrong, actually, this includes also blocks and many now fundamental components

src/core/foundation.hpp
```diff
@@ -108,44 +108,44 @@ struct CBCoro {
 
 namespace chainblocks {
 #ifdef CB_COMPRESSED_STRINGS
-CBOptionalString getCompiledCompressedString(uint32_t crc);
+CHAINBLOCKS_API CBOptionalString getCompiledCompressedString(uint32_t crc);
```
sinkingsugar, line 111: don't export - as said before
Also those are internal API which we want to break often

src/core/foundation.hpp
```diff
@@ -856,8 +858,8 @@ inline void swap(CBVar &v1, CBVar &v2) {
 } // namespace std
 
 namespace chainblocks {
-NO_INLINE void _destroyVarSlow(CBVar &var);
-NO_INLINE void _cloneVarSlow(CBVar &dst, const CBVar &src);
+CHAINBLOCKS_API NO_INLINE void _destroyVarSlow(CBVar &var);
```
sinkingsugar, line 861: Again no export, expose only chainblocksInterface

src/extra/desktop.win.cpp
```diff
@@ -664,7 +664,7 @@ struct SendKeyEvent : public SendKeyEventBase {
       }
 
       auto lparam = MapVirtualKey(vkCode, MAPVK_VK_TO_VSC) << 16;
-      lparam = lparam | (state == 0) ? 0x00000001 : 0xC0000001;
+      lparam = lparam | (state == 0 ? 0x00000001 : 0xC0000001);
```
sinkingsugar, line 667: Is this some auto refactor? They look both suspicious :D

sinkingsugar (changes requested):
Amazing work, Thanks.

I added some comments.
One thing I wish we don't do as you see from comments is export things, only chainblocksInterface should be exported cos allows us to keep the system isolated, we can control and manipulate CBCore when we break things.
But maybe your intentions were different.

.gitmodules
```diff
@@ -79,3 +82,15 @@
 [submodule "deps/pareto"]
 	path = deps/pareto
 	url = https://github.com/chainblocks/pareto.git
+[submodule "deps/bgfx.cmake"]
+	path = deps/bgfx.cmake
+	url = https://github.com/chainblocks/bgfx.cmake.git
+[submodule "deps/brotli"]
```
sinkingsugar, line 88: I kinda like externalproject, those used to be modules actually :D

.gitmodules
```diff
@@ -79,3 +82,15 @@
 [submodule "deps/pareto"]
 	path = deps/pareto
 	url = https://github.com/chainblocks/pareto.git
+[submodule "deps/bgfx.cmake"]
```
sinkingsugar, line 85: Not sure on this

guusw:
> 
> 
> Amazing work, Thanks.
> 
> I added some comments.
> One thing I wish we don't do as you see from comments is export things, only chainblocksInterface should be exported cos allows us to keep the system isolated, we can control and manipulate CBCore when we break things.
> But maybe your intentions were different.

This should probably be in separate pull request but it was mostly to have mal interact with chainblocks if chainblocks is built as a dynamic library.
Will move it out of this PR before merging

.gitmodules
guusw: I like being able to use the target as part of the build system, it makes it easier to include another library (include paths & libs) without having to directly deal with library names and include paths.
It also easily allows you to switch any library between shared/static without having to change the lib names linked against.
Just thinking ahead a bit 😄

guusw: Also iterating on changes in dependencies becomes easier as cmake checks for changes in them

src/extra/desktop.win.cpp
guusw: I believe you mean `lparam | (state == 0 ? 0x00000001 : 0xC0000001)`
Clang was giving a warning that | takes precedence over ? so it would evaluate to
`(lparam | state == 0) ? 0x00000001 : 0xC0000001`

.gitmodules
sinkingsugar: Yeah you are right, not sure though why I moved to externalproject.
But indeed if the external project uses cmake probably it's overkill!

cmake/Rust.cmake
```diff
@@ -0,0 +1,50 @@
...
+endif()
+
+if(Rust_BUILD_SUBDIR_HAS_TARGET)
+  set(Rust_BUILD_SUBDIR ${Rust_CARGO_TARGET}/${Rust_BUILD_SUBDIR_CONFIGURATION})
+else()
+  set(Rust_BUILD_SUBDIR ${Rust_BUILD_SUBDIR_CONFIGURATION})
+endif()
+message(STATUS "Rust_BUILD_SUBDIR = ${Rust_BUILD_SUBDIR}")
+
+if(Rust_USE_LTO OR USE_LTO)
```
guusw, line 34: This should be Rust_USE_LTO only

deps/CMakeLists.txt
```diff
@@ -35,6 +37,7 @@ add_subdirectory(kissfft)
 
 option(BGFX_BUILD_EXAMPLES "" OFF)
 option(BGFX_INSTALL "" OFF)
+option(BIMG_DECODE_ENABLE "" OFF)
```
sinkingsugar, line 40: Don't we need this?

guusw: This was in the original CMakeLists too
```
if(NOT BUILD_CORE_ONLY)
  add_compile_definitions(
    CB_WITH_EXTRAS
    BIMG_DECODE_ENABLE=0
    ```

Comment (sinkingsugar) [reply deps/CMakeLists.txt:40]: ah right

**Comment by sinkingsugar** [comment]
What's the problem with ExternalProject btw?
It's very good to not mix codebases imho. I really would have disagreed on this change if we had chance to discuss 😄

**Comment by sinkingsugar** [comment]
@guusw what's the current state of this?
Also it will need some fixes when merging.

File: src/extra/CMakeLists.txt [hunk 13]
```diff
@@ -35,7 +35,7 @@ endif()
 
 target_link_libraries(chainblocks-extra
```
sinkingsugar, line 36: Are all those things statically linked no?
static linking destroys inlining, this is a breaking chance that needs to be profiled.

guusw: Many of these are targets that only contain include paths just to group together what belongs to what.

cpp/c files that were previously added directly to the chainblocks target are now static libraries but I believe this doesn't affect inlining since that happens through headers only

sinkingsugar (changes requested):
Wasm3 is diverging , I merged upstream recently on devel.
Tinygltf could use an upstream bump, so does implot. Could do another branch though for that.

.gitmodules
```diff
@@ -52,13 +53,14 @@
 [submodule "deps/wasm3"]
 	path = deps/wasm3
 	url = https://github.com/chainblocks/wasm3.git
-	branch = main
+	branch = cmake_cleanup
```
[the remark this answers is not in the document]
sinkingsugar: This is diverging things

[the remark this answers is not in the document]
guusw: Moved the commits back earlier. removed branch= statements from .gitmodules.

cmake/Platform.cmake
```diff
@@ -0,0 +1,141 @@
...
+
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+  add_compile_definitions(CPUBITS64)
+elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
+  add_compile_definitions(CPUBITS32)
+endif()
+
+# Default arch if ARCH is not set
+if(NOT ARCH)
+  if(X86 AND NOT EMSCRIPTEN)
```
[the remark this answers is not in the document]
guusw: Not sure, but this was how it's originally set `-march=sandybridge` on `x86 and not emscripten` (same for rust)

I removed all the other non-default cases, but you could still use them as
`cmake -DARCH=core2/i686/native/etc.` during configure

guusw:
Restored all the deps subcommits to same revisions as devel

cmake/Platform.cmake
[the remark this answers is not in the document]
sinkingsugar: hmm I see, can't rem why

sinkingsugar:
I'm still confused about submodules. I see that our master of tinygltf is still pointing to your commit for example.

guusw:
Forgot that made the change on the master branch there, I put it back to the commit before that

sinkingsugar:
Almost ready to merge. Nice.
After reviewing bgfx.cmake I really don't like this extra cost of maintenance, can we restore the old way using amalgamated files?

sinkingsugar:
Also those are missing I think https://github.com/fragcolor-xyz/chainblocks/blob/e9d389b52f4d792ddbc8b1444684b7158eac109d/CMakeLists.txt#L247-L256

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
```
Kryptos-FR, line 4: That shouldn't be there IMO. End of line should either be left to the user (git has a checkout config to deal with that) or on a per file extension basis (using `.gitattributes` file).

Each system (Windows, Linux, Mac) has its own convention and tools/editor/CLI work better with the native line endings.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
```
Kryptos-FR, line 5: I am a little worried that this would affect all kind of files (incl. binary files).

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
```
Kryptos-FR, line 3: That wildcard is dangerous, we should have an exhaustive list of extensions.

Kryptos-FR (changes requested):

.editorconfig
guusw: Unless there are binary tools that read .editorconfig files I don't think it does

.editorconfig
guusw: Removed

.editorconfig
guusw: Removed

guusw:
> Almost ready to merge. Nice. After reviewing bgfx.cmake I really don't like this extra cost of maintenance, can we restore the old way using amalgamated files? @guusw

I still prefer to use this. I use some of the dependencies inside the graphics branch and will probably use some of the other formats supported by bimg for cubemaps/environments.

It's maintained by someone under the bgfx developer's github name. There are indeed some changes but it doesn't have to be synced at the same time as bgfx/bimg/bx since it's only the build system.

guusw:
> Also those are missing I think
> 
> https://github.com/fragcolor-xyz/chainblocks/blob/e9d389b52f4d792ddbc8b1444684b7158eac109d/CMakeLists.txt#L247-L256

They are in `src/core/CMakeLists.txt`@100
Boost is in `cmake/Boost.cmake`

guusw:
Opening new PR with local branch

PR #92, opened 2021-08-11
Fix incorrect trimming of keywords when used as key

### LINKED ISSUE #91: EDN keywords get incorrectly trimmed when used as key in a map
### Description

Consider the following EDN input:
```clojure
; input.edn
{:keyword "value1", "key string" "value2"}
```

Read this input in a simple chainblock program:
```clojure
(defnode main)
(defchain keyword_test
  "keyword_test"
  (read-string (slurp "input.edn")))

(schedule main keyword_test)
(run main)
```

### Expected

The result of `(read-string)` is a `CBTable` with two entries:
```
"keyword" -> "value1"
"key string" -> "value2"
```

### Actual

The result of `(read-string)` is a `CBTable` with two entries, but the keyword is missing the last character:
```
"keywor" -> "value1"
"key string" -> "value2"
```

Kryptos-FR:
Keywords start with a ':' while simple keys are quoted strings.

When used as key in a map, quoted strings need to be "unescaped", and keywords need to only lose the leading ':' character.

Fixes #91

### DIFF
--- a/src/mal/CBCore.cpp
+++ b/src/mal/CBCore.cpp
@@ -936,7 +936,8 @@ std::vector<malCBlockPtr> blockify(const malValuePtr &arg) {
     for (auto [k, v] : v->m_map) {
       auto cbv = varify(v);
       vars.emplace_back(cbv);
-      cbMap[unescape(k)] = cbv->value();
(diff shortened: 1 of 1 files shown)


sinkingsugar (approved):
LGTM

PR #93, opened 2021-08-16
fix writepng and premultiplied alpha

sinkingsugar:
as well as flag premul svg image output

### DIFF
--- a/src/core/blocks/serialization.cpp
+++ b/src/core/blocks/serialization.cpp
@@ -361,6 +361,8 @@ struct LoadImage : public FileBase {
 };
 
 struct WritePNG : public FileBase {
+  std::vector<uint8_t> _scratch;
+
   static CBTypesInfo inputTypes() { return CoreInfo::ImageType; }
   static CBTypesInfo outputTypes() { return CoreInfo::ImageType; }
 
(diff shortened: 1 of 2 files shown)


src/core/blocks/serialization.cpp
```diff
@@ -385,9 +387,75 @@ struct WritePNG : public FileBase {
...
+            uint8_t g =
+                input.payload.imageValue.data[(ih * lineSize) + (iw * 4) + 1];
+            _scratch[(ih * lineSize) + (iw * 4) + 1] =
+                uint8_t(float(g) / fa + 0.5);
+            uint8_t b =
+                input.payload.imageValue.data[(ih * lineSize) + (iw * 4) + 2];
+            _scratch[(ih * lineSize) + (iw * 4) + 2] =
+                uint8_t(float(b) / fa + 0.5);
+          } else if (pixsize == 2) {
+            uint16_t *source =
```
sinkingsugar, line 417: Notice, no coverage yet

src/core/blocks/serialization.cpp
```diff
@@ -385,9 +387,75 @@ struct WritePNG : public FileBase {
...
+            dest[(ih * lineSize) + (iw * 4) + 0] =
+                uint16_t(float(r) / fa + 0.5);
+            uint16_t g = source[(ih * lineSize) + (iw * 4) + 1];
+            dest[(ih * lineSize) + (iw * 4) + 1] =
+                uint16_t(float(g) / fa + 0.5);
+            uint16_t b = source[(ih * lineSize) + (iw * 4) + 2];
+            dest[(ih * lineSize) + (iw * 4) + 2] =
+                uint16_t(float(b) / fa + 0.5);
+          } else if (pixsize == 4) {
+            float *source =
```
sinkingsugar, line 433: Notice, no coverage yet

src/core/blocks/serialization.cpp
```diff
@@ -385,9 +387,75 @@ struct WritePNG : public FileBase {
...
-      throw ActivationError("Failed to write PNG file.");
+
+    // demultiply alpha if needed, limited to 4 channels
+    if (c == 4 &&
+        (input.payload.imageValue.flags & CBIMAGE_FLAGS_PREMULTIPLIED_ALPHA) ==
+            CBIMAGE_FLAGS_PREMULTIPLIED_ALPHA) {
+      _scratch.resize(w * h * 4 * pixsize);
+      int lineSize = w * 4;
+      for (int ih = 0; ih < h; ih++) {
+        for (int iw = 0; iw < w; iw++) {
```
Kryptos-FR, line 398: Is it done in two nested loops for better readability or is there another reason?

src/core/blocks/serialization.cpp
```diff
@@ -385,9 +387,75 @@ struct WritePNG : public FileBase {
...
+          if (pixsize == 1) {
+            uint8_t a =
+                input.payload.imageValue.data[(ih * lineSize) + (iw * 4) + 3];
+            float fa = float(a) / 255.0f;
+            _scratch[(ih * lineSize) + (iw * 4) + 3] = a;
+            uint8_t r =
+                input.payload.imageValue.data[(ih * lineSize) + (iw * 4) + 0];
+            _scratch[(ih * lineSize) + (iw * 4) + 0] =
+                uint8_t(float(r) / fa + 0.5);
```
Kryptos-FR, line 407: Might save a few CPU cycles to do the index calc only once:
```cpp
auto index = (ih * lineSize) + (iw * 4) + 0;
uint8_t r = input.payload.imageValue.data[index];
 _scratch[index] = uint8_t(float(r) / fa + 0.5);
```

src/core/blocks/serialization.cpp
```diff
@@ -361,6 +361,8 @@ struct LoadImage : public FileBase {
 };
 
 struct WritePNG : public FileBase {
+  std::vector<uint8_t> _scratch;
```
Kryptos-FR, line 364: Is it necessary to have a field used as buffer? Can't it be a temporary variable?

Kryptos-FR (commented):
A few remarks, but algo looks good otherwise.

src/core/blocks/serialization.cpp
sinkingsugar: Just simplicity, I think any half decent compiler will optimize properly, but not to the extent of packing into 32 bit int.
The thing is that this block is slow in any case, we don't really need to optimize them yet.
Using (Await (WritePNG)) is advised (will run on a worker thread)

sinkingsugar: If I pack in 32 bits then 16bit and 32bit should be done with vectors :D - let's just add comments about it for really when there is time to waste

src/core/blocks/serialization.cpp
sinkingsugar: I think any half decent compiler will optimize those really but for readability let's do that

src/core/blocks/serialization.cpp
sinkingsugar: Yeah caching, if for some reason this block is called many times there will be no allocations at all (if size <=)

sinkingsugar: Btw this is often done all over the codebase exactly for this reason.

Kryptos-FR (approved):

PR #95, opened 2021-08-25
Sprite.Sheet and Sprite.Draw blocks

Kryptos-FR:
Implements blocks `Sprite.Sheet` and `Sprite.Draw`.

Setup a spritesheet by reading an `.edn`file (can be generated from a spine `.atlas` file using [spine-rust](https://github.com/Kryptos-FR/spine-rust)):
```clojure
(Setup
  (LoadImage "data/skeleton.png") (GFX.Texture2D) = .texture
  (read-string (slurp "data/skeleton.edn")) = .atlas
  (Sprite.Sheet :Atlas .atlas) = .sheet )
```

Draw the first sprite of the sprite sheet:
```clojure
(Sprite.Draw :Texture .texture :Sheet .sheet :Index 0
             :Shader .shader :Blend {"Src" Blend.SrcAlpha "Dst" Blend.InvSrcAlpha "Op" BlendOp.Add})
```

Index can be a variable (so it can later be animated):
```clojure
(->
  1 >= .index)
(Sprite.Draw :Texture .texture :Sheet .sheet :Index .index
             :Shader .shader :Blend {"Src" Blend.SrcAlpha "Dst" Blend.InvSrcAlpha "Op" BlendOp.Add})
```

Some considerations:
* `(Sprite.Draw)` needs to be given a shader. We should have a default fallback.
* `:Blend` do have a default fallback, which is the same as here (i.e. `Src * alpha + Dst * (1 - alpha)`).

See [sprite_test.edn](https://github.com/fragcolor-xyz/chainblocks/blob/3094b179da17c6394b3dbd81396d5a0d954db13c/src/tests/sprite_test.edn) for a complete example.


src/extra/sprite.cpp
```diff
@@ -0,0 +1,406 @@
...
+           sizeof(float) * 4);
+    bgfx::setTransform(idx, 1);
+
+    // step: create quad
+    auto index = _index.get().payload.intValue;
+    const auto &region = sheet->regions[index % sheet->regions.size()];
+    bgfx::TransientVertexBuffer vb;
+    // FIXME: bound/size doesn't work if the sprite is rotated
+    createQuad(&vb, region.width / (float)currentView.width,
+               region.height / (float)currentView.height);
```
Kryptos-FR, line 193: This division is not correct if a rotation is applied to the sprite.

For example in `src\tests\sprite_test.edn`, if we change the transform to be
```clojure
90.0 (Math.DegreesToRadians) (Math.AxisAngleZ) (Math.Rotation) = .r
```

I believe we must used the projection matrix of the current view or something similar, but I'm not sure of the math involved.

src/extra/sprite.cpp
```diff
@@ -0,0 +1,406 @@
...
+public:
+  struct Region {
+    uint32_t width;
+    uint32_t height;
+    float uvs[4];
+  };
+
+  uint32_t width;
+  uint32_t height;
+  bool premultiply;
```
Kryptos-FR, line 96: For now it is not used, but I suppose it should influence the blend state somehow (when `true`).

Kryptos-FR: I changed the blend state to `Src * 1 + Dst * (1 - alpha)` in such case.

src/extra/sprite.cpp
guusw: Addressed in sprite_camera branch

Kryptos-FR:
Have a look at the two other branches `sprite_camera` and `sprite_camera2` for attempts to fix the transform issue (especially when rotation is involved).

src/extra/sprite.cpp
```diff
@@ -0,0 +1,453 @@
...
+      auto table = regions.elements[i].payload.tableValue;
+      auto api = table.api;
+
+      auto region = Region{};
+
+      auto bounds = api->tableAt(table, "bounds")->payload.seqValue;
+      region.width = bounds.elements[2].payload.intValue;
+      region.height = bounds.elements[3].payload.intValue;
+
+      if (api->tableContains(table, "uvs")) {
```
guusw, line 64: I'm thinking about if we actually need uvs. maybe we can just always specify pixel coordinates in the skeleton instead.

We could support both but it causes some confusion if both are specified like in the test sprite sheets since it leaves some values unused.

src/extra/sprite.cpp
```diff
@@ -0,0 +1,453 @@
...
+                         Enums::VAlign vAlign = Enums::VAlign::Top) {
+    if (6 ==
+        bgfx::getAvailTransientVertexBuffer(6, PosTexCoord0Vertex::ms_layout)) {
+      bgfx::allocTransientVertexBuffer(vb, 6, PosTexCoord0Vertex::ms_layout);
+      PosTexCoord0Vertex *vertex = (PosTexCoord0Vertex *)vb->data;
+
+      const float zz = 0.0f;
+
+      // note: alignments define the anchor point
+      const float minx = ((int)Enums::HAlign::Left - (int)hAlign) * width / 2;
```
guusw, line 358: For pixel perfect sprites this should be floor/ceiled so it doesn't get .5 fractional coordinates
Maybe it can be the default behavior but turned of with `:Snap false` or `:AlignToGrid false`, etc.

guusw:
There's a debug.edn/debug.png that's not used. maybe use that for the test or remove it.

guusw (changes requested):
There's a debug.edn/debug.png that's not used. maybe use that for the test or remove it.

src/tests/data/skeleton.edn
```diff
@@ -0,0 +1,38 @@
+{:filter  ["Linear", "Linear"]
+ :format  "RGBA8888"
+ :name    "skeleton.png"
+ :pma     false
+ :regions [{:bounds  [2, 42, 352, 785]
+            :index   -1
+            :name    "first animation test v1002"
+            :offsets [324, 162, 1000, 1149]
```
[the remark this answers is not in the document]
Kryptos-FR: It is converted from the spine format. I decided to include everything since it could also be used in authoring tools, including some that we will build later.

src/extra/sprite.cpp
```diff
@@ -0,0 +1,453 @@
...
+        for (uint32_t j = 0; j < uvs.len && j < 4; ++j) {
+          region.uvs[j] = (float)uvs.elements[j].payload.floatValue;
+        }
+      } else {
+        // calculate UVs
+        double x = bounds.elements[0].payload.intValue;
+        double y = bounds.elements[1].payload.intValue;
+        double u = x / (double)this->width;
+        double v = y / (double)this->height;
+        bool rotate = api->tableAt(table, "rotate")->payload.boolValue;
```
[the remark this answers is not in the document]
Kryptos-FR: Yeah this is a rotation in the atlas texture. It only matters when calculating UVs. It is just in case those were not already included in the .edn file.

src/extra/sprite.cpp
Kryptos-FR: Good point.

src/extra/sprite.cpp
Kryptos-FR: How does it work in the shader? Doesn't it need to have texture coordinates?

Kryptos-FR:
The debug file is useful for any kind of testing. So I thought to include it. I sometimes used it in the sprite_test.edn even though at the moment it uses the other skeleton sprite.

src/extra/sprite.cpp
guusw: I meant leave them out of the skeleton file and just always compute them using the bounds

src/extra/sprite.cpp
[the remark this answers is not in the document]
guusw: shouldn't it also be rotated back when rendering? (x1=y0, y1=-x0)
currently it only swaps the width/height from the bounds

src/tests/data/skeleton.edn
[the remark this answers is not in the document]
guusw: Can we change the converter to only output what we use currently? (if it's not too much work)
People will probably use the tests as code samples, and there will be a bunch of stuff that does nothing

src/extra/sprite.cpp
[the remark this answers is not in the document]
Kryptos-FR: That's only what's needed.

src/extra/sprite.cpp
Kryptos-FR: It was to save a few computations 😉 but maybe not worth it.

src/extra/sprite.cpp
Kryptos-FR: Looking at it again, in the current `sprite` branch, the coordinates are normalized between 0 and 1, so can't use floor/ceiling there.

However, whatever solution we decide to use to fix the rotation issue (either branches `sprite_camera`, `sprite_camera2` or another solution), we can think of adding this "snap to grid" option.

sinkingsugar:
What do we want to do with this @guusw @Kryptos-FR ?

Kryptos-FR:
@sinkingsugar @guusw it was based on bgfx, which we don't have anymore.

That said, the basic EDN "look" of it could be kept. It all comes down to whether we want to support that kind of sprite atlases. We should have a meeting about it.

sinkingsugar:
Still it's a PR, so I'd close for now.

Kryptos-FR:
Closing for now. Will need some rework if we include it later.

PR #99, opened 2021-09-08
Update copyright notices in source code files

Kryptos-FR:
Added or updated  copyright notices in all source files (`.clj`, `.cpp`, `.edn`, `.h`, `.hpp`, `.js`, `.py`), as well as docker files.

Fixed SPDX identifier: `BSD-3-Clause` instead of the full name (see https://spdx.org/licenses/).

Note that mal source files are under the MPL-2.0 license instead. And `CBCore.cpp` is under both.

To simplify, copyright notices only include the first year of publication. There is no need to update the year, even if the file is modified in the future. Similar recommendations can be found at https://opensource.google/docs/copyright/.

```
/* SPDX-License-Identifier: BSD-3-Clause */
/* Copyright © 2019 Fragcolor Pte. Ltd. */
```

### DIFF
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright © 2019 Fragcolor Pte. Ltd.
+
 FROM ubuntu:20.04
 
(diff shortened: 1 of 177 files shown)


sinkingsugar (approved):
LGTM

PR #100, opened 2021-09-11
Set display in bgfx platform data

guusw:
Fixes a crash when creating a BGFX window on linux.
It's calling XGetXCBConnection(nullptr) if the display is not set in the platform data

Simplified SDL_GetNativeWindowPtr definition a bit to remove the specific types, since they're not used anyways

### DIFF
--- a/src/extra/bgfx.cpp
+++ b/src/extra/bgfx.cpp
@@ -325,12 +325,8 @@ void PushFont(Font::Enum _font) {
 } // namespace ImGui
 
 namespace BGFX {
-// delay this at end of file, otherwise we pull a header mess
-#if defined(_WIN32)
-HWND SDL_GetNativeWindowPtr(SDL_Window *window);
-#elif defined(__linux__)
 void *SDL_GetNativeWindowPtr(SDL_Window *window);
-#endif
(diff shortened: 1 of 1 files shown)


src/extra/bgfx.cpp
```diff
@@ -774,6 +768,7 @@ struct MainWindow : public BaseWindow {
     // set platform data this way.. or we will have issues if we re-init bgfx
     bgfx::PlatformData pdata{};
     pdata.nwh = _sysWnd;
+    // pdata.ndt = SDL_GetNativeDisplayPtr(_window);
```
sinkingsugar, line 771: how about this?

sinkingsugar (approved):
Cool, I think there is just an extra comment to remove.

src/extra/bgfx.cpp
guusw: 😅 don't know how I managed to forget the actual fix, thanks

sinkingsugar:
Does this fix sporadic linux CI failure too ? Not sure though , as that is more VM related maybe

guusw:
Don't think so, I think I encountered this because I built linux with the vulkan BGFX backend.

PR #101, opened 2021-09-13
Update and improve documentation for blocks

Kryptos-FR:
Add license notice in metadata (visible on GitHub but not on the generated website).

```markdown
---
authors: Fragcolor & contributors
license: CC-BY-SA-4.0
---
```

Add sample `.edn` files for blocks related to source files:
- assert.cpp
- flow.cpp
- math.hpp
- bigint.cpp
- chains.cpp

Also update some of the `help()` function in those same files.

The samples for assert, flow, math are trivial. However those for bigint and chains might need a deeper look at/review.
All samples can be run using `./docs/samples/run_sample.edn`. For example:

```
build/cbl.exe docs/samples/run_sample.edn --file blocks/Math/Abs.edn
```
Later we can make it a CI job.


sinkingsugar:
docs/samples/data/testChain.bin might be an extra file we don't need, do we?

Kryptos-FR:
> docs/samples/data/testChain.bin might be an extra file we don't need, do we?

We can generate it during the test.

sinkingsugar (approved):
Nice work 💯

PR #102, opened 2021-09-16
Make built-in types lowercase

Kryptos-FR:
The idea is to be consistent with the following rule:
- built-ins are lowercase and interpreted by mal (hence they don't need to run inside a chain)
- blocks are Capitalized and they do need to run in the context of a chain

This should help writing code by having a visual guide (sometimes it is complicated to mix blocks and built-ins).

However, since some basic types have existed for a while with a capitalized letter, we know have uppercase overrides for backward compatibility and convenience. It is a slight deviation from the above rule but since it only concerns basic structures, it shouldn't matter too much.

### DIFF
--- a/src/mal/CBCore.cpp
+++ b/src/mal/CBCore.cpp
@@ -201,6 +201,22 @@ void installCBCore(const malEnvPtr &env, const char *exePath,
   rep("(defmacro! defblocks (fn* [name args & blocks] `(defn ~(symbol (str "
       "name)) ~args (chainify (vector ~@blocks)))))",
       env);
+
+  // overrides for some built-in types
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
Given that, once this change is merged, I should probably update the samples in the documentation to use the lowercase version.

sinkingsugar (approved):
LGTM, not sure while it's failing around though.

Kryptos-FR:
> not sure while it's failing around though

Me neither. Is there a way to manual force another CI build?
And which test files should I try locally to try to reproduce if there is a regression?

Kryptos-FR:
It was a simple mistake: I forgot the parentheses in the `def!` expression.

PR #104, opened 2021-09-16
Documentation - experimental API

Kryptos-FR:
The main enhancement in this PR is the ability to flag a block as "experimental" and to have the documentation display a warning so that users are aware it can break things.

It is implemented as a new function on the `CBlock` struct:
```cpp
CBTable * properties(CBlock *block);
```

Having it defined on the block itself would allow to add all kind of metadata later one and to add some checks for instance in `hasten` to forbid the use of certain unstable APIs.

On the documentation side, it inserts a note:
```
!!! warning "Experimental/unstable"
    This API is experimental and might produce unexpected results or crash.
```
which, thanks to the [admonition extension](https://squidfunk.github.io/mkdocs-material/reference/admonitions/), renders it nicely in mkdocs.

There are few other minor changes in the documentation, mostly cosmetic:
- change the base colors to better match the Fragcolor brand (b130447889a56db8e605a0116622ae6e3cf1a870)
- inject Creative Commons logos matching the license at the bottom of each page (38f4db2161ae60fe3d217f1e39e385350e623c44)

*Note: the documentation hasn't been re-generated with those changes yet. I will be done after the review, if changes are accepted.*


include/utility.hpp
```diff
@@ -517,6 +517,14 @@ template <class CB_CORE> struct TTableVar : public CBVar {
 
   TTableVar(TTableVar &&other) : CBVar() { std::swap(*this, other); }
 
+  TTableVar(std::initializer_list<std::pair<std::string_view, CBVar>> pairs)
+      : TTableVar() {
+    for (auto &kv : pairs) {
+      *payload.tableValue.api->tableAt(payload.tableValue, kv.first.data()) =
```
sinkingsugar, line 523: This is wrong, it would need `cloneVar` (this is a C api so no magic happens)
I'd suggest using `(*this)[key]` instead.

sinkingsugar (changes requested):
Just one small issue

include/utility.hpp
Kryptos-FR: I'm confused with what the issue is. From what I can tell the `[]` operator does the same thing.

include/blockwrapper.hpp
```diff
@@ -93,6 +94,17 @@ template <class T> struct BlockWrapper {
...
 
+    // properties
+    if constexpr (has_properties<T>::value) {
+      result->properties =
+          static_cast<CBPropertiesProc>([](CBlock *b) -> const CBTable * {
+            return reinterpret_cast<BlockWrapper<T> *>(b)->block.properties();
+          });
+    } else {
+      // properties is optional!
+      result->properties = nullptr;
```
sinkingsugar, line 105: Since Rust is implementing it with none and returning a null table
We could just do the same here and return a null table.
Usually in the case above, rust side I implement a `hasCompose` etc for example.

src/extra/audio.cpp
```diff
@@ -64,6 +64,9 @@ struct Device {
   static CBTypesInfo inputTypes() { return CoreInfo::AnyType; }
   static CBTypesInfo outputTypes() { return CoreInfo::AnyType; }
 
+  static inline TableVar _props{{"experimental", Var(true)}};
```
sinkingsugar, line 67: Maybe the compiler optimizes but this could be shared just like params (common_types.hpp) since it's often repeated.
Same in rust.

sinkingsugar: Btw it's just nitpick optimization

sinkingsugar (changes requested):
Some small points. Otherwise LGTM.

src/extra/audio.cpp
Kryptos-FR: I was thinking about that but later when we want to add other tags, how will it compose?

include/blockwrapper.hpp
Kryptos-FR: Correct, let's go with the null table.

src/extra/audio.cpp
Kryptos-FR: I added a new constructor to allow composition.

sinkingsugar (approved):
LGTM

PR #107, opened 2021-09-22
Various fixes guus

guusw:
Some code-only changes extracted from the cmake branch.

Most of these were encountered when compiling with clang.
It doesn't fix everything but it compiles and doesn't crash on startup

- Change wasm3 module to https://github.com/chainblocks/wasm3.git (main branch)
- Global initialization order fix for clang
- Missing header  \<optional\>
- Missing inline in header function
- Explicit usage of CBType values where they're ambiguous
- EXPORT macro in source file
- Don't install signal handlers when a debugger is attached
- make "Overriding enum type" a debug log so it doesn't show up on stdout during test discovery

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -51,7 +51,8 @@
 	url = https://github.com/chainblocks/libbacktrace.git
 [submodule "deps/wasm3"]
 	path = deps/wasm3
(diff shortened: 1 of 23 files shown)


sinkingsugar (approved):
LGTM, to check why CI fails though

PR #108, opened 2021-09-25
Refactor CI jobs for Windows

Kryptos-FR:
- use a matrix for Debug/Release as well as 32/64bits
- split build and test

It should make the documentation workflow a bit faster as we skip the tests.


sinkingsugar (approved):
LGTM - can remove .sh files no?

PR #112, opened 2021-09-29
Implement Take for all types

guusw:
This PR adds a number_types.hpp/cpp that has properties about all chainblocks number types Int/Float/Color.

The implementation of Take was changed to use these properties.
The implementation of To<type> (ToFloat3, ToInt3, etc.) was also changed to use these properties

The main reason for these changes is to support integer types for Take.
It also moves common properties/operations on number types to one location so that if it's ever needed by other blocks the same code can be used.


include/chainblocks.h
```diff
@@ -70,13 +70,6 @@ enum CBInlineBlocks : uint32_t {
   CoreRefTable,
   CoreUpdate,
   CoreSwap,
-  CoreTakeSeq,
```
sinkingsugar, line 73: These are still useful if we want to literally inline them during the threaded execution loop.
Would avoid completely an indirect call and just simply inline all in a cache line - there are drawbacks though, I would do it for some type, like TakeSeq maybe, or float vectors

src/core/runtime.hpp
```diff
@@ -236,26 +236,6 @@ activateBlock(CBlock *blk, CBContext *context, const CBVar &input) {
     auto cblock = reinterpret_cast<chainblocks::BlockWrapper<Input> *>(blk);
     return cblock->block.activate(context, input);
   }
-  case CoreTakeSeq: {
```
sinkingsugar, line 239: This is the place where inlining optimizations might happen as mentioned before at Enum, they might be worth for Seq, String, Bytes to avoid indirections

src/core/blocks/casting.cpp
```diff
@@ -242,278 +242,113 @@ template <CBType FROMTYPE> struct ToBytes {
...
+      throw ComposeError("Conversion not implemented for this type");
+    }
+
+    _outputNumberType = NumberTypeLookup::getInstance().get(_outputVectorType->numberType);
+
+    return _outputVectorType->type;
+  }
+
+  void parseVector(CBVar &output, const CBVar &input) {
+    const VectorTypeTraits *inputVectorType =
```
sinkingsugar, line 269: Input type won't change, can we move those lookups into `warmup` and cache them?

src/core/blocks/casting.cpp
```diff
@@ -242,278 +242,113 @@ template <CBType FROMTYPE> struct ToBytes {
...
+      dstPtr += conversion->outStride;
+    }
+  }
+
+  void parseSeqElements(CBVar &output, const CBSeq &sequence) {
+    uint8_t *dstPtr = (uint8_t*)&output.payload;
+    for (size_t i = 0; i < sequence.len; i++) {
+      const CBVar &elem = sequence.elements[i];
+
+      const NumberTypeTraits *elemNumberType =
```
sinkingsugar, line 298: to warmup?

sinkingsugar (changes requested):
Looks great, just commented about some performance concern

include/number_types.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+enum class NumberType {
+  Float32,
+  Float64,
+  UInt8,
+  Int8,
+  Int16,
+  Int32,
+  Int64,
+  Invalid
```
Kryptos-FR, line 17: I would put `Invalid` as the first entry, since `0` is the default value for an enum. Easier to spot mistakes.

include/number_types.hpp
```diff
@@ -0,0 +1,139 @@
...
+std::map<CBType, NumberType> getCBTypeToNumberTypeMap();
+const std::vector<NumberType> &getCBTypeToNumberTypeArrayMap();
+
+inline NumberType cbTypeToNumberType(CBType type) {
+  auto &mapping = getCBTypeToNumberTypeArrayMap();
+  cbassert((size_t)type < mapping.size());
+  return mapping[(size_t)type];
+}
+
+struct NumberConvertOutOfRangeEx {
```
Kryptos-FR, line 29: I don't know about the best practice in C++, but in .NET exception names are nouns not verbs.
There fore I would have named it `NumberConversionOutOfRangeEx`.

src/core/number_types.cpp
```diff
@@ -0,0 +1,278 @@
...
+    {Int2, NumberType::Int64},
+    {Int3, NumberType::Int32},
+    {Int4, NumberType::Int32},
+    {Int8, NumberType::Int16},
+    {Int16, NumberType::Int8},
+    {Float, NumberType::Float64},
+    {Float2, NumberType::Float64},
+    {Float3, NumberType::Float32},
+    {Float4, NumberType::Float32},
+    {Color, NumberType::UInt8},
```
Kryptos-FR, line 19: nit pick: let's keep the same order everywhere for easier maintenance.

* see number_types.hpp lines 10 to 16
* see number_types.cpp lines 49 to 55

src/core/number_types.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }                                                                          \
+  };
+
+NUMBER_TYPE_TRAITS(NumberType::Int64, int64_t);
+NUMBER_TYPE_TRAITS(NumberType::Int32, int32_t);
+NUMBER_TYPE_TRAITS(NumberType::Int16, int16_t);
+NUMBER_TYPE_TRAITS(NumberType::Int8, int8_t);
+NUMBER_TYPE_TRAITS(NumberType::UInt8, uint8_t);
+NUMBER_TYPE_TRAITS(NumberType::Float64, double);
+NUMBER_TYPE_TRAITS(NumberType::Float32, float);
```
Kryptos-FR, line 129: nit pick: order

src/core/number_types.cpp
```diff
@@ -0,0 +1,278 @@
...
+  }
+};
+
+template <typename TIn> struct TNumberStringOperations {};
+
+template<> struct TNumberStringOperations<float> {
+  static void parse(float* out, const char* input, char** inputEnd) {
+    out[0] = std::strtof(input, inputEnd);
+  }
+};
```
Kryptos-FR, line 71: nit pick: order of this method and the following ones.

include/number_types.hpp
```diff
@@ -0,0 +1,139 @@
...
+#include <cassert>
+
+enum class NumberType {
+  Float32,
+  Float64,
+  UInt8,
+  Int8,
+  Int16,
+  Int32,
+  Int64,
```
Kryptos-FR, line 16: nit pick: order of this enum, compared to the implementations.

Kryptos-FR (changes requested):
A few remarks related to code quality.

src/core/blocks/casting.cpp
guusw: The switch is there for each element in the original TO_SOMETHING implementation.
```
 for (uint32_t i = 0; i < 1 && i < input.payload.seqValue.len; i++) {   \
          if (convert(output, input.payload.seqValue.elements[i]))             \
            return output;                                                     \
        }  
...
 bool convert(CBVar &dst, const CBVar &src) {                               \
      switch (src.valueType) {                                                 \
```
Can I check all the input element types of a seq during warmup and is it possible for them to always be known during warmup?

sinkingsugar: The original is probably not so an accurate implementation sadly.
Yes blocks can enforce sequence types, in this case they should enforce that the input sequence is always of the same type...
But I see in this case it's not it takes `any`. 
I'm afraid that this approach is much slower than the previous one though in that case if we have to do hash and lookup. (although likely fast yet not the same of a switch)

Possible solution:
* Evaluate in compose what input types you are getting `data.inputType`
* If it's a sequence with a single type you can optimize the lookup, type won't ever change (except if Any)
* If it's a sequence with multiple types we need to do the lookup variation of the activation as we don't know at runtime

include/number_types.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+  const NumberConversion *getConversion(NumberType from, NumberType to) {
+    const NumberTypeTraits *fromTypeTraits = get(from);
+    if (!fromTypeTraits)
+      return nullptr;
+
+    return fromTypeTraits->conversionTable.get(to);
+  }
+
+  static inline NumberTypeLookup &getInstance() {
```
sinkingsugar, line 102: hey @guusw I thought you were doing a hash table lookup.. but this is just an array. Should be as fast as switch.
My bad.

sinkingsugar: Still further optimizing if you know it ahead of time in compose would be great!

PR #113, opened 2021-10-03
cbl macro for rust and related fixes

sinkingsugar:
```rust
let chain_var = cbl!("(defchain x (Msg \"hello\"))");
```


sinkingsugar:
Keeping it disabled for now in the end, but I fixed a few minor things + small PR anyway so I'd rather merge it

.github/workflows/main.yml
```diff
@@ -52,12 +52,11 @@ jobs:
           rustup target add wasm32-unknown-emscripten
       - name: Set up rust (mt)
         if: ${{ matrix.threading == 'mt' }}
-        run: |        
+        run: |
```
Kryptos-FR, line 55: Do you have a formatter for YAML in VS code? I noticed that I sometimes involuntarily add those white spaces when pasting.

I tried the RedHat one but it changes single quotes to double quotes which I don't always want.

CMakeLists.txt
```diff
@@ -1029,6 +1033,27 @@ if(EMSCRIPTEN)
 endif()
 ###
 
+### mal as dll
+add_library(
+  libcb SHARED
```
Kryptos-FR, line 1038: What's the implication? Does it mean `cbl.exe` is not a standalone anymore and will have the mal dll aside?

rust/src/lib.rs
```diff
@@ -136,8 +136,10 @@ macro_rules! blocks {
             blk
     }};
 
-    (@block $a:expr) => { blocks!(@block Const $a); };
+    (@block . $a:expr) => { blocks!(@block Const $a); };
```
Kryptos-FR, line 139: What's the reason for this `.`?

Maybe a single line of comment to explain it, has it might not be obvious to a new reader (it's not for me at least).

Kryptos-FR (commented):
LGTM, just a few comments for my own good.

.github/workflows/main.yml
sinkingsugar: VS code default "remove whitespaces"

CMakeLists.txt
sinkingsugar: I wanted to get "cargo test" running.. WIP :D

rust/src/lib.rs
sinkingsugar: Good point.. to be honest it's a workaround still
I wanted to make just regular literals work but I could not
e.g.: `10 (Log)`
but in this macro `(. 10) (Log)` has to be used.
I will implement a `(.)` builtin maybe in mal to allow copy pasting of such scripts.. but not sure we are going to use either.. again WIP but don't wanna loose it in history. 
Will add this text in comments

Kryptos-FR (approved):

Issue #110, opened 2021-09-28
static thread local chain hashing set memory corruption on global destruction


sinkingsugar:
Windows and specifically when using mingw TLS is messy. We got some global destruction STL issues.

This fails consistently when running:
`gdb --args cbl.exe .\src\tests\subchains.clj`

See this image:
![image](https://user-images.githubusercontent.com/7008900/135001030-73ab55e7-d65d-41a0-ad18-c045e54e7378.png)

Location: https://github.com/fragcolor-xyz/chainblocks/blob/4ccf2302e53806ba17dbb8988d8be0dc64eed031/src/core/runtime.cpp#L2433
But this might also be an issue (also the 2 can be refactored): https://github.com/fragcolor-xyz/chainblocks/blob/4ccf2302e53806ba17dbb8988d8be0dc64eed031/src/core/runtime.cpp#L2358

I've tried refactoring a few times with no success with easy/trivial solutions. It's not easy to get rid of those collections given the way hashing works (stateless).

PR #115, opened 2021-10-09
Fix static thread local chain hashing set memory corruption on global destruction #110

sinkingsugar:
Best solution so far.
The reason being this crash happens ONLY `atexit`. And so for this specific set leaking is not a big deal at all.


sinkingsugar:
Valgrind show us the issue.
Also this solution potentially can leak badly if many different threads are created..

PR #121, opened 2021-10-14
Serialized -> External variables

sinkingsugar:
Added a way to declare and manage variables lifetime outside chains.
This is the starting point to have chains driven from external variables and as well as having variables sharing between chains (we have :Global actually but it's a bit different)

Some small bonus fixes.

### DIFF
--- a/src/core/runtime.hpp
+++ b/src/core/runtime.hpp
@@ -1206,20 +1206,6 @@ struct Serialization {
         chain->addBlock(blockVar.payload.blockValue);
         // blow's owner is the chain
       }
-      // // variables len
-      // read((uint8_t *)&len, sizeof(uint32_t));
-      // auto varsLen = len;
-      // for (uint32_t i = 0; i < varsLen; i++) {
-      //   read((uint8_t *)&len, sizeof(uint32_t));
-      //   buf.resize(len + 1);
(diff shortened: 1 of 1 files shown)


Kryptos-FR:
Could you add some high level concepts and explanation on how it is supposed to work?

I see that lot of code related to variable serialization was removed/replaced with the new code. However, it's hard to grasp what actually changed.

sinkingsugar:
> Could you add some high level concepts and explanation on how it is supposed to work?
> 
> I see that lot of code related to variable serialization was removed/replaced with the new code. However, it's hard to grasp what actually changed.

About the serialization code removed.
Actually those bits never worked. The Serialized notation was meant to be used in a total different way which right now I'm not caring about anymore as we promoted edn as the fragments language.
What I did here was trying to repurpose it but later realized it was useless.

External variables are much simpler.
Basically chains can have external variables. Those literally point to variables with lifetime managed outside the chain itself.
This will allow us to write a proper way to edit and link variables between chains.
Right it's done with mal built-in `set-var` but in the future a Rust or C++ orchestrator app will manage this.
Using `set-var` basically the lifetime is managed by mal ref-counting.

sinkingsugar:
Moreover to use external variables you don't need to do anything. 
Just set them (associating to a chain) and then use normally inside chains.

PR #129, opened 2021-10-26
ImGui widgets and demo

Kryptos-FR:
Add a lot of ImGui widgets and fix some others.

New widgets:
* ArrowButton (202b3283fb99aa4dfd5475264996551887532dd5)
* BulletText (a66a2f4e2705bae52c0ea63b8f3656f00b15d228)
* CheckboxFlags (86b469c5518271da9ccd8cb4b61e852775b7fa99)
* CollapsingHeader (b76e2e61f3c4e523bf686913d6474db5d1d8d8ae)
* Combo & ListBox (47e808407cbf286ebf881158d7df766a9d1a05e0)
* Disable (0497e5fe0ccfbf12c49994dacb8b2849e48bd1b1)
* FPS: displays the frame per second as calculated by ImGui (5d4809c7959f4f29052c2d9d401ce60383bd2fb3)
* Group (a88774849d978b68c42130dbcc6f6df9a1b5684c)
* HelpMarker: displays a `(?)` aside another widget and a text inside a popup on hover (6c2ca00cbd279f0768160f5de009324aa10c7be6)
* Menu, MenuItem, MenuBar, MainMenuBar (ed7e0b2f6244fa0d3a44661acf0c6bb27088e42c)
* NewLine (d7d0a2720f59f35fe1945011696767f6b57957a9)
* ProgressBar (4ab0d04b318c2f6df7039b81926e283e086fb6fe)
* Sliders: for `int` and `float` and variations such as `int2` (a197974782beaa44b3b36506fe50c77bf55181ec)
* Spacing (3e1ff8b3c3d5ea1259da8e33d59370a5fbf9bfae)
* Tab (a749cf3e5ac3c958113045f907af8168eb0f9dcc)
* Table (976bf02c172ddcc153069fa40678d63e52d4f455)
* Tooltip (05305462747eb4e20a0b7456bd138f14bc5ee3c8)
* Version: displays ImGui version (useful for debugging, and make sure we are not building against a wrong version from a different `deps`) (a1473cd9f2a124e88996a06aad120515a29aa808)

Fixes/changes:
* ChildWindow: fix size (e0f8a24afc364ccabd9275e0a6a22aa53e8c9903)
* Window:
  * close button (6165d21b03c6816134a18fb6d47c9e7648be81e5)
  * don't activate inner blocks when collapsed (17053ec31220f61cf287a075976d01a4a55f01fb)
  * auto resize (6aa3ccc1fbecf3508c1b46eab52ec691af681eb7)
* Plot: allow to set plot area width and height (fb65a3b25aee0d001fd540cc3b86a21cbf5f818c)
* Text: add wrapping (6cc86858d31bd515d47ccfc2ec2671035a1831da) and formatting support (6f4983f9750b34b30fcf6916c05b1a7aa430fda5) *(might need more love)*
* TextInput: add hint (82cc355a0536081ca685af9be461d3b27450e537)
* TreeNode: collapse by default (55c2f3030e75ac3174054ccc588c4afd79d3bb69)

Breaking change:
* Checkbox spelled with lowercase `b` (75d1188e055416740191b0dfca8f2c3c3f865e8a)

On top of those changes, a demo script which try to be as close as possible to the original ImGui demo (but is currently missing quite a lot). (3bb0913f41693917439c57a156e752e069c58225)


src/core/ops_internal.cpp
```diff
@@ -142,7 +142,7 @@ std::ostream &operator<<(std::ostream &os, const CBVar &var) {
     if (var.payload.stringValue == nullptr)
       os << "NULL";
     else
-      os << "\"" << var.payload.stringValue << "\"";
+      os << var.payload.stringValue;
```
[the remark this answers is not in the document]
Kryptos-FR: That code path doesn't seem to have any impact on `(read-var)` or `(set-var)`.

I did try to change the code in `variables.clj` to create a variable with a spaces in it, and with or without that change it doesn't work anyway, meaning variables with spaces are currently not supported.

I don't know if it is a regression, but it is not caused by that particular change.

src/extra/imgui.hpp
```diff
@@ -35,5 +37,85 @@ struct Context {
...
+  static constexpr uint32_t _NAME_##CC = _CC_;                                 \
+  static inline FlagsInfo<_NAME_> _NAME_##EnumInfo{#_NAME_, CoreCC,            \
+                                                   _NAME_##CC};                \
+  static inline Type _NAME_##Type = Type::Enum(CoreCC, _NAME_##CC)
+
+#define REGISTER_FLAGS_EX(_NAME_, _CC_)                                        \
+  REGISTER_FLAGS(_NAME_, _CC_);                                                \
+  static inline Type _NAME_##SeqType = Type::SeqOf(_NAME_##Type);              \
+  static inline Type _NAME_##VarType = Type::VariableOf(_NAME_##Type);         \
+  static inline Type _NAME_##VarSeqType = Type::VariableOf(_NAME_##SeqType)
```
Kryptos-FR, line 58: These macros could be moved to a more general header file, but I'm not sure which one.

`REGISTER_CBLOCK` is in `include\blockwrapper.hpp` but it doesn't seem like the right place for them.

src/extra/imgui.hpp
```diff
@@ -35,5 +37,85 @@ struct Context {
...
+    OpenOnDoubleClick = ::ImGuiTreeNodeFlags_OpenOnDoubleClick,
+    OpenOnArrow = ::ImGuiTreeNodeFlags_OpenOnArrow,
+    Leaf = ::ImGuiTreeNodeFlags_Leaf,
+    Bullet = ::ImGuiTreeNodeFlags_Bullet,
+    SpanAvailWidth = ::ImGuiTreeNodeFlags_SpanAvailWidth,
+    SpanFullWidth = ::ImGuiTreeNodeFlags_SpanFullWidth
+  };
+  REGISTER_FLAGS_EX(GuiTreeNodeFlags, 'guTN'); // FourCC = 0x6775544E
+
+  template <typename EnumType> static EnumType getFlags(CBVar var) {
```
Kryptos-FR, line 101: Similarly, this could be moved alongside the macros.

However it does need to include the `magic_enum` header to work (because of  `bitwise_operators` namespace), so we need to carefully select in which file to put it.

include/utility.hpp
```diff
@@ -303,6 +305,46 @@ template <class CB_CORE, typename E> class TEnumInfo {
     }
     info.labels.elements = &clabels[0];
     info.labels.len = uint32_t(labels.size());
+
+    for (auto &v : vseq) {
+      values.emplace_back(CBEnum(v));
```
sinkingsugar, line 310: If you use emplace your call is already a constructor call. So CBEnum should be redundant in this case.

include/utility.hpp
```diff
@@ -303,6 +305,46 @@ template <class CB_CORE, typename E> class TEnumInfo {
...
+    }
+    info.values.elements = &values[0];
+    info.values.len = uint32_t(values.size());
+
+    assert(info.values.len == info.labels.len);
+    CB_CORE::registerEnumType(vendorId, enumId, info);
+  }
+};
+
+template <class CB_CORE, typename E> class TFlagsInfo {
```
sinkingsugar, line 320: Isn't this repeating the previous class but just changing some types?
Can prob template that part.

src/extra/imgui.hpp
sinkingsugar: Generally Utility or Chainblocks.hpp

sinkingsugar (commented):
Getting there. Will finish review soon.

src/extra/imgui.hpp
Kryptos-FR: Actually, neither can work here since we need `EnumInfo` to be defined. I will put it in `foundation.hpp`.

include/utility.hpp
Kryptos-FR: Doesn't compile with some cryptic errors if I don't put that cast.
> C:/Tools/msys64/mingw64/include/c++/10.3.0/ext/new_allocator.h:150:4: error: cannot convert 'const chainblocks::Math::Mean::MeanKind' to 'int' in initialization

include/utility.hpp
Kryptos-FR: Power of `constexpr` 😄:
```cpp
template <class CB_CORE, typename E, bool isFlag = false> class TEnumInfo {
private:
  static constexpr auto enum_names() {
    if constexpr (isFlag) {
      return magic_enum::flags::enum_names<E>();
    } else {
      return magic_enum::enum_names<E>();
    }
  }
  static constexpr auto enum_values() {
    if constexpr (isFlag) {
      return magic_enum::flags::enum_values<E>();
    } else {
      return magic_enum::enum_values<E>();
    }
  }
  static constexpr auto eseq = enum_names();
  static constexpr auto vseq = enum_values();
```
It really feels magical.

src/extra/imgui.cpp
```diff
@@ -274,10 +271,10 @@ struct Style : public Base {
   static CBColor vec42Color(const ImVec4 &color) {
     // remember, we edited the shader to do srgb->linear
```
sinkingsugar, line 272: I think this statement is wrong, I'd remove

src/extra/imgui.cpp
```diff
@@ -541,10 +538,13 @@ struct Window : public Base {
   chainblocks::BlocksVar _blks{};
   std::string _title;
   bool firstActivation{true};
-  Var _pos{}, _width{1.0}, _height{1.0};
```
sinkingsugar, line 544: why?
I mean we still use relative 0.0, 1.0 I hope even if you added None

src/extra/imgui.cpp
```diff
@@ -805,7 +851,9 @@ template <CBType CT> struct Variable : public Base {
   ExposedInfo _expInfo{};
   bool _exposing = false;
 
-  virtual void cleanup() {
+  void cleanup() { cleanupVariable(); }
+
+  void cleanupVariable() {
```
sinkingsugar, line 856: Can we remove this totally and use ParamVar instead btw?

src/extra/imgui.cpp
```diff
@@ -883,7 +931,7 @@ template <CBType CT> struct Variable : public Base {
       } else {
         _variable_name = value.payload.stringValue;
       }
-      cleanup();
+      cleanupVariable();
```
sinkingsugar, line 934: This is not so good actually, ParamVar better (Not blaming you, it's very old code originally :) )

src/extra/imgui.cpp
```diff
@@ -902,41 +950,87 @@ template <CBType CT> struct Variable : public Base {
...
+  ExposedInfo _expInfo{};
+  bool _exposing = false;
 
-  CBVar activate(CBContext *context, const CBVar &input) {
-    IDContext idCtx(this);
+  void cleanup() { cleanupVariable(); }
 
-    if (!_variable && _variable_name.size() > 0) {
-      _variable = referenceVariable(context, _variable_name.c_str());
+  void cleanupVariable() {
```
sinkingsugar, line 965: ParamVar if possible

src/extra/imgui.cpp
```diff
@@ -947,8 +1041,14 @@ struct Text : public Base {
...
     } break;
-    case 1:
-      _color = value;
+    case 1: {
+      if (value.valueType == None) {
+        _variable_name.clear();
+      } else {
+        _variable_name = value.payload.stringValue;
+      }
+      cleanupVariable();
```
sinkingsugar, line 1050: Can't rem why I did this really. Anyway ParamVar would be ideal as mentioned above

src/extra/imgui.cpp
```diff
@@ -1247,323 +1392,2342 @@ struct TreeNode : public Base {
...
+struct FPS : public Base {
+  static CBTypesInfo outputTypes() { return CoreInfo::FloatType; }
+
+  CBVar activate(CBContext *context, const CBVar &input) {
+    ImGuiIO &io = ::ImGui::GetIO();
+    return Var(io.Framerate);
+  }
+};
+
+struct Version : public Base {
```
sinkingsugar, line 2979: We don't need this.
In fact `GUI` could be replaced one day with another library, we should detach from `ImGui`

sinkingsugar (approved):
I'd love to see a Table block that takes a mutable CBTable as input/param but maybe can be done later on

src/extra/imgui.cpp
Kryptos-FR: Yes we do (if the parameter is a float). It is just to support the auto-size-to-content when it is None.

Kryptos-FR:
> Table block that takes a mutable CBTable as input/param

Added in backlog.

PR #132, opened 2021-10-29
Substrate rpcs and extrinsic support

sinkingsugar:


sinkingsugar:
@Kryptos-FR I know it's not your thing but would be still good to have an OCD look :D

Cargo.toml
```diff
@@ -1,4 +1,7 @@
 [workspace]
 members = [
     "rust",
-]
\ No newline at end of file
+]
+
+[patch.crates-io]
+chrono = { git = "https://github.com/sinkingsugar/chrono" }
```
Kryptos-FR, line 7: Shouldn't we have the fork in the chainblocks org?

build-win.sh
```diff
@@ -62,5 +62,7 @@ echo "Running test: time"
 ./cbl ../src/tests/time.edn
 echo "Running test: eth"
 ./cbl ../src/tests/eth.edn
+echo "Running test: crypto"
```
Kryptos-FR, line 65: Let's remove those files.

build-win32.sh
```diff
@@ -54,5 +54,7 @@ echo "Running test: wasm"
 ./cbl ../src/tests/wasm.clj
 echo "Running test: eth"
 ./cbl ../src/tests/eth.edn
+# echo "Running test: crypto"
+# ./cbl ../src/tests/crypto.edn
```
Kryptos-FR, line 58: Idem (remove).

include/chainblocks.hpp
```diff
@@ -14,7 +14,7 @@
 #include <vector>
 
 namespace chainblocks {
-constexpr uint32_t CoreCC = 'sink'; // 1936289387
+constexpr uint32_t CoreCC = 'frag'; // 1718772071
```
Kryptos-FR, line 17: Should I also put the numerical value as a comment in places where I have defined a new enum? That seems like a good idea to detect potential collisions.

In which case, should we have a pattern for that comment? Like `// CC = 1718772071`. That way we can gather all of them with a simple search query.

include/common_types.hpp
```diff
@@ -15,6 +15,7 @@ At runtime just dlopen the dll, that's it!
 namespace chainblocks {
 
 enum class BasicTypes {
+  None,
```
Kryptos-FR, line 18: Is it needed (explicitly) or is it to avoid `Any` having the value `0`?

src/core/blocks/core.hpp
```diff
@@ -1543,6 +1544,9 @@ struct Sequence : public SeqBase {
 
   void addType(Types &inner, BasicTypes type) {
     switch (type) {
+    case BasicTypes::None:
```
Kryptos-FR, line 1547: In which cases does this happen?

include/common_types.hpp
```diff
@@ -15,6 +15,7 @@ At runtime just dlopen the dll, that's it!
...
   Any,
```
Kryptos-FR, line 19: Wondering if `Any` should have the value `-1` since it is a special case. Bonus point, only the value for `Any` would change from previously `0` to now `-1`. The numerical value for any other enum value would stay the same as before.

I'm saying that in case it causes a breaking change.

src/tests/crypto.edn
```diff
@@ -0,0 +1,135 @@
...
+  ; test from bytes
+  (HexToBytes)
+  (| (Sr25519.PublicKey) (ToHex)
+     (Assert.Is "0x44a996beb1eef7bdcab976ab6d2ca26104834164ecf28fb375600576fcc6eb0f" true))
+  (| (Ed25519.PublicKey) (ToHex)
+     (Assert.Is "0xd75a980182b10ab7d54bfed3c964073a0ee172f3daa62325af021a68f707511a" true))
+  (| (ECDSA.PublicKey) (ToHex)
+     (Assert.Is "0x048db55b05db86c0b1786ca49f095d76344c9e6056b2f02701a7e7f3c20aabfd913ebbe148dd17c56551a52952371071a6c604b3f3abe8f2c8fa742158ea6dd7d4"))
+
+  "caution juice atom organ advance problem want pledge someone senior holiday very"
```
Kryptos-FR, line 25: Is this your secret passphrase 😄?

src/tests/crypto.edn
```diff
@@ -0,0 +1,135 @@
...
+        (FromJson) (Log)))
+
+  (bytes "my_key") (ToHex) = .offchain-req
+  (if test-rpcs  ; this one requires --enable-offchain-indexing 1 --rpc-methods=Unsafe
+    (-> {"id" 1 "jsonrpc" "2.0" "method" "offchain_localStorageGet" "params" ["PERSISTENT" .offchain-req]} (ToJson) (Http.Post rpc-server)
+        (FromJson) (Log)))
+  ;
+  )
+(schedule root main)
+(run root)
```
Kryptos-FR, line 135: Nit pick: missing newline at the end of the file.

rust/src/types.rs
```diff
@@ -2263,4 +2338,14 @@ lazy_static! {
   };
   pub static ref FLOAT4X2_TYPES: Vec<Type> = vec![*FLOAT4X2_TYPE];
   pub static ref FLOAT4X2S_TYPE: Type = Type::seq(&FLOAT4X2_TYPES);
+  pub static ref TYPES_ENUM: Type = {
```
Kryptos-FR, line 2341: Shouldn't that be `ENUM_TYPES` to follow the same pattern as the values above?

rust/src/types.rs
```diff
@@ -2263,4 +2338,14 @@ lazy_static! {
...
+    let mut t = common_type::enumeration;
+    t.details.enumeration = CBTypeInfo_Details_Enum {
+      vendorId: 0x66726167, // 'frag'
+      typeId: 0x74797065, // 'type'
+    };
+    t
+  };
+  pub static ref TYPES_ENUM_TYPES: Vec<Type> = vec![*TYPES_ENUM];
+  pub static ref TYPES_ENUMS: Type = Type::seq(&TYPES_ENUM_TYPES);
```
Kryptos-FR, line 2350: **/!\\** I'm introducing an array of enum type in `ImGui_demo` branch (see 3aeb42d48424bfddc3e1ca2dcb689ee6229b8251).

rust/src/blocks/physics/mod.rs
```diff
@@ -30,7 +30,7 @@ lazy_static! {
   static ref SIMULATION_TYPE: Type = {
     let mut t = common_type::object;
     t.details.object = CBTypeInfo_Details_Object {
-      vendorId: 0x73696e6b, // 'sink'
+      vendorId: 0x66726167, // 'frag'
```
Kryptos-FR, line 33: Shouldn't there be a global constant for that value?

rust/src/blocks/hash.rs
```diff
@@ -198,11 +199,95 @@ add_hasher2!(
...
+      fn registerName() -> &'static str {
+        cstr!($name_str)
+      }
+      fn hash() -> u32 {
+        compile_time_crc32::crc32!($hash)
+      }
+      fn name(&mut self) -> &str {
+        $name_str
+      }
+      fn inputTypes(&mut self) -> &std::vec::Vec<Type> {
```
Kryptos-FR, line 226: `help` functions? 😉

Kryptos-FR (commented):
A few remarks.

Cargo.toml
sinkingsugar: indeed, this was an experiment that worked. nice catch.

include/chainblocks.hpp
sinkingsugar: Sounds good, let's also maybe add the 0x hex too, or rather let's use hex instead

include/common_types.hpp
sinkingsugar: I needed it in `(Substrate.Decode` for `nil` values actually

rust/src/types.rs
sinkingsugar: Sadly in rust for now they are re-defined :( hopefully one day we can fix

rust/src/blocks/physics/mod.rs
sinkingsugar: True

src/tests/crypto.edn
sinkingsugar: haha,
nah copied from some substrate web page ;)

include/common_types.hpp
sinkingsugar: It does not really matter honestly.
This enum is used only in Sequence, Table and now Substrate.Decode - the order and value does not matter for those and does not matter for mal either as they get only texual version.

src/core/blocks/core.hpp
sinkingsugar: None is a value, you could have a sequence on nones :)
Also we need to cover all cases anyway.

PR #136, opened 2021-11-09
Some code changes required for the snake game prototype

Kryptos-FR:
* `RadioButton` can work with any type of value (was limited to `int` before)
* Implement `KeyUp and `KeyDown` blocks
* Implement `For` block
* Allow to use a variable (`Max` parameter) with the `RandomInt` and `RandomFloat`  blocks


sinkingsugar:
I really really tried hard not to have a For block until now 😅

If we want one can we try to find a better name though?
Technically Repeat (check parameters) can be used instead and more user friendly with the Repeat->Until pattern. That's why I didn't want For.

Kryptos-FR:
I don't necessarily find `Repeat` to be more user friendly: you can't start say at 10 until 50 for instance. With `Repeat` you have to initialize the value either before the block (and "leaking" one variable outside that block) or inside (and require to use a `Setup` block) which I find more cumbersome and less intuitive.  In contrast, the `For` block gives the current index as input which is very practical. Granted, in the snake example I do end up putting that input into a variable, but there are cases where it could flow directly into the first block of the `Action` parameter.

To me `Repeat` is similar to `while`, and most languages do have both constructs.

examples/imgui-reload/main.clj
```diff
@@ -19,6 +19,7 @@
     (GUI.Window
      :Title "Hot-reloaded window"
      :Width 400 :Height 720 :Pos (int2 0, 0)
+     :Flags [GuiWindowFlags.NoResize, GuiWindowFlags.NoMove, GuiWindowFlags.NoCollapse]
```
sinkingsugar, line 22: Curious, just for testing/coverage?

src/core/blocks/core.cpp
```diff
@@ -639,6 +639,89 @@ struct PrependTo : public XpendTo {
   }
 };
 
+struct ForBlock {
```
sinkingsugar, line 642: We might want to "inline" this block, can you move to the `.hpp` once it's finished?

src/core/blocks/core.cpp
```diff
@@ -639,6 +639,89 @@ struct PrependTo : public XpendTo {
...
+      if (state != CBChainState::Continue)
+        break;
+    }
+
+    return input;
+  }
+
+private:
+  static inline Parameters _params = {
+      {"From", CBCCSTR("TODO!"), {CoreInfo::IntType, CoreInfo::IntVarType}},
```
sinkingsugar, line 716: Todo

src/core/blocks/core.cpp
```diff
@@ -639,6 +639,89 @@ struct PrependTo : public XpendTo {
...
+
+  void warmup(CBContext *context) {
+    _from.warmup(context);
+    _to.warmup(context);
+    _blocks.warmup(context);
+  }
+
+  CBTypeInfo compose(const CBInstanceData &data) {
+    auto dataCopy = data;
+    dataCopy.inputType = CoreInfo::IntType;
```
sinkingsugar, line 691: 👍 
You are now officially a chainblockser 🤣

src/extra/imgui.cpp
```diff
@@ -537,10 +537,7 @@ struct Window : public Base {
   std::string _title;
   bool firstActivation{true};
   Var _pos{}, _width{}, _height{};
-  bool _movable{false};
-  bool _closable{false};
-  bool _resizable{false};
-  bool _showMenuBar{false};
+  ParamVar _flags;
```
sinkingsugar, line 540: Are the defaults good?

src/extra/imgui.cpp
```diff
@@ -862,7 +836,7 @@ template <CBType CT> struct Variable : public Base {
           // we found a variable, make sure it's the right type and mark
           // exposing off
           _exposing = false;
-          if (var.exposedType.basicType != CT) {
+          if (CT != CBType::Any && var.exposedType.basicType != CT) {
             throw CBException("ImGui - Variable: Existing variable type not "
```
sinkingsugar, line 840: Should be `GUI` I must have forgotten when renamed all things

src/extra/imgui.cpp
```diff
@@ -3755,10 +3754,10 @@ struct Disable : public Base {
   void setParam(int index, const CBVar &value) {
     switch (index) {
     case 0:
-      _disable = value;
+      _blocks = value;
       break;
```
sinkingsugar, line 3758: Breaking change, no impact but we should take note somehow of those (even a trello board or whatever would do)

src/extra/inputs.cpp
```diff
@@ -278,41 +279,29 @@ struct Mouse : public Base {
...
 
     const auto hidden = _hidden.get().payload.boolValue;
-    if (hidden && !_isHidden) {
-      SDL_ShowCursor(SDL_DISABLE);
-      _isHidden = true;
-    } else if (!hidden && _isHidden) {
-      SDL_ShowCursor(SDL_ENABLE);
-      _isHidden = false;
+    if (hidden != _isHidden) {
+      SDL_ShowCursor(hidden ? SDL_DISABLE : SDL_ENABLE);
```
sinkingsugar, line 283: lol my code was silly, I must have used copilot or something - I did it in a rush for sure.

src/extra/inputs.cpp
```diff
@@ -406,6 +395,157 @@ template <SDL_EventType EVENT_TYPE> struct MouseUpDown : public Base {
...
+    auto it = _keyMap.find(str);
+    if (it != _keyMap.end())
+      return it->second;
+
+    return SDLK_UNKNOWN;
+  }
+
+private:
+  static inline Parameters _params = {
+      {"Key", CBCCSTR("TODO!"), {{CoreInfo::StringType}}},
```
sinkingsugar, line 482: TODO

src/tests/bgfx.clj
```diff
@@ -311,7 +310,7 @@
                                     (-> (ToBytes)
                                         (GUI.HexViewer)))))
       (GUI.Window "Another window" :Pos (Float2 0.5 0.5) :Width 0.25 :Height 0.25
-                  :AllowMove true
+                  :Flags [GuiWindowFlags.NoResize, GuiWindowFlags.NoCollapse]
```
sinkingsugar, line 313: We need better defaults

src/tests/bgfx_small.clj
```diff
@@ -18,8 +18,8 @@
      :Title "My Window" :Width 400 :Height 200
      :Contents
      ~[(GUI.Window "My ImGui Window"
-                   :Width 400 :Height 200
-                   :Pos (Int2 0 0)
+                   :Width 400 :Height 200 :Pos (Int2 0 0)
+                   :Flags [GuiWindowFlags.NoResize, GuiWindowFlags.NoMove, GuiWindowFlags.NoCollapse]
```
sinkingsugar, line 22: Better defaults

sinkingsugar (changes requested):

examples/imgui-reload/main.clj
Kryptos-FR: I replaced the boolean with flags so this is the equivalent.

src/core/blocks/core.cpp
Kryptos-FR: Yeah I have a bunch of todos and also I didn't implement the `help` functions.

src/extra/imgui.cpp
Kryptos-FR: Good question. The defaults are now "moveable", "resizeable" and "collapsible".

That's because I followed the same flags (and same values) as in ImGui since it's simpler.

If we want different defaults (and for instance invert the logic around move, collapse and resize) then we need to create independent flags and have a conversion function. That implies a lot of if/else potentially which I wanted to avoid.

src/extra/imgui.cpp
Kryptos-FR: Should we start to tag some version?

src/tests/bgfx.clj
Kryptos-FR: That won't help here since you need to define one value anyway. Even if `NoResize`, `NoCollapse` and `NoMove` were defaults, you would still need to write that same line.

src/tests/bgfx_small.clj
Kryptos-FR: Those three could be defaults.

src/extra/imgui.cpp
Kryptos-FR: On second thought, let's go with just better defaults, as long as there are properly documented. On that note, we need to come up with something better when the values of enum are generated in the documentation.

src/core/blocks/core.cpp
Kryptos-FR: What does it change to have it in the header?

I noticed a bunch of blocks are defined there and use different macros to register themselves (the old factory ones). Are those macros necessary when defined in the header or should they also be refactored?

src/extra/imgui.cpp
sinkingsugar: Yes we should and prob create a changelog.md for those changes

src/core/blocks/core.cpp
sinkingsugar: So the header thing is just so that runtime.cpp or whatever actual compilation unit file (.cpp) can 100% inline it as it comes from header.
The old macros are not necessary anymore, the inlining happens in activateBlock:
See for example Const which doesn't use old macro anymore
https://github.com/fragcolor-xyz/chainblocks/blob/c87c6be56d9a0beef3ead89a207ccb5cc1e1b325/src/core/runtime.hpp#L194

This in optimized ASM will end up being a variable access, literally, 2/3 instructions

sinkingsugar: Also seed this, this is how for should be done.
https://github.com/fragcolor-xyz/chainblocks/blob/c87c6be56d9a0beef3ead89a207ccb5cc1e1b325/src/core/runtime.hpp#L238

sinkingsugar (approved):

PR #137, opened 2021-11-21
Graphics - Materials / Automated rendering tests

guusw:
### Material API for shaders
Doc describing the material API: https://md.fragcolor.xyz/CJ99UMacRAmsLj9Xnm2P3w

### Automated graphics tests
Test code inside `src/gfx/tests` to test all graphics/windowing features supported by our engine.
`src/gfx/tests/test_data.cpp/hpp` contains code to load/store per platform/render API snapshots and either compare them against the previous or store a new copy

### Automated graphics tests for asm.js
`emrun` provided with the emscripten SDK provides usefull functionality for running the above tests in a browser and reading back the results.
The test main.cpp and test_data.cpp contain calls to `emrun_copyFileToOutput` that copies the test results and generated test data (screenshots) to a dump_out folder

### Builtin data management
We require some files to always be included in order to run (shader headers, fonts).
For asm.js this is done through `--preload-file`.
For other platforms that can bundle files alongside binaries `src\gfx\paths.hpp` provides some logic to find the correct paths for bundled assets independently of the working directory/executable location.

Previous content that was referenced from the bgfx source folder (fonts/precompiled shaders) have been moved to the `content` folder along with the makefiles. This allows us to keep the bgfx repo synced to upstream
Before removing draft status I plan to remove the precompiled shaders in favour of using the material system.

### Bundled shaderc as a static/shared library
Allows the bgfx shader compiler to run directly from built executables. This removes the dependency on this being built externally and greatly reduces the amount of wrapper code required to work on asm.js when linked statically for example.

There are some issues:
- inputs/outputs temporary files to/from the filesystem
- not multi-threaded (due to preprocessor)
- need to manually sync changes from bgfx shaderc

### TODO
- Replace usage of shaders with materials in current tests
- Update GFX.Model/Draw & GLTF block implementations to use materials
- Add graphics tests to CI runs where possible to run on test runners/virtual framebuffer

##
The main purpose of these changes is the provide a system that allow custom shaders/materials to work together with rendering functionality provided by the engine.

While there are a bunch of changes here I believe it's faster to make these at the same time rather than making some of these changes individually and having to update the tests each time which will later be changed anyway (this includes everything related to GFX.Model/Shader/Draw)

Dependent on `guus/cmake_cleanup` since I needed the project structure.


sinkingsugar:
Looks of interesting stuff, a quick point I really would have appreciated we discussed before executing:
* shaderc as library.. I tried for many days (weeks?) and scrapped it because it makes our binaries huge and it's something that can much easier scaled via opening processes... the current implementation (is already half supporting multi-thread) should just be ported to use edn for easier maintenance (our philosophy should be use chainblocks when possible). It's also not possible to use directly in WASM btw.

To be continued...

src/extra/bgfx.cpp
```diff
@@ -18,3243 +28,2171 @@
...
-      _rheight = canvasH;
-      bgfx::reset(_rwidth, _rheight, BgfxFlags);
-    }
-#endif
+		const auto tnow = std::chrono::high_resolution_clock::now();
+		std::chrono::duration<float> ddt = tnow - _deltaTimer.Start;
+		std::chrono::duration<float> adt = tnow - _absTimer.Start;
+		_deltaTimer.Start = tnow;
+
+		gfx::FrameRenderer frameRenderer(_gfxContext, gfx::FrameInputs(ddt.count(), adt.count(), _frameCounter++, events));
```
sinkingsugar, line 236: Slow down.. you are doing so many allocations here... It's a huge class.

guusw:
> Looks of interesting stuff, a quick point I really would have appreciated we discussed before executing:
> 
>     * shaderc as library.. I tried for many days (weeks?) and scrapped it because it makes our binaries huge and it's something that can much easier scaled via opening processes... the current implementation (is already half supporting multi-thread) should just be ported to use edn for easier maintenance (our philosophy should be use chainblocks when possible). It's also not possible to use directly in WASM btw.
> 
> 
> To be continued...

Yeah as noted there are some issues. The thing is you don't have to include the hlsl/spirv/metal backend along if you're building for asm.js which makes the shader compiler very small. Already have the tests passing using this.

Since we're not using multi-threading shader compilation at the moment it might be worth using this for simplicity (independent builds, separate js files)

Anyways this was mainly to have the graphics code independently testable. It should be possible to use the exiting code for shader compilation when being used within chainblocks.

src/extra/bgfx.cpp
guusw: We can optimize later 😉 
Also this is once per frame

sinkingsugar: This is not the way I wrote this codebase until now. Those kind of optimisations should be very preventive.
It's the reason warmup/cleanup/activate exist. And I hope you embrace it.
The whole reason of having such states, chains and blocks is because those kind of optimisations should be intuitive.
Heap allocations should never happen in activate unless it's a one-off that gets cached. 
I think I preached this to @Kryptos-FR as well 😄

sinkingsugar:
@guusw I never really planned to have a concept of "scene" btw, but rather chain graphs composing the whole final result.
Too many concepts I'm not sure of, we should go back to the drawing board. As everything seems to spawn complexity we don't want right now.

All I wanted was completing PBR shaders and add animations by what we specced out.

guusw:
Closing this since it's the old bgfx implementation.

PR #161, opened 2022-02-21
Fix ios missing -lSystem

guusw:


Kryptos-FR:
Change it to a draft request to make sure it is not merged.

Also if the only reason to create the request was to trigger the CI job, it is not needed. CI jobs can be triggered manually on any branch.

guusw:
That was the main reason, I like having it trigger automatically but I can use manual triggers. @Kryptos-FR

PR #164, opened 2022-03-01
Documentation for blocks Assoc & GUI.NextColumn

satendrapalsingh:
For both blocks
> Add sample code
> Add missing definition
> Modify parameter descriptions

### DIFF
--- a/docs/samples/blocks/General/Assoc.edn
+++ b/docs/samples/blocks/General/Assoc.edn
@@ -1,20 +1,20 @@
-;; Update an array/ vector
-[10 20] = .array
-(Log) ;; prints original .array => [10 20]
-[0 2 1 3] (Assoc .array) ;; index-0 ele => 2, index-1 ele => 3
-.array (Log) ;; prints updated .array => [2 3] 
-.array (Assert.Is [2 3] false)
+;; Update a sequence
+[10 20] = .sequence
+(Log) ;; prints original target sequence => [10 20]
+[0 2 1 3] (Assoc .sequence) ;; input sequence updates target sequence [index-0 ele => 2, index-1 ele => 3]
+.sequence (Log) ;; prints updated target sequence => [2 3] 
+.sequence (Assert.Is [2 3] true)
 
 ;; Update a global-var table (Global = true; table available to all the chains in the same node)
 {:key1 [10 20] :key2 [30 40]} (Set .tableG :Global true)
 (Log) ;; prints original table => {:key1 [10 20] :key2 [30 40]}
-[0 2 1 3] (Assoc .tableG "key1") ;; input seq updates value of key "key1"
+[0 2 1 3] (Assoc .tableG "key1") ;; input sequence updates value of key "key1" in table
 .tableG (Log) ;; prints updated table => {:key1 [2 3] :key2 [30 40]}
-.tableG (Assert.Is {:key1 [2 3] :key2 [30 40]} false)
+.tableG (Assert.Is {:key1 [2 3] :key2 [30 40]} true)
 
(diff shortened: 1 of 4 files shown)


docs/samples/blocks/General/Assoc.edn
```diff
@@ -0,0 +1,20 @@
...
+[0 2 1 3] (Assoc .tableG "key1") ;; input sequence updates value of key "key1" in table
+.tableG (Log) ;; prints updated table => {:key1 [2 3] :key2 [30 40]}
+.tableG (Assert.Is {:key1 [2 3] :key2 [30 40]} true)
+
+;; Update a local-var table (:Global = false, table available to only this chain in the node)
+{:key1 [10 20] :key2 [30 40]} = .table
+(Log) ;; prints original table => {:key1 [10 20] :key2 [30 40]}
+[0 2 1 3] (Assoc .table "key2") ;; input sequence updates value of key "key2" in table
+.table (Log) ;; prints updated table => {:key1 [10 20] :key2 [2 3]}
+.table (Assert.Is {:key1 [10 20] :key2 [2 3]} true)
```
Kryptos-FR, line 20: Make sure to have an empty line at the end of the file. It is a common practice to avoid some conflicts/unnecessary diffs when modifying the file later.

src/core/blocks/core.cpp
```diff
@@ -1006,8 +1006,43 @@ struct Assoc : public VariableBase {
...
+       {CoreInfo::StringOrAnyVar}},
+      {"Key",
+       CBCCSTR("Table key for the value that is to be updated. Parameter "
+               "applicable if target is table."),
+       {CoreInfo::StringStringVarOrNone}},
+      {"Global",
+       CBCCSTR("If the variable is or should be available to all the chains in "
+               "the same node. The default value (false) makes the variable "
+               "local to the "
+               "chain."),
```
Kryptos-FR, line 1021: It should be on a single line.

Maybe run `ninja format` once more, after putting the whole string on a single line first.

src/core/blocks/core.cpp
```diff
@@ -1006,8 +1006,43 @@ struct Assoc : public VariableBase {
...
+  }
+
   static CBTypesInfo inputTypes() { return CoreInfo::AnySeqType; }
+  static CBOptionalString inputHelp() {
+    return CBCCSTR(
+        "Input sequence that identifies element to be updated and "
+        "its new value. The sequence must contain even-number of elements "
+        "with the 1st element in each pair denoting the target index to update "
+        "and the 2nd element holding the update value for the corresponding "
+        "index.");
```
Kryptos-FR, line 1037: Here also it looks like the split is incorrect.

Unfortunately, `ninja format` doesn't seem to rebuild the whole string and only looks at each line length separately. Therefore, you do have to put it back in one line for the split to occur again correctly.

Kryptos-FR (changes requested):
Looks good to me.

A few nit picking remarks.

docs/samples/blocks/General/Assoc.edn
satendrapalsingh: Fixed. Will keep it in mind for future.

src/core/blocks/core.cpp
satendrapalsingh: Oh, I don't know how I messed that up! Fixed.

src/core/blocks/core.cpp
satendrapalsingh: I have moved the last word 'index' into the previous line.

Kryptos-FR (approved):

PR #165, opened 2022-03-01
Formatting rules

guusw:
- Increases column limit to 130
- Set AlignEscapedNewlines: Left (for macros)
- Setup .clang-format to be used when formatting
This closes #103 as well (the limit is 120 there)


Kryptos-FR (commented):
`160` seems a bit too much, I think `140` would be a good compromise.

Files in the `include` folder have not been formatted. And after doing so, it appears `160` is unfortunate since that's exactly the size of the `BasicTypes` enum when inlined in file `include\common_types.hpp` 😅

guusw:
Yeah 160 just about fits half a screen when I split vscode vertically, could be just a bit less

Kryptos-FR (approved):

PR #166, opened 2022-03-01
CI servers fix

Kryptos-FR:
Forces server editions for Windows and macOs since we currently have some build and/or test issues.


Kryptos-FR:
Not conclusive for macOS, will push the change for windows only.

PR #179, opened 2022-03-11
Enhance documentation for existing and undocumented functions

satendrapalsingh:
Add description and code samples for the following functions:
> `defchain`
> `defnode`
> `defloop`
> `run`
> `schedule`

### DIFF
--- a/docs/docs/functions/macros.md
+++ b/docs/docs/functions/macros.md
@@ -140,7 +140,7 @@ A node is a self-contained execution-context and software environment (like a se
 
 Defines a new looped chain.
 
-`(defchain my-loop)` is actually a shorthand for the more verbose looped chain definition.
+`(defloop my-loop)` is actually a shorthand for the more verbose looped chain definition.
 === "Code"
 
     ```clojure linenums="1"
--- a/docs/docs/functions/macros.md
+++ b/docs/docs/functions/macros.md
@@ -95,7 +95,7 @@ Defines a new macro.
 
 Defines a new node on which chains can be scheduled (to run).
 
-A node is a self-contained execution-context and software environment (like a server) that executes the chain code logic. It can run both on the local hardware as well as  peer-to-peer hardware over the network (blockchain). 
+A node is a self-contained execution-context and software environment (like a server) that executes the chain code logic. It can run both on the local hardware as well as peer-to-peer hardware over the network (blockchain). 
 
 === "Code"
 
@@ -111,8 +111,8 @@ A node is a self-contained execution-context and software environment (like a se
     ;; schedule both the chains on this node
     (schedule main chain-hi)
     (schedule main chain-bye)
-    ;; run all the scheduled chains on this node (FPS and iterations apply only to looped chains)
-    (run main 0.02 5)
+    ;; run all the scheduled chains on this node
+    (run main 0.02)
     ```
 === "Result"
 
--- a/docs/docs/functions/macros.md
+++ b/docs/docs/functions/macros.md
@@ -129,10 +129,37 @@ A node is a self-contained execution-context and software environment (like a se
 
 Defines a new chain that is looped.
 
+Such a chain, when invoked (via `run <node>`), executes in a loop at the frequency given by the 2nd parameter of the `run` command and for a number of iterations specified by the 3rd parameter of the `run`command.
+
 === "Code"
 
     ```clojure linenums="1"
+    (defnode main)
+    ;; define a looped chain (chain-hello)
+    (defloop chain-hi
+        (Msg "Hello World!"))
+    ;; define another looped chain (chain-bye)
+    (defloop chain-bye
+        (Msg "Goodbye World"))
+    ;; schedule the looped chains on this node
+    (schedule main chain-hi)
+    (schedule main chain-bye)
+    ;; run all the chains on this node at 50 FPS and for a max 5 iterations
+    (run main 0.02 5)
+    ```
+=== "Result"
 
+    ```
+    [info] [2022-03-07 22:28:54.682] [T-8432] [logging.cpp::94] [chain-hi] Hello World!
+    [info] [2022-03-07 22:28:54.682] [T-8432] [logging.cpp::94] [chain-bye] Goodbye World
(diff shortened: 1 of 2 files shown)


Kryptos-FR:
Somewhere in those pages we should indicate that both `defchain` and `defloop` are shortcuts.

The long form are:
``` clj
;; defchain
(def my-chain
  (Chain "my-chain"
    ;; blocks here
  ))

;; defloop
(def my-loop
  (Chain "my-loop" :Looped
    ;; blocks here
  ))
```

docs/docs/functions/macros.md
```diff
@@ -39,11 +39,34 @@ Defines new blocks that can be inserted into an existing chain.
 
 Defines a new chain that isn't looped.
 
+Such a chain executes only once when invoked (via `run <node>`).
```
Kryptos-FR, line 42: I think we should take the habit of using parentheses since that's the syntax it will end up being in the code.
```suggestion
Such a chain executes only once when invoked (via `(run <node>)`).
```

docs/docs/functions/macros.md
```diff
@@ -39,11 +39,34 @@ Defines new blocks that can be inserted into an existing chain.
...
+    ;; define another non-looped chain (chain-bye)
+    (defchain chain-bye
+        (Msg "Goodbye World"))
+    ;; schedule the non-looped chains on the node
+    (schedule main chain-hello)
+    (schedule main chain-bye)
+    ;; run all the scheduled chains on the node
+    (run main)
+    ;; Even if provided, the FPS and iteration parameters are ignored for non-looped chains
+    ;; (run main 0.02 5)
```
Kryptos-FR, line 61: @sinkingsugar is it correct stating it this way? For instance, if a non-looped chain has a `(Recur)` block, would that be impacted by the delay parameter?

docs/docs/functions/macros.md
```diff
@@ -70,23 +93,74 @@ Defines a new macro.
 
 ## defnode
 
-Defines a new node on which chains can be scheduled.
+Defines a new node on which chains can be scheduled (to run).
+
+A node is a self-contained execution-context and software environment (like a server) that executes the chain code logic. It can run both on the local hardware as well as  peer-to-peer hardware over the network (blockchain).
```
Kryptos-FR, line 98: Nit: one too many space between `as` and `peer-to-peer`.
```suggestion
A node is a self-contained execution-context and software environment (like a server) that executes the chain code logic. It can run both on the local hardware as well as peer-to-peer hardware over the network (blockchain). 
```

docs/docs/functions/macros.md
```diff
@@ -70,23 +93,74 @@ Defines a new macro.
...
+    [info] [2022-03-07 22:14:51.776] [T-14836] [logging.cpp::94] [chain-hi] Hello World!
+    [info] [2022-03-07 22:14:51.791] [T-14836] [logging.cpp::94] [chain-hi] Hello World!
+    [info] [2022-03-07 22:14:51.823] [T-14836] [logging.cpp::94] [chain-hi] Hello World!
     ```
 
 ## defloop
 
 Defines a new chain that is looped.
 
+Such a chain, when invoked (via `run <node>`), executes in a loop at the frequency given by the 2nd parameter of the `run` command and for a number of iterations specified by the 3rd parameter of the `run`command.
```
Kryptos-FR, line 132: Same remark about using parentheses.
```suggestion
Such a chain, when invoked (via `(run <node>)`), executes in a loop at the frequency given by the 2nd parameter of the `(run)` command and for a number of iterations specified by the 3rd parameter of the `(run)` command.
```

docs/docs/functions/misc.md
```diff
@@ -74,6 +74,80 @@ license: CC-BY-SA-4.0
 
 ```
 
+## run
+
+Executes all the chains that have been scheduled on a given node.
+
+`run` takes three arguments: the name of the node, number of seconds to elapse between chain executions, and the maximum number of chain executions allowed.
```
Kryptos-FR, line 81:
```suggestion
`(run)` takes three arguments: the name of the node, number of seconds to elapse between chain executions, and the maximum number of chain executions allowed. 
```

Kryptos-FR (changes requested):

docs/docs/functions/macros.md
sinkingsugar: this is not correct, non looped (main, e.g. defchain) chains still can yield, and so they still can iterate, and this determines the iteration speed/fps

docs/docs/functions/macros.md
satendrapalsingh: Fixed

docs/docs/functions/macros.md
satendrapalsingh: This description of the `run` parameters is moved out of `defchain` and into `run`.

docs/docs/functions/macros.md
satendrapalsingh: Fixed

docs/docs/functions/misc.md
satendrapalsingh: Rewrote this paragraph. Fixed now.

satendrapalsingh:
Documented the fact that `defchain` and `defloop` are shortcuts for the more verbose chain definition functions.

docs/docs/functions/macros.md
```diff
@@ -70,23 +104,87 @@ Defines a new macro.
...
+    [info] [2022-03-07 22:14:51.791] [T-14836] [logging.cpp::94] [chain-hi] Hello World!
+    [info] [2022-03-07 22:14:51.823] [T-14836] [logging.cpp::94] [chain-hi] Hello World!
     ```
 
 ## defloop
 
-Defines a new chain that is looped.
+Defines a new looped chain.
+
+`(defchain my-loop)` is actually a shorthand for the more verbose looped chain definition.
```
Kryptos-FR, line 143: Should be `defloop`
```suggestion
`(defloop my-loop)` is actually a shorthand for the more verbose looped chain definition.
```

Kryptos-FR (changes requested):
A small copy-paste mistake to fix. Otherwise, LGTM.

Kryptos-FR (approved):

PR #240, opened 2022-05-17
Allow to use a context variable to set the port of `Http.Server`

Kryptos-FR:
Make the `:Port`parameter of `Http.Server` accept a context variable.

The rationale is that the port could be computed from previous blocks or that it could be part of a "retry" pattern with a list of possible ports. This will be useful in the Unity plugin to recover from a crashed server (where the socket might not have been released properly) or if two instances of Unity with the plugin are used from the same machine.

```clj
7070 = .port
(Http.Server :Port .port :Handler handler)
```

Documentation has a note to indicate that the port is only set once when the server initializes and if the variable is modified afterwards, the server is not affected.

### DIFF
--- a/docs/details/blocks/Http/Server.md
+++ b/docs/details/blocks/Http/Server.md
@@ -1,0 +1,2 @@
+The `:Port` is only set once, even if given a variable and that variable is changed during the execution of the loop.
+Having a variable is useful if the port to use is the result of a previous computation, otherwise a constant `int` can directly be used.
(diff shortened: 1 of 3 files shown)


sinkingsugar:
What happens if the server is running already and the port number changes?

To me this makes little sense. The chain of the server can be just launched by another chain that computes the port and then injects the parameter before starting the server chain. As a non variable.

I like that in chainblocks we make distinction between variables and non variables because exactly they optimize certain things in the runtime. This is one such case imho.

We might instead add chain launch parameters so that it can be done outside the textual scripting environment.

Kryptos-FR:
As stated in the description and documentation, once started the server is not affected by an update of the variable.

How do you do this non-varia le thing? I tried and couldn't but I do need to select or change the port depending on application logic, which is not possible when the constant is hard coded in the .edn file.

PR #251, opened 2022-05-24
CI workflow split

Kryptos-FR:
**Description**

This PR adds separate workflows for building on `Windows`, `Linux`, `macOS`, `iOS` and `Wasm` platforms.

Those workflows can either be triggered manually, in which case they present the caller a few options such as building in Debug or Release mode ; or they can be called from the main CI workflow.

Eventually they should be callable from other repositories from the same org. There could be some issues with secrets not being shared which we will need to review later.

**Test plan**

- [ ] main workflow is not broken
- [ ] individual workflows can be executed successfully

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -1,0 +1,36 @@
+name: Build (iOS)
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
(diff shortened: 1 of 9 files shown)


sinkingsugar (approved):
Cool stuff

PR #254, opened 2022-05-26
Testing self-hosted actions

sinkingsugar:
Added a separate workflow that runs GPU tests on linux and collects/uploads coverage for it

-Guus

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -54,7 +54,7 @@ jobs:
           echo "::set-output name=run-tests::${{ github.event.inputs.run-tests || inputs.run-tests }}"
           echo "::set-output name=run-extra-tests::${{ github.event.inputs.run-extra-tests || inputs.run-extra-tests }}"
       - name: Checkout shards
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           repository: fragcolor-xyz/shards
           fetch-depth: 1
@@ -104,7 +104,7 @@ jobs:
             echo "::set-output name=submodules::false"
           fi
       - name: Checkout shards
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           repository: fragcolor-xyz/shards
           fetch-depth: 2
@@ -244,13 +244,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout shards
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           repository: fragcolor-xyz/shards
           fetch-depth: 2
           submodules: recursive
       - name: Checkout glTF-Sample-Models
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           repository: KhronosGroup/glTF-Sample-Models
           path: external/glTF-Sample-Models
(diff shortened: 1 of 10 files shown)


.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,101 @@
...
+jobs:
+  #
+  # Build shards for linux
+  #
+  Linux:
+    name: Build (${{ github.event.inputs.build-type || inputs.build-type }})
+    runs-on: self-hosted
+    outputs:
+      run-tests: ${{ steps.setup.outputs.run-tests }}
+      run-extra-tests: ${{ steps.setup.outputs.run-extra-tests }}
```
Kryptos-FR, line 30: Those don't exist. If fact there is no need for job outputs in this workflow, since there is only one.

```suggestion
```

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,101 @@
...
+          repository: fragcolor-xyz/shards
+          fetch-depth: 1
+          submodules: recursive
+      - name: Set up dependencies
+        run: |
+          ./bootstrap
+      - name: Build
+        run: |
+          cmake -Bbuild -G Ninja -DCODE_COVERAGE=1 -DCMAKE_BUILD_TYPE=${{ steps.setup.outputs.build-type }}
+          cmake --build build --target shards test_gfx
```
Kryptos-FR, line 48: Not using `ninja shards`?

.github/workflows/test-linux-gpu.yml
```diff
@@ -0,0 +1,101 @@
...
+          path: |
+            coverage
+            test_gfx.xml
+          if-no-files-found: error
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: shards-linux-gpu rejected test data
+          path: src/gfx/tests/data/default/rejected
+          if-no-files-found: ignore
```
Kryptos-FR, line 101: Missing empty line at end of file.

LautrecOfCarim (approved):

Kryptos-FR (changes requested):

.github/workflows/test-linux-gpu.yml
guusw: I prefer this form since it's agnostic to the generator being used

Kryptos-FR (approved):

.github/workflows/build-linux.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: fragcolor-xyz/shards
-          fetch-depth: 1
+          fetch-depth: 2
```
Kryptos-FR, line 60: Is that necessary? As far as I know, it was only required when doing coverage so that the codecov action can calculate the commit hash.

sinkingsugar: we are doing coverage here

Kryptos-FR: Not on that job, no. Coverage is done on the next job.

guusw: Yeah wrong file, wanted to add it in the gpu test file

guusw:
Also, updated the checkout actions to [v3](https://github.com/actions/checkout/tree/v3) because the older one crashed the self hosted runner & for consistency

guusw:
Codecov comment  didn't update but seems fine on site itself (https://app.codecov.io/gh/fragcolor-xyz/shards/commit/4a338ebc0d3d090bebce41a47adfd01624e63228 , 75% coverage)

PR #255, opened 2022-05-26
CI workflow split (2)

Kryptos-FR:
Following #251, move the tests for each platform into their respective reusable workflow.

This lets manual workflow run decide to run or not the tests. Additionally, this puts everything related to a specific platform into the same file which helps maintenance.


guusw:
I think having run tests default to `true` would be more useful

.github/workflows/build-ios.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           rustup toolchain install nightly
           rustup +nightly target add aarch64-apple-ios
           rustup +nightly target add x86_64-apple-ios
+          rustup default nightly
```
guusw, line 27: Was this needed for something? Doesn't matter too much, although it shouldn't be required.

guusw (approved):
Some questions but looks good

.github/workflows/build-ios.yml
Kryptos-FR: It is required for the rust cache action. See commit d930819d3bb2d975378408f131863abee4631896.

Kryptos-FR:
> I think having run tests default to true would be more useful

@guusw I know that people are lazy so they would unlikely uncheck the boxes when running manually. I'm trying to avoid useless computation and use of machine resource.

PR #259, opened 2022-06-02
Fix and cleanup documentation-build warnings

satendrapalsingh:
> rename asset files (replace 'cb' with 'sh')
- `..\assets\build-cb_acc-env-var.png`

> remove unnecessary asset file and link
- `..\assets\build-cb_use-build-link.png`

> remove unused asset file
- `..\assets\build-cb_run-mingw64.png`

> add information on `cargo clean` to build troubleshooting tips


satendrapalsingh:
@Kryptos-FR I have removed the 'building and linking CBL.exe' screenshot image as I feel it's not very useful.
But if you think we should have it, I'll put it back in (of course it will change as now the shards.exe linking)

Issue #263, opened 2022-06-08
Add missing impl TryFrom<&Var> in Rust


inabe-a:
TryFrom implementations for some types are missing in types.rs

I've already started some of the work here:
https://github.com/fragcolor-xyz/shards/tree/rust_var_as_i32

Ideally we should cover/consolidate as many types as we can and clean up the code if necessary.

Kryptos-FR:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/shards/263?planning-poker) with ZenHub @inabe-a @sinkingsugar

PR #276, opened 2022-06-13
Rust var as i32

inabe-a:
Adding From<T>, TryFrom<T> impl in Rust for Var for vector types with up to 4 elements.

https://github.com/fragcolor-xyz/shards/issues/263

**Has a test plan**

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1230,9 +1230,7 @@ impl From<u64> for Var {
       valueType: SHType_Int,
       ..Default::default()
     };
-    unsafe {
-      res.payload.__bindgen_anon_1.intValue = i64::from_ne_bytes((v).to_ne_bytes());
-    }
+    res.payload.__bindgen_anon_1.intValue = i64::from_ne_bytes((v).to_ne_bytes());
     res
   }
 }
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1706,6 +1706,19 @@ impl TryFrom<&Var> for i64 {
   }
 }
 
+impl TryFrom<&Var> for i32 {
+  type Error = &'static str;
+
+  #[inline(always)]
+  fn try_from(var: &Var) -> Result<Self, Self::Error> {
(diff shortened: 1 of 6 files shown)


sinkingsugar:
`half`/`f16` is interesting but not being "standard" makes it less useful.
Is the rust community mostly converging using `half` crate though? In that case maybe why not.

guusw (changes requested):

inabe-a:
Tested locally, but I haven't enabled `cargo test` here. We probably want to push that separately.
https://discord.com/channels/969754837220737054/969766235967459419/986549166673129512

rust/src/types.rs
```diff
@@ -3280,3 +3290,96 @@ lazy_static! {
...
+
+  test_to_from_vec4!(i16, i16::MAX, "[i16,4] conversion failed");
+  test_to_from_vec4!(i16, i16::MIN, "[i16,4] conversion failed");
+  test_to_from_vec4!(u16, u16::MAX, "[u16,4] conversion failed");
+  test_to_from_vec4!(u16, u16::MIN, "[u16,4] conversion failed");
+  test_to_from_vec4!(half::f16, half::f16::ZERO, "[half::f16,4] conversion failed");
+  test_to_from_vec4!(half::f16, half::f16::MAX, "[half::f16,4] conversion failed");
+  test_to_from_vec4!(half::f16, half::f16::MIN, "[half::f16,4] conversion failed");
+  test_to_from_vec4!(half::f16, half::f16::MIN_POSITIVE, "[half::f16,4] conversion failed");
+  test_to_from_vec4!(half::f16, half::f16::EPSILON, "[half::f16,4] conversion failed");
```
inabe-a, line 3382: Shards is so sharp, it cuts the epsilon in half 🎉

sinkingsugar:
Considering that I'm now propagating errors et al in #279 we should reconsider the overflow handling. We might wanna error on i64 overflow.

Issue #265, opened 2022-06-08
Make sure setParam/getParam on Shards are non-fatal


inabe-a:
Investigate Shards' setParam(...)/getParam(...).

Either:
  - [ ] Confirm and test they're only ever called from try/catch
  - [ ] Make sure they never throw
In either case it might be useful to wrap them in noexcept(false) or noexcept(true)

Rationale:
 - Users must not be able to crash the program by the use of scripts

PR #279, opened 2022-06-16
Refactor Shards error propagation

sinkingsugar:
aka remove some exceptions...

This is the first part of the refactor, eventually we want to remove exceptions **totally** from `activation` calls.
But for now we just wrap activations in a nice landing pad that translates into an Error value type.

### DIFF
--- a/include/dllshard.hpp
+++ b/include/dllshard.hpp
@@ -135,7 +135,7 @@ public:
   }
 
   static SHWireState runShards2(Shards shards, SHContext *context, const SHVar &input, SHVar &output) {
-    return sCore._core->runShards(shards, context, &input, &output);
+    return sCore._core->runShards2(shards, context, &input, &output);
   }
 
   static void log(const char *msg) { sCore._core->log(msg); }
--- a/include/dllshard.hpp
+++ b/include/dllshard.hpp
@@ -130,9 +130,12 @@ public:
     return sCore._core->composeShards(shards, callback, userData, data);
   }
 
-  static SHWireState runShards(Shards shards, SHContext *context, const SHVar &input, SHVar &output,
-                               const bool handleReturn = false) {
-    return sCore._core->runShards(shards, context, &input, &output, handleReturn);
+  static SHWireState runShards(Shards shards, SHContext *context, const SHVar &input, SHVar &output) {
+    return sCore._core->runShards(shards, context, &input, &output);
+  }
+
+  static SHWireState runShards2(Shards shards, SHContext *context, const SHVar &input, SHVar &output) {
+    return sCore._core->runShards(shards, context, &input, &output);
   }
 
   static void log(const char *msg) { sCore._core->log(msg); }
@@ -141,11 +144,18 @@ public:
 
   static const char *rootPath() { return sCore._core->getRootPath(); }
 
-  static SHVar asyncActivate(SHContext *context, std::function<SHVar()> f) {
-    return sCore._core->asyncActivate(context, &f, [](auto ctx, auto data) {
-      auto f = reinterpret_cast<std::function<SHVar()> *>(data);
-      return (*f)();
-    });
+  static SHVar asyncActivate(SHContext *context, std::function<SHVar()> activate, std::function<void()> cancellation) {
+    auto tup = std::make_tuple(activate, cancellation);
+    return sCore._core->asyncActivate(
+        context, &tup,
+        [](auto ctx, auto data) {
+          auto f = reinterpret_cast<decltype(tup) *>(data);
+          return std::get<0>(*f)();
+        },
+        [](auto ctx, auto data) {
+          auto f = reinterpret_cast<decltype(tup) *>(data);
+          return std::get<1>(*f)();
+        });
   }
 
 private:
@@ -332,6 +342,8 @@ public:
 inline void registerShard(const char *fullName, SHShardConstructor constructor, std::string_view _) {
   Core::registerShard(fullName, constructor);
 }
+
+inline void abortWire(SHContext *ctx, const char *msg) { Core::abortWire(ctx, msg); }
 }; // namespace shards
 
 #endif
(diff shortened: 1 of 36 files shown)


rust/src/shard.rs
```diff
@@ -193,10 +201,20 @@ unsafe extern "C" fn shard_destroy<T: Shard>(arg1: *mut CShard) {
...
+      message: core::ptr::null(),
+      code: 0,
+    },
+    Err(error) => {
+      (*blk).error = Some(CString::new(error).expect("CString::new failed"));
+      SHError {
+        message: (*blk).error.as_ref().unwrap().as_ptr(),
+        code: 1,
+      }
+    }
```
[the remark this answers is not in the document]
sinkingsugar: I think it's not generating default for `SHError`, need to check 
And yes styling, I think Match looks better personally.

sinkingsugar:
Why I had no ci???

include/shardwrapper.hpp
```diff
@@ -153,10 +154,17 @@ template <class T> struct ShardWrapper {
 
     // setParam
     if constexpr (has_setParam<T>::value) {
-      result->setParam = static_cast<SHSetParamProc>(
-          [](Shard *b, int i, const SHVar *v) { reinterpret_cast<ShardWrapper<T> *>(b)->shard.setParam(i, *v); });
+      result->setParam = static_cast<SHSetParamProc>([](Shard *b, int i, const SHVar *v) {
```
inabe-a, line 157: Does this already cover set/get params we talked about? https://github.com/fragcolor-xyz/shards/issues/265

sinkingsugar: Yeah I implemented error handling for set/get yep

sinkingsugar:
Critical to merge soon

include/dllshard.hpp
```diff
@@ -130,9 +130,12 @@ class Core {
...
   }
 
-  static SHWireState runShards(Shards shards, SHContext *context, const SHVar &input, SHVar &output,
-                               const bool handleReturn = false) {
-    return sCore._core->runShards(shards, context, &input, &output, handleReturn);
+  static SHWireState runShards(Shards shards, SHContext *context, const SHVar &input, SHVar &output) {
+    return sCore._core->runShards(shards, context, &input, &output);
+  }
+
+  static SHWireState runShards2(Shards shards, SHContext *context, const SHVar &input, SHVar &output) {
```
Kryptos-FR, line 137: Do we need two functions now? They seem to both do the same.

include/shardwrapper.hpp
```diff
@@ -153,10 +154,17 @@ template <class T> struct ShardWrapper {
...
+        try {
+          reinterpret_cast<ShardWrapper<T> *>(b)->shard.setParam(i, *v);
+          return SHError{0};
```
Kryptos-FR, line 160: I feel like for clarity we should have a static `constexpr` like `SHError::Success` to replace all the instances of `SHError{0}`.

include/utility.hpp
```diff
@@ -433,7 +433,11 @@ template <class SH_CORE> class TShardsVar {
   void destroy() {
     for (auto it = _shardsArray.rbegin(); it != _shardsArray.rend(); ++it) {
       auto blk = *it;
-      blk->cleanup(blk);
+      auto errors = blk->cleanup(blk);
+      if (errors.code != 0) {
```
Kryptos-FR, line 437: I think this `0` value should be named for clarity. Maybe something like `NO_ERROR`.

Alternatively it could be an inlined function like `errors.hasError()`.

src/core/shards/flow.cpp
```diff
@@ -451,36 +451,36 @@ struct Maybe : public BaseSubFlow {
   }
 
   SHVar activate(SHContext *context, const SHVar &input) {
-    SHVar output{};
+    SHVar output = input;
```
Kryptos-FR, line 454: Does it clone or copy here?

Kryptos-FR (commented):
Overall LGTM.
A few code suggestions for code quality.

include/dllshard.hpp
sinkingsugar: If you navigate inside they are, one handles return, one does not (useful in Repeat etc flow blocks).
We probably need a comment pass at some point so that every entry point is covered.

include/shardwrapper.hpp
sinkingsugar: True, for now it's fine but once we do the big refactor to remove exceptions might be better to have something more clear.

src/core/shards/flow.cpp
sinkingsugar: simple blit copy since it's basic type `SHVar`
It's just to fix the flow and make it consistent.

include/dllshard.hpp
sinkingsugar: check shards.h, It's commented

include/shardwrapper.hpp
sinkingsugar: Sorted

include/utility.hpp
sinkingsugar: Sorted

rust/src/shard.rs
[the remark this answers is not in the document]
sinkingsugar: Ok sorted that too with default

include/dllshard.hpp
Kryptos-FR: I don't get it. The body of both function is visible here and they both have the exact same line:

```cpp
return sCore._core->runShards(shards, context, &input, &output);
```

So unless the second should call a different function (`sCore._core->runShards2` maybe), they are the same. What am I missing?

sinkingsugar: Oops I see. You were not questioning the functionality 😅
That's a mistake yes ! Thanks 🙏

PR #280, opened 2022-06-16
Docs misc fixes

satendrapalsingh:
> Fix missed Chainblocks-to-Shards-renaming terms
> Fix `Match` shard's code samples


docs/samples/shards/General/Match/3.edn
```diff
@@ -1,10 +1,9 @@
...
- :Passthrough
- true) ;; input allowed to passthrough to Match shard's output
-(Log) ;; print Match shard's output => "false"
+;; multiple matches + nil match at start + with passthrough
+1
+(Match [
+        nil (-> "Matched nil")     ;; case processed, match found
+        1 (-> "Matched 1")         ;; case ignored
+        (+ 0 1) (-> "Matched 1a")  ;; case ignored
+        2 (-> "Matched 2")]        ;; case ignored
```
Kryptos-FR, line 7: For the sample with `:Passthrough true` I think it would be good to log/msg on each case so that the console output also shows which case was matched.

Something like:
```suggestion
        nil (-> (Msg "Matched nil"))     ;; case processed, match found
        1 (-> (Msg "Matched 1"))         ;; case ignored
        (+ 0 1) (-> (Msg "Matched 1a"))  ;; case ignored
        2 (-> (Msg "Matched 2"))]        ;; case ignored
```

docs/samples/shards/General/Match/4.edn
```diff
@@ -1,10 +1,8 @@
...
-  nil (-> (Msg "Any Match is OK"))] ;; nil case matches everything
- :Passthrough
- true)
-(Log) ;; print Match shard's output => "false"
+;; no matches + nil match in the middle + with passthrough
+1
+(Match [
+        2 (-> "Matched 2")         ;; case processed, match not found
+        nil (-> "Matched nil")     ;; case processed, match found
+        3 (-> "Matched 3")]         ;; case ignored
```
Kryptos-FR, line 6: Same remark.

```suggestion
        2 (-> (Msg "Matched 2"))         ;; case processed, match not found
        nil (-> (Msg "Matched nil"))     ;; case processed, match found
        3 (-> (Msg "Matched 3"))]         ;; case ignored
```

Kryptos-FR (changes requested):
Overall LGTM, a few suggestions in the sample where `:Passthrough true` is used.

docs/samples/shards/General/Match/3.edn
satendrapalsingh: Yes, that's a good idea. Fixed (with fixup, rebase, force-push).

Kryptos-FR (approved):

PR #281, opened 2022-06-18
Egui Integration Foundation

sinkingsugar:
Since I'm fixing many things on the way let's open a new PR when we implement all the relative egui Shards @Kryptos-FR

### DIFF
--- a/rust/src/shard.rs
+++ b/rust/src/shard.rs
@@ -39,8 +39,12 @@ use std::ffi::CString;
 use std::os::raw::c_char;
 
 pub trait Shard {
-  fn registerName() -> &'static str;
-  fn hash() -> u32;
+  fn registerName() -> &'static str
+  where
+    Self: Sized;
+  fn hash() -> u32
+  where
+    Self: Sized;
 
   fn name(&mut self) -> &str;
   fn help(&mut self) -> OptionalString {
@@ -71,14 +75,20 @@ pub trait Shard {
     None
   }
 
-  fn hasCompose() -> bool {
+  fn hasCompose() -> bool
+  where
+    Self: Sized,
+  {
     false
   }
   fn compose(&mut self, _data: &InstanceData) -> Result<Type, &str> {
     Ok(Type::default())
   }
 
-  fn hasComposed() -> bool {
+  fn hasComposed() -> bool
+  where
+    Self: Sized,
+  {
     false
   }
   fn composed(&mut self, _wire: &SHWire, _results: &ComposeResult) {}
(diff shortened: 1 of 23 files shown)


Kryptos-FR:
I did a cleanup in another branch. I'm AFK today but I'll clean it up tomorrow. Let's not merge this PR yet.

sinkingsugar:
@guusw CI fails somehow
Also needs errors refactor branch

rust/src/shard.rs
```diff
@@ -39,8 +39,12 @@ use std::ffi::CString;
 use std::os::raw::c_char;
 
 pub trait Shard {
-  fn registerName() -> &'static str;
-  fn hash() -> u32;
+  fn registerName() -> &'static str
+  where
+    Self: Sized;
```
Kryptos-FR, line 44: Wouldn't it be easier to derive `Shard` from `Sized` instead of marking every single function?

```rust
pub trait Shard: Sized {
  // ...
}
```

rust/src/types.rs
```diff
@@ -3240,6 +3458,19 @@ impl PartialEq for Type {
 
 pub const FRAG_CC: i32 = 0x66726167; // 'frag'
 
+pub static INT_TYPES_SLICE: &[Type] = &[common_type::int];
+pub static INT2_TYPES_SLICE: &[Type] = &[common_type::int2];
+pub static FLOAT_TYPES_SLICE: &[Type] = &[common_type::float];
+pub static FLOAT3_TYPES_SLICE: &[Type] = &[common_type::float3];
```
Kryptos-FR, line 3464: Rust question. The fact that we use slices here, does it mean that the compiler will likely allocate a single contiguous chunk of memory (array) for all the static and return a slice reference for a portion of it?

rust/src/shards/gui.edn
```diff
@@ -0,0 +1,23 @@
...
+              :Bottom (-> ...) ; can be None
+              )
+
+  (GUI.Window .this-ui
+              :Title "My Window" ; can be None (window options)
+              :Content (-> ...))
+  
+  ; requires gpu context etc
+  ; but ideally rest of blocks can run from other threads and such
+  (GUI.Render .this-ui)
```
Kryptos-FR, line 21: So technically, the GUI could be a defined in a dedicated wire, and rendered from another one (parent or sibling)?

Kryptos-FR (approved):
LGTM. I think we can merge this "foundation" so that I can xork on each separate egui elements next.

rust/src/shard.rs
sinkingsugar: Wouldn't work apparently cos those calls are static  :)

rust/src/types.rs
sinkingsugar: yep

rust/src/shards/gui.edn
sinkingsugar: yep

PR #282, opened 2022-06-19
minor rust fixes

sinkingsugar:
Try fix depend bot alert
Try remove shardsc.rs


guusw:
I don't think we can remove that patch yet, they still haven't released a new version with emscripten fixes

sinkingsugar:
> I don't think we can remove that patch yet, they still haven't released a new version with emscripten fixes

Indeed, but it's also cos we don't bump our crates (and we can't as too much work yet )

PR #283, opened 2022-06-21
C++ shard parameter macros 

guusw:
Add a set of macros to define parameters quickly.

Usage:
```cpp
class ShardName {
  PARAM_VAR(_varName0, "DisplayName", "Help text", {Types});
  // = Var _varName0;

  PARAM_PARAMVAR(_varName1, "DisplayName", "Help text", {Type::VariableOf(Types)});
  // = ParamVar _varName1; 

  PARAM(ShardsVar, _varName2, "DisplayName", "Help text", {Types});
  // = ShardsVar _varName2;

  PARAM_IMPL(ShardName, PARAM_IMPL_FOR(_varName0), PARAM_IMPL_FOR(_varName1), PARAM_IMPL_FOR(_varName2));
  // implements parameters()/getParam()/setParam()
  // (and defines a static internal getIterableParams() that can be used to iterate over parameters)

  // Automatically calls warmup/cleanup for parameters that implement it (e.g. ParamVar/ShardsVar)
  void warmup(SHContext *context) { PARAM_WARMUP(context); }
  void cleanup() { PARAM_CLEANUP(); }
};
```

Also adding some additional functionality to Var to use alongside it:
- missing conversion to `const char*`
- check for `isNone` & `isSet` (it's inverse)

### DIFF
--- a/include/shards.hpp
+++ b/include/shards.hpp
@@ -553,7 +553,6 @@ struct Var : public SHVar {
   Var& operator=(const SHVar& other) { memcpy((void *)this, (void *)&other, sizeof(SHVar)); return *this; }
 
   bool isNone() const { return valueType == SHType::None; }
-  bool isSet() const { return !isNone(); }
 
   explicit operator bool() const {
     if (valueType != Bool) {
--- a/include/shards.hpp
+++ b/include/shards.hpp
@@ -550,6 +550,11 @@ struct Var : public SHVar {
 
   explicit Var(const SHVar &other) { memcpy((void *)this, (void *)&other, sizeof(SHVar)); }
 
+  Var& operator=(const SHVar& other) { memcpy((void *)this, (void *)&other, sizeof(SHVar)); return *this; }
+
+  bool isNone() const { return valueType == SHType::None; }
+  bool isSet() const { return !isNone(); }
+
   explicit operator bool() const {
     if (valueType != Bool) {
       throw InvalidVarTypeError("Invalid variable casting! expected Bool");
@@ -609,6 +614,14 @@ struct Var : public SHVar {
       return static_cast<double>(payload.intValue);
     } else {
       throw InvalidVarTypeError("Invalid variable casting! expected Float");
+    }
+  }
+
+  explicit operator const char *() const {
+    if (valueType == SHType::String || valueType == SHType::Path || valueType == SHType::ContextVar) {
+      return payload.stringValue;
+    } else {
+      throw InvalidVarTypeError("Invalid variable casting! expected String/Path/ContextVar");
     }
   }
 
--- a/src/core/params.hpp
+++ b/src/core/params.hpp
@@ -1,0 +1,138 @@
+#ifndef B0328A63_0B69_4191_94D5_38783B9F20C9
+#define B0328A63_0B69_4191_94D5_38783B9F20C9
+
+#include "stddef.h"
+#include <foundation.hpp>
+#include <type_traits>
+#include "shardwrapper.hpp"
+
+// Template helpers for setParam/getParam
+namespace shards {
+
+#define PARAM(_type, _name, _displayName, _help, _types)                                   \
+  static inline ParameterInfo _name##ParameterInfo = {_displayName, SHCCSTR(_help), _types}; \
+  _type _name;
+
+#define PARAM_VAR(_name, _displayName, _help, _types) PARAM(Var, _name, _displayName, _help, _types)
+#define PARAM_PARAMVAR(_name, _displayName, _help, _types) PARAM(ParamVar, _name, _displayName, _help, _types)
+
+struct IterableParam {
+  size_t offset;
+  const ParameterInfo *paramInfo;
+
+  void (*setParam)(void *varPtr, SHVar var){};
+  SHVar (*getParam)(void *varPtr){};
+  void (*warmup)(void *varPtr, SHContext *ctx){};
+  void (*cleanup)(void *varPtr){};
+
+  void *getVarPtr(void *obj) const { return (void *)(size_t(obj) + offset); }
+  template <typename T> T &get(void *obj) const { return *(T *)getVarPtr(obj); }
+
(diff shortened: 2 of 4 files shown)


src/core/params.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+// Implements setParam()/getParam()
+#define PARAM_GET_SET()                                             \
+  void setParam(int index, const SHVar &value) {                    \
+    size_t numParams;                                               \
+    const IterableParam *params = getIterableParams(numParams);     \
+    if (index < numParams) {                                        \
+      params[index].setParam(params[index].getVarPtr(this), value); \
+    } else {                                                        \
+      throw std::logic_error("Param index out of range");           \
```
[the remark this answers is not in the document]
guusw: InvalidParameterIndex I assume? let me replace that.

src/core/params.hpp
```diff
@@ -0,0 +1,138 @@
...
+    static SHParametersInfo result = []() {                                     \
+      SHParametersInfo result{};                                                \
+      size_t numParams;                                                         \
+      const IterableParam *params = getIterableParams(numParams);               \
+      arrayResize(result, numParams);                                           \
+      for (size_t i = 0; i < numParams; i++) {                                  \
+        result.elements[i] = *const_cast<ParameterInfo *>(params[i].paramInfo); \
+      }                                                                         \
+      return result;                                                            \
+    }();                                                                        \
```
Kryptos-FR, line 91: C++ question: does using a closure like this still make it statically allocated (i.e. resolved at compile time) or dynamically allocated? In the later case, does it mean that the resulting binary is potentially smaller since memory will only be determined at runtime?

src/extra/gfx/gltf.cpp
```diff
@@ -19,13 +20,9 @@ struct GLTFShard {
...
 
-  static SHParametersInfo parameters() {
-    static Parameters params{
-        {"Transform", SHCCSTR("The transform variable to use"), {TransformVarType}},
-        {"Path", SHCCSTR("The static path to load a model from during warmup"), {CoreInfo::StringType}},
-    };
-    return params;
-  }
+  PARAM_PARAMVAR(_transformVar, "Transform", "The transform variable to use", {TransformVarType});
+  PARAM_VAR(_staticModelPath, "Path", "The static path to load a model from during warmup", {CoreInfo::StringType});
```
Kryptos-FR, line 24: This used to be a simple `std::string` and is now a `Var`. What is the impact? SHGouldn't we have a macro to still allow the direct usage of `std::string`? I think they were performance reason for doing so. @sinkingsugar

src/core/params.hpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+// implements cleanup() for parameters
+// call from cleanup manually
+#define PARAM_CLEANUP()                                         \
+  {                                                             \
+    size_t numParams;                                           \
+    const IterableParam *params = getIterableParams(numParams); \
+    for (size_t i = 0; i < numParams; i++)                      \
+      params[i].cleanup(params[i].getVarPtr(this));             \
```
Kryptos-FR, line 133: `cleanup` should probably use the reverse order. Not sure it is really needed but it's kind of a convention to "destroy" in reverse "creation" order.

src/core/params.hpp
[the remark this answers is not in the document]
sinkingsugar: Yeah

include/shards.hpp
```diff
@@ -550,6 +550,11 @@ struct Var : public SHVar {
 
   explicit Var(const SHVar &other) { memcpy((void *)this, (void *)&other, sizeof(SHVar)); }
 
+  Var& operator=(const SHVar& other) { memcpy((void *)this, (void *)&other, sizeof(SHVar)); return *this; }
+
+  bool isNone() const { return valueType == SHType::None; }
+  bool isSet() const { return !isNone(); }
```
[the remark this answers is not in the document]
sinkingsugar: I think so too yeah

src/core/params.hpp
sinkingsugar: There are no allocations at all.
Compiler will create a struct with the captured things, copy them inside and feed them to a call.

sinkingsugar: Moreover nothing is captured here

src/core/params.hpp
sinkingsugar: correct, cleanup is safer in reverse order definitely

sinkingsugar: Remind me to fix ShardsVar in rust like that too :)

src/extra/gfx/gltf.cpp
sinkingsugar: Not sure I get it, where did we use std::string?

sinkingsugar: Oh, ParamVar clones so Var is good

src/core/params.hpp
guusw: Oh, I think you meant using the call like that. This initializes it at runtime, but still only once (lazy initialization)

Kryptos-FR: Yes my question was regarding `static SHParametersInfo result` allocation.

src/core/params.hpp
guusw: Can't imagine a case where it matters for params but I can add it for consistency's sake

src/extra/gfx/gltf.cpp
Kryptos-FR: > Not sure I get it, where did we use std::string?

About everywhere in `imgui.cpp`

src/core/params.hpp
sinkingsugar: True but could be

src/extra/gfx/gltf.cpp
sinkingsugar: Yeah I got that after, but as I said its cloning so fine

sinkingsugar (approved):
👍👍

Issue #219, opened 2022-04-20
Basic gizmo shards


guusw:
Add gizmo drawing shards to be able to draw:
- [x] Axis lines
- [x] Move gizmo

guusw:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/chainblocks/219?planning-poker) with ZenHub @Kryptos-FR @sinkingsugar

PR #284, opened 2022-06-21
Editor helpers

guusw:
This adds several editor grouped under the term 'Helpers':
- Gizmo input & rendering
- Line & point based shapes

This completes #219 

handle


sinkingsugar:
super, how can I try it? can you point me to some .edn code?

guusw:
You can run [src/tests/gfx_helpers.edn](https://github.com/fragcolor-xyz/shards/pull/284/files#diff-c01ad6f0266d48eb6e18338f7b208694a1a93c20be8addc983a387ffd7ba1088) for the script from the GIF @sinkingsugar

src/extra/helpers/helpers.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    return _shardsOutput;
+  }
+};
+
+extern void registerHighlightShards();
+extern void registerGizmoShards();
+extern void registerShapeShards();
+void registerShards() {
+  REGISTER_SHARD("Helpers.Context", HelperContextShard);
```
sinkingsugar, line 133: Not sure on this, anyway the naming is confusing.

Kryptos-FR:
Can we merge #336 first? Since this branch is rebased on top of it, it makes reviewing harder with the unrelated changes.

guusw:
Yeah that was the idea

src/extra/helpers/helpers.cpp
sinkingsugar: I think we should do `Helpers` -> `Gizmos`, they are all gizmos by definition. An "unnecessary" tool (when the product is shipped).
This allows you to remove the `Gizmo` suffix you used at times.

sinkingsugar (approved):
LGTM Added a comment about naming.
I think it should be `Gizmos.*`

src/extra/helpers/helpers.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include <gfx/helpers/wireframe.hpp>
+
+namespace shards {
+// Shards for rendering visual helpers
+// Gizmos, grids, lines, etc.
+namespace Helpers {
+
+struct Context {
+  static constexpr uint32_t TypeId = 'HLPc';
+  static inline shards::Type Type{{SHType::Object, {.object = {.vendorId = gfx::VendorId, .typeId = TypeId}}}};
```
Kryptos-FR, line 18: Could probably use the new helper `Type::Object`.
```suggestion
  static inline shards::Type Type::Object(gfx::VendorId, TypeId);
```

src/extra/helpers/highlight.cpp
```diff
@@ -0,0 +1,59 @@
...
+      OVERRIDE_ACTIVATE(data, activateSingle);
+    }
+    return CoreInfo::NoneType;
+  }
+
+  SHVar activateSingle(SHContext *shContext, const SHVar &input) {
+    Context &helperContext = getContext();
+
+    SHTypeInfo inputType{.basicType = SHType::Object,
+                         .object = {.vendorId = input.payload.objectVendorId, .typeId = input.payload.objectTypeId}};
```
Kryptos-FR, line 31: Could probably use the new helper Type::Object.

```suggestion
    auto inputType = Type::Object(input.payload.objectVendorId, input.payload.objectTypeId);
```

src/gfx/math.hpp
```diff
@@ -40,6 +40,11 @@ template <size_t Alignment> inline constexpr size_t alignTo(size_t size) {
   }
 }
 
+template <typename T> inline bool isWithinRange(T val, T target, T tolerance = 0.05f) {
+  T delta = std::abs(val - target);
+  return delta > -tolerance && delta < tolerance;
```
Kryptos-FR, line 45: Is that to allow a negative tolerance? Wouldn't it be better to assert that tolerance is always strictly positive (and even different from zero).

src/gfx/math.hpp
```diff
@@ -40,6 +40,11 @@ template <size_t Alignment> inline constexpr size_t alignTo(size_t size) {
   }
 }
 
+template <typename T> inline bool isWithinRange(T val, T target, T tolerance = 0.05f) {
```
Kryptos-FR, line 43: Isn't that a big default? Usually I have seen epsilon be around 1E-3 or lower.

Kryptos-FR (changes requested):
Some minor changes.

But  also get an error after running the `gfx_helpers.edn` script (in both *release* and *debug*):
> [error] [2022-08-24 10:55:12.599] [T-33716] [helpers.cpp::50] Found 11 dangling reference(s) to Helpers.Context

src/tests/gfx_helpers.edn
```diff
@@ -0,0 +1,76 @@
+(def timestep (/ 1.0 120.0))
```
Kryptos-FR, line 1: Might want to also rename the file `gfx-helpers.edn` to follow the naming of the other tests script.

Same with `gfx-highlight.edn`.

src/gfx/math.hpp
guusw: It's okay for the current usage, renamed it to justify the larger default epsilon

src/gfx/math.hpp
guusw: Overlooked that, removed the abs

guusw: Removed the abs

guusw:
> Some minor changes.
> 
> But also get an error after running the `gfx_helpers.edn` script (in both _release_ and _debug_):
> 
> > [error] [2022-08-24 10:55:12.599] [T-33716] [helpers.cpp::50] Found 11 dangling reference(s) to Helpers.Context

It's a bit weird because of how I'm setting the context variable and it's used within the `:Content` shards, similar to how UI works I believe.

It actually doesn't make sense to check the references in cleanup once there are multiple contexts, since the ref count will only be 1 once the last context is being cleaned up.
This is because they all reference the same variable but the value changes based on which is currently being run.

I removed the reference check but made sure to set before & clear after for activate/warmup/cleanup so that it's not accidentally used when it goes out of scope

sinkingsugar:
> Some minor changes.
> 
> But also get an error after running the `gfx_helpers.edn` script (in both _release_ and _debug_):
> 
> > [error] [2022-08-24 10:55:12.599] [T-33716] [helpers.cpp::50] Found 11 dangling reference(s) to Helpers.Context

That's very bad, @guusw

guusw:
@sinkingsugar The warning was removed, check my last comment about it

src/extra/gizmos/context.cpp
```diff
@@ -29,25 +30,24 @@ struct HelperContextShard : public gfx::BaseConsumer {
...
+    {
+      ObjectVarScope _scope(*_contextVarRef, &_context, Context::Type);
+      PARAM_CLEANUP();
+    }
 
-    PARAM_CLEANUP();
+    baseConsumerCleanup();
 
     if (_contextVarRef) {
-      if (_contextVarRef->refcount > 1) {
```
sinkingsugar: this is not ideal, we already have leaks which I don't know you fixed yet or not. Generally these red flags are the culprit...

sinkingsugar: Can you instead keep this and fixup refcount when you really know what is going on behind the scene? We do that in (Ref) for example

sinkingsugar (changes requested):

src/extra/gizmos/context.cpp
sinkingsugar: To be clear, those checks exits to check external behaviors not your own code which you know. but they are there to check they interact well with all the rest.

guusw: @sinkingsugar I'm not sure what you mean with fixing up refcount.
In case there's 2 of these contexts, they'll both reference the same variable. I can't really know when the variable is no longer used by either.

sinkingsugar (approved):

src/extra/gizmos/context.cpp
sinkingsugar: We already have leaks for months and these are red flags as well. I don't have the time to check sadly in depth. So I trust you do the right thing.

src/extra/gizmos/context.cpp
```diff
@@ -0,0 +1,138 @@
...
+  bool _mouseButtonState{};
+
+  void warmup(SHContext *context) {
+    baseConsumerWarmup(context, true);
+
+    // Reference context variable
+    _contextVarRef = referenceVariable(context, Context::contextVarName);
+
+    {
+      ObjectVarScope _scope(*_contextVarRef, &_context, Context::Type);
```
sinkingsugar, line 37: This is weird what is it?

src/extra/gizmos/context.cpp
```diff
@@ -0,0 +1,138 @@
...
+      PARAM_WARMUP(context);
+    }
+  }
+
+  void cleanup() {
+    {
+      ObjectVarScope _scope(*_contextVarRef, &_context, Context::Type);
```
sinkingsugar, line 44: This is weird what is it?

guusw:
@guusw  Fix nested contexts

src/extra/gizmos/context.cpp
guusw: Replaced this scope thing with just a function with an inner callback.
I like that structure better for this.

Also fixed the scenario where they happen to be nested multiple levels

Kryptos-FR (approved):

src/extra/CMakeLists.txt
```diff
@@ -9,11 +9,15 @@ set(extra_SOURCES
   gfx/shader/translator.cpp
```
inabe-a, line 9: Prefer target_sources (https://cmake.org/cmake/help/latest/command/target_sources.html) instead of SET.
Have a look at Modern CMake

src/extra/gfx/camera.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include <gfx/math.hpp>
+#include <gfx/window.hpp>
+#include <gfx/fmt.hpp>
+#include "../gfx.hpp"
+
+using namespace shards;
+
+namespace gfx {
+
+struct FreeCameraShard : public gfx::BaseConsumer {
```
inabe-a, line 13: The name is unclear. FreeFloating might be a good fit, but perhaps there is a better name, preferably another single word to use.

src/extra/gfx/camera.cpp
```diff
@@ -0,0 +1,187 @@
...
+             PARAM_IMPL_FOR(_lookSpeed));
+
+  struct InputState {
+    struct Keys {
+      bool left{};
+      bool right{};
+      bool forward{};
+      bool backward{};
+    } keys;
+    struct Look {
```
inabe-a, line 36: All structs and classes with non-trivial behavior should have a comment explaining their function.

src/extra/gfx/camera.cpp
```diff
@@ -0,0 +1,187 @@
...
+      if (_input.keys.left)
+        wasdVec.x += -1.0f;
+      if (_input.keys.right)
+        wasdVec.x += 1.0f;
+      if (_input.keys.forward)
+        wasdVec.y += 1.0f;
+      if (_input.keys.backward)
+        wasdVec.y += -1.0f;
+
+      // TODO
```
inabe-a, line 146: Unclear comment - explain what needs to be done.

src/gfx/mesh_utils.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef DC6B12F5_63B6_4FBD_BC92_DFF99E0E3323
```
inabe-a, line 1: Where do these names come from?

src/gfx/math.hpp
```diff
@@ -40,6 +40,11 @@ template <size_t Alignment> inline constexpr size_t alignTo(size_t size) {
   }
 }
 
+template <typename T> inline bool isRoughlyEqual(T val, T target, T tolerance = 0.05f) {
+  T delta = val - target;
+  return delta > -tolerance && delta < tolerance;
```
inabe-a, line 45: Epsilon is preferred esp. since you are already using delta.

src/gfx/math.hpp
```diff
@@ -40,6 +40,11 @@ template <size_t Alignment> inline constexpr size_t alignTo(size_t size) {
   }
 }
 
+template <typename T> inline bool isRoughlyEqual(T val, T target, T tolerance = 0.05f) {
```
inabe-a, line 43: typename std::enable_if<std::is_floating_point<T>::value, or another number type where you can assign 0.05f; or remove the default value

src/gfx/linalg.hpp
```diff
@@ -14,6 +15,66 @@ inline void packFloat4x4(const float4x4 &mat, float *dst) {
...
+  float3 x = transform.x.xyz();
+  float3 y = transform.y.xyz();
+  float3 z = transform.z.xyz();
+  scale.x = linalg::length(x);
+  scale.y = linalg::length(y);
+  scale.z = linalg::length(z);
+  rotationMatrix = float3x3(x / scale.x, y / scale.y, z / scale.z);
+}
+
+inline float3 generateTangent(float3 direction) {
```
inabe-a, line 60: GetTangent(...)
Generate suggests it's cached somewhere

src/gfx/linalg.hpp
```diff
@@ -14,6 +15,66 @@ inline void packFloat4x4(const float4x4 &mat, float *dst) {
...
+    outDistance = linalg::dot(toPlane, planeNormal) / denom;
+    return (outDistance >= 0);
+  }
+
+  return false;
+}
+
+inline float3 extractTranslation(const float4x4 &transform) { return transform.w.xyz(); }
+
+inline void decomposeTRS(const float4x4 &transform, float3 &translation, float3 scale, float3x3 &rotationMatrix) {
```
inabe-a, line 49: scale is passed by value

src/gfx/linalg.hpp
```diff
@@ -14,6 +15,66 @@ inline void packFloat4x4(const float4x4 &mat, float *dst) {
...
+  float3 tMax = (boxMax - rayOrigin) / rayDir;
+  float3 t1 = linalg::min(tMin, tMax);
+  float3 t2 = linalg::max(tMin, tMax);
+  float tNear = linalg::max(linalg::max(t1.x, t1.y), t1.z);
+  float tFar = linalg::min(linalg::min(t2.x, t2.y), t2.z);
+  return float2(tNear, tFar);
+};
+
+// Compute intersection of a ray with a plane
+// Returns distance in t
```
inabe-a, line 32: t is not a parameter in the argument list nor a unit. If the intention is to return outDistance in the range of [0, 1] explain that. Or if t can have values above 1 explain that outDistance is the distance between rayOrigin and the point of intersection.

Other formal parameters need to be explained too. planePoint - any point which lies on the plane. The user of the function must be able to determine how it works from the signature.

src/gfx/linalg.hpp
```diff
@@ -14,6 +15,66 @@ inline void packFloat4x4(const float4x4 &mat, float *dst) {
   }
 }
 
+// compute the near and far intersections of the cube (stored in the x and y components) using the slab method
+// no intersection means vec.x > vec.y (really tNear > tFar)
+inline float2 intersectAABB(const float3 &rayOrigin, const float3 &rayDir, float3 boxMin, float3 boxMax) {
```
inabe-a, line 20: The parameters qualifiers are inconsistent, some are passed by const reference and some by value, but they all have the same type?

src/gfx/math.hpp
Kryptos-FR: Not sure if you are talking about renaming `tolerance`or `delta. Most likely `tolerance`.

I would agree if the value used was very small and indeed an epsilon (i.e. a value that adding or subtracting doesn't produce a noticeable difference). but I think here it is a rough comparison which might have noticeable difference, so IMO `tolerance` is more appropriate.

src/gfx/linalg.hpp
Kryptos-FR: I would argue the exact opposite (as I come from a .NET background where there are properties with getters and setters).

Maybe we can find a middle ground: `inline float3 retrieveTangent(float3 direction)`?

src/gfx/helpers/translation_gizmo.hpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  void render(InputContext &inputContext, GizmoRenderer &renderer) {
+    for (size_t i = 0; i < 3; i++) {
+      auto &handle = handles[i];
+
+      // Debug draw
+      float4 color = float4(.7, .7, .7, 1.);
+      uint32_t thickness = 1;
+      bool hovered = inputContext.hovered && inputContext.hovered == &handle;
```
inabe-a, line 88: Nit: hovering? Unless the action has occurred and concluded. If the user hovers above the control for 2 seconds and moves away is the debug draw still on?

src/gfx/helpers/gizmos.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+  for (auto &gizmo : gizmos) {
+    gizmo->render(input, renderer);
+  }
+  gizmos.clear();
+
+  renderer.end(drawQueue);
+}
+} // namespace gizmos
+} // namespace gfx
```
inabe-a, line 33: Git helps too :)

sinkingsugar:
@guusw did you address @inabe-a comments , at least partially some are important like commenting what some of the structs do :)

guusw:
I'll go over them

src/extra/CMakeLists.txt
guusw: This shouldn't be confused with global sources, this is just a variable passed to `add_library` containing the source files.
I just like to group them in a separate variable usually for clarity

src/extra/gfx/camera.cpp
guusw: MovableCamera maybe?

src/extra/gfx/camera.cpp
```diff
@@ -0,0 +1,187 @@
...
+  PARAM_PARAMVAR(_panSpeed, "PanSpeed", "Controls middle mouse pan speed",
+                 {CoreInfo::NoneType, CoreInfo::FloatType, CoreInfo::FloatVarType});
+  PARAM_PARAMVAR(_lookSpeed, "LookSpeed", "Controls right mouse look speed",
+                 {CoreInfo::NoneType, CoreInfo::FloatType, CoreInfo::FloatVarType});
+  PARAM_IMPL(FreeCameraShard, PARAM_IMPL_FOR(_flySpeed), PARAM_IMPL_FOR(_scrollSpeed), PARAM_IMPL_FOR(_panSpeed),
+             PARAM_IMPL_FOR(_lookSpeed));
+
+  struct InputState {
+    struct Keys {
+      bool left{};
```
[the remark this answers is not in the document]
guusw: That's true, I'll change that.

src/gfx/mesh_utils.hpp
guusw: I'm using this extension https://marketplace.visualstudio.com/items?itemName=akiramiyakoda.cppincludeguard

src/gfx/math.hpp
guusw: That was my reasoning for using tolerance yes ^

src/gfx/linalg.hpp
guusw: Good catch!

src/extra/gfx/camera.cpp
guusw: Fixed the issue while moving some things around here

src/gfx/linalg.hpp
guusw: Fixed

src/gfx/linalg.hpp
guusw: Fixed comment

src/extra/gizmos/shapes.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+void registerShapeShards() {
+  REGISTER_SHARD("Gizmos.Line", LineShard);
+  REGISTER_SHARD("Gizmos.Circle", CircleShard);
+  REGISTER_SHARD("Gizmos.Rect", RectShard);
+  REGISTER_SHARD("Gizmos.Box", BoxShard);
+  REGISTER_SHARD("Gizmos.Point", PointShard);
+}
+} // namespace Gizmos
+} // namespace shards
```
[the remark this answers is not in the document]
guusw: oh, line endings?

PR #287, opened 2022-06-23
Expose draw queues & various graphics fixes

guusw:
- Some changes that expose draw queues to be used by shards.
This makes it possible to send them to separate pipelines.

- Add glTF shard `Copy` source to duplicate an existing instance.

- Fix a double free in derivedTypeInfo


src/extra/gfx.cpp
```diff
@@ -21,11 +21,13 @@ struct DrawablePassShard {
 
   static inline Parameters params{
       {"Features", SHCCSTR("Features to use."), {Type::VariableOf(Types::PipelineStepSeq), Types::PipelineStepSeq}},
+      {"Queue", SHCCSTR("The queue to draw from (optional). Uses the default queue if not specified"), {Types::DrawQueue, Type::VariableOf(Types::DrawQueue)}},
```
Kryptos-FR, line 24: If it's optional, shouldn't it include `CoreInfo::NoneType` as a valid input type?

src/extra/gfx/drawable.cpp
```diff
@@ -178,18 +178,63 @@ struct DrawableShard {
...
   static inline shards::Types SingleDrawableTypes = shards::Types{Types::Drawable, Types::DrawableHierarchy};
   static inline Type DrawableSeqType = Type::SeqOf(SingleDrawableTypes);
   static inline shards::Types DrawableTypes{Types::Drawable, Types::DrawableHierarchy, DrawableSeqType};
 
   static SHTypesInfo inputTypes() { return DrawableTypes; }
   static SHTypesInfo outputTypes() { return CoreInfo::AnyType; }
-  static SHParametersInfo parameters() { return SHParametersInfo(); }
+  static SHParametersInfo parameters() {
+    static Parameters params{
+        {"Queue", SHCCSTR("The queue to add the draw command to"), {Types::DrawQueue, Type::VariableOf(Types::DrawQueue)}},
```
Kryptos-FR, line 191: `CoreInfo::NoneType`

src/core/runtime.cpp
```diff
@@ -1511,7 +1511,7 @@ SHTypeInfo deriveTypeInfo(const SHVar &value, const SHInstanceData &data, bool *
       const auto varName = value.payload.stringValue;
       for (auto info : data.shared) {
         if (strcmp(info.name, varName) == 0) {
-          return info.exposedType;
+          return cloneTypeInfo(info.exposedType);
```
sinkingsugar, line 1514: is this freed? we were actually not cloning though?

src/extra/gfx/drawable.cpp
```diff
@@ -234,8 +285,70 @@ struct DrawShard : public BaseConsumer {
   SHVar activate(SHContext *shContext, const SHVar &input) { throw ActivationError("GFX.Draw: Unsupported input type"); }
 };
 
+struct DrawQueueShard {
+  static SHTypesInfo inputTypes() { return CoreInfo::AnyType; }
```
Kryptos-FR, line 289: If the input isn't used and doesn't pass through, I think we should define it as `CoreInfo::NoneType` instead of `Any`.

src/gfx/context.cpp
```diff
@@ -153,7 +153,10 @@ struct ContextMainOutput {
 
   void resizeSwapchain(WGPUDevice device, WGPUAdapter adapter, const int2 &newSize) {
     WGPUTextureFormat preferredFormat = wgpuSurfaceGetPreferredFormat(wgpuWindowSurface, adapter);
-    preferredFormat = WGPUTextureFormat_BGRA8Unorm;
+
+    // Force the backbuffer to srgb format so we don't have to convert manually in shader
+    preferredFormat = WGPUTextureFormat_BGRA8UnormSrgb;
```
sinkingsugar, line 158: That's a hard choice.
We all good with this @guusw @inabe-a @Kryptos-FR ?
Afaik conversion in the shader leaves everything very flexible and the cost is tiny. But this was the case mostly with bgfx supporting opengl and dx9... with webgpu we can just forget that and use Srgb.

sinkingsugar (approved):
🙏👍

src/gfx/shader/generator.cpp
```diff
@@ -37,39 +37,91 @@ void GeneratorContext::writeGlobal(const char *name, const FieldType &type) {
...
   result += fmt::format("{}.{}", globalsVariableName, name);
 }
 
-bool GeneratorContext::hasInput(const char *name) { return inputs.find(name) != inputs.end(); }
+bool GeneratorContext::hasInput(const char *name) {
+  if (inputs.find(name) != inputs.end())
+    return true;
+
+  return false;
+}
```
Kryptos-FR, line 49: Previous code seemed correct and easier to read, no?

src/extra/gfx.cpp
guusw: This is about the case where you pass a variable that is not set, right?
I guess there is a case where you dynamically set the variable from a branch ,can add it.

src/extra/gfx/drawable.cpp
guusw: Doesn't that break if you have any previously unconsumed output?
Say you do some math and then you try to use this shard, It'll complain that input is not none?

guusw: @sinkingsugar

src/extra/gfx.cpp
Kryptos-FR: That's the case where you don't pass anything (don't even set the parameter) or when you set it to `nil`. I'm not sure if it is enforced (that if None is not part of the accepted type and the parameter is not then the shard validation fails. But maybe it should be. @sinkingsugar).

guusw: It already kind of works without since it can only be a ContextVariable, ignore the plain Types::DrawQueue, that was a mistake

guusw: Yeah I thought they were always optional, that seems to be how it works currently (not checked if they're missing or not)

inabe-a:
I'll review this on the weekend if it's ok!

src/extra/gfx.cpp
sinkingsugar: From the UX perspective if it's not populated the chain won't run I guess. It's better to have None and explicit fail during compose if still None though. It's something we will have to stabilize once we build a non-code editor anyway.

src/extra/gfx/drawable.cpp
sinkingsugar: UX and future UI wise setting None is actually a good point. 👍 (we won't show any connection possible on top of the shard visualization!) But I think None will spawn other issues (as it's used during compose maybe?) Can try to set None and see if works out!

guusw: Ok let me try

Kryptos-FR: I have used `None` in other places (mostly UI) without trouble so far. I also thinks it is good to indicate that either a shard is not affected by whatever inputs comes before, or that it doesn't return anything (kind of a black hole), though I don't think we have anything right now that outputs `None` (we usually pass through).

guusw: Yeah no issues, will annotate with NoneType for optional & no input/output from now on.

src/extra/gfx.cpp
guusw: Added NoneType in these cases to the list of parameter types

src/extra/gfx/gltf.cpp
```diff
@@ -16,13 +16,14 @@ struct GLTFShard {
...
+
   static SHTypesInfo inputTypes() { return CoreInfo::AnyTableType; }
   static SHTypesInfo outputTypes() { return Types::DrawableHierarchy; }
 
   static SHParametersInfo parameters() {
     static Parameters params{
-        {"Transform", SHCCSTR("The transform variable to use"), {TransformVarType}},
-        {"Path", SHCCSTR("The static path to load a model from during warmup"), {CoreInfo::StringType}},
+        {"Transform", SHCCSTR("The transform variable to use (Optional)"), {CoreInfo::NoneType, TransformVarType}},
+        {"Path", SHCCSTR("The static path to load a model from during warmup (Optional)"), {CoreInfo::NoneType, CoreInfo::StringType}},
```
Kryptos-FR, line 26: Can use `CoreInfo::StringOrNone`

In addition, because the string is optional, `setParam()` and `getParam()` should be adpated (see for example in `imgui.cpp` lines 794-884):
```cpp
  void setParam(int index, const SHVar &value) {
    switch (index) {
    case 0:
      _transformVar = value;
      break;
    case 1:
      if (value.valueType == None) {
        _staticModelPath .clear();
      } else {
        _staticModelPath = value.payload.stringValue;
      }
      break;
    }
  }
  SHVar getParam(int index) {
    switch (index) {
    case 0:
      return _transformVar;
    case 1:
      return _staticModelPath.size() == 0 ? Var::Empty : Var(_staticModelPath);
    default:
      return Var::Empty;
    }
  }
```

guusw: Ok added that

Kryptos-FR (approved):

src/gfx/renderer.cpp
```diff
@@ -214,7 +220,9 @@ struct RendererImpl final : public ContextData {
   size_t frameIndex = 0;
   const size_t maxBufferedFrames = GFX_RENDERER_MAX_BUFFERED_FRAMES;
 
+  // TODO: Replace with render graph https://github.com/fragcolor-xyz/shards/issues/172
```
inabe-a, line 223: +1

inabe-a (approved):
Fix the optional and the ptr to shared_ptr and submit.

src/extra/gfx.cpp
```diff
@@ -16,16 +16,18 @@ namespace gfx {
 using shards::Mat4;
 
 struct DrawablePassShard {
-  static SHTypesInfo inputTypes() { return CoreInfo::AnyType; }
+  static SHTypesInfo inputTypes() { return CoreInfo::NoneType; }
```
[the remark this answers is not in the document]
guusw: Yeah NoneType works and seems correct. That way the UI can show that the shard has no Input/Output instead of Any which could be anything

src/gfx/gltf/gltf.cpp
```diff
@@ -154,7 +154,7 @@ struct Loader {
     }
 
     // Make position the first attribute to aid renderdoc debugging
-    if (positionIndex >= 0 && positionIndex != 0) {
```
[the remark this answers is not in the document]
guusw: Haha yes

src/gfx/pipeline_step.hpp
```diff
@@ -2,6 +2,7 @@
 #define GFX_PIPELINE_STEP
 
 #include "fwd.hpp"
+#include <variant>
```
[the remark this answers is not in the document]
guusw: There's a typedef at the bottom

src/extra/gfx/main_window.cpp
```diff
@@ -141,6 +141,10 @@ struct MainWindow : public Base {
     globals->renderer = std::make_shared<Renderer>(*globals->context.get());
     globals->imgui = std::make_shared<ImGuiRenderer>(*globals->context.get());
 
+    globals->drawQueue = std::make_shared<DrawQueue>();
+    auto clearDrawQueue = [dq = globals->drawQueue]() { dq->clear(); };
+    globals->onBeginFrame.push_back(MainWindowGlobals::FrameCallback{.context = this, .callback = clearDrawQueue});
```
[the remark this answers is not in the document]
guusw: It's not supposed to be, I'm using it more as an event.
So you could add into this vector to be notified and you are responsible for removing it when you no longer need to be.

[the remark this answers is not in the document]
guusw: Anyway, I simplified it since I don't actually use it anywhere else.

PR #295, opened 2022-06-30
egui gfx backend

guusw:
- Add egui rendering backend & standalone test
- A bit of restructuring on how bindings are generated.
- Setup bindings to gfx headers using mostly opaque types
- Add a create with bindgen scripts for gfx headers


src/extra/rust_interop.cpp
```diff
@@ -0,0 +1,46 @@
...
+  if (var.valueType != SHType::Object)
+    throw std::logic_error("Invalid type");
+  if (var.payload.objectVendorId != typeInfo.object.vendorId)
+    throw std::logic_error("Invalid object vendor id");
+  if (var.payload.objectTypeId != typeInfo.object.typeId)
+    throw std::logic_error("Invalid object type id");
+  return reinterpret_cast<T *>(var.payload.objectValue);
+}
+
+SHVar MainWindowGlobals_getDefaultQueue(const SHVar &mainWindowGlobals) {
```
sinkingsugar, line 28: thread safety of all those is dubious though.... maybe they should be thread_local

src/gfx/context.cpp
```diff
@@ -176,7 +181,7 @@ struct ContextMainOutput {
     swapchainDesc.format = swapchainFormat;
     swapchainDesc.width = newSize.x;
     swapchainDesc.height = newSize.y;
-    swapchainDesc.presentMode = WGPUPresentMode_Fifo;
+    swapchainDesc.presentMode = WGPUPresentMode_Immediate;
```
sinkingsugar, line 184: isn't Fifo better?

src/gfx/rust/build/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
Kryptos-FR, line 1: Missing end of line.

src/gfx/rust/build/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "gfx-build"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bindgen = "0.60.1"
```
Kryptos-FR, line 7: Missing end of line.

src/extra/rust_interop.cpp
guusw: Are you talking about getMainWindowGlobalsType/getQueueType ?

Apparently since C++11 these are considered thread safe https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

src/gfx/context.cpp
guusw: I think it still switches to that automatically on some platforms that enforce it. but we want to control our own frame-rate so we can target 120 or 240 for example for better UI response times

src/extra/rust_interop.cpp
sinkingsugar: Not really well defined but anyways do we assume that only one instance of those is ever needed in a single process? Feels like bgfx 😅

guusw: MainWindowGlobals_getDefaultQueue just gets a value inside this context variable. You could still technically have multiple windows (each has their own `mainWindowGlobals` but is set so eveything inside :Content can access it)

I only use static in this file for the types, which don't change

Kryptos-FR (approved):

src/gfx/context.cpp
```diff
@@ -431,7 +448,7 @@ void Context::requestDevice() {
   deviceDesc.nextInChain = &deviceExtras.chain;
 #endif
 
-  getLogger()->debug("Requesting wgpu device");
+  SPDLOG_LOGGER_DEBUG(getLogger(), "Requesting wgpu device");
```
sinkingsugar, line 451: Still wondering why you didn't use existing macros here

src/gfx/egui/src/standalone_test.rs
```diff
@@ -0,0 +1,74 @@
+use super::*;
+use crate::Renderer;
+use egui::RichText;
+use std::sync::Mutex;
```
sinkingsugar, line 4: Why?

src/gfx/egui/src/standalone_test.rs
```diff
@@ -0,0 +1,74 @@
...
+        max: egui::pos2(screen_rect_size.x, screen_rect_size.y),
+    });
+
+    let draw_scale_vec = gfx_Window_getDrawScale_ext(window);
+    let draw_scale = f32::max(draw_scale_vec.x, draw_scale_vec.y);
+    input.pixels_per_point = Some(draw_scale);
+
+    let full_output = app.ctx.run(input, |ctx| {
+        egui::CentralPanel::default().show(ctx, |ui| {
+            let mut state = STATE.lock().unwrap();
```
sinkingsugar, line 54: I don't think we need any synchronization as shards ensures that no?

cmake/osx_rust_build.sh
```diff
@@ -1,3 +1,7 @@
 #!/bin/bash
 set -ex
-SDKROOT=`xcrun --sdk macosx --show-sdk-path` $@
\ No newline at end of file
+TARGET_SDKROOT=`xcrun --sdk $1 --show-sdk-path`
```
sinkingsugar, line 3: What's this script? I need to run this all the time on Mac? If so annoying.

src/gfx/context.cpp
```diff
@@ -27,7 +29,10 @@ static inline std::shared_ptr<spdlog::logger> &getLogger() {
 #ifdef WEBGPU_NATIVE
 static WGPUBackendType getDefaultWgpuBackendType() {
 #if GFX_WINDOWS
-  return WGPUBackendType_D3D12;
```
sinkingsugar, line 30: Maybe revert this? Anyway let's keep a closer look I guess. Maybe need to test both in CI too

src/gfx/egui/src/standalone_test.rs
sinkingsugar: Just realised this is on a test file 🤣

sinkingsugar (approved):
Just some small things but ok

src/gfx/context.cpp
guusw: I want to send the graphics logs to it's own logger so I can filter them out / tag them / etc.

cmake/osx_rust_build.sh
sinkingsugar: Can you document it? We need some iOS instructions at least.

src/gfx/context.cpp
sinkingsugar: Any idea why runs slow? Cos of some mistakes they did like you said last time?

guusw: Yeah, one of those open issues about fixing D3D12. Copying data to buffers is currently very slow

PR #296, opened 2022-07-04
Capture variables (Spawn/Detach)

sinkingsugar:
Make Spawn and Detach capture used variables


Kryptos-FR:
Is it just a copy or would changes on the captured variable be reflected in the calling wire?

sinkingsugar:
> Is it just a copy or would changes on the captured variable be reflected in the calling wire?

Just a copy. As we can't predict flow basically.

sinkingsugar:
@guusw @Kryptos-FR this PR opens up a lot of scenarios like:
```clojure
(defloop ui-wire
  (UI.Window 
    (-> 
      ; use UI to give a transform
      .... = .my-cube-transform
      (UI.Button "Spawn" (-> (Spawn my-cube)))))
(GFX.MainWindow 
  (-> 
    ; this creates and ticks a mesh here! mesh wires will have all variables (GFX!!)
    (Branch [ui-wire])))
```

sinkingsugar:
@satendrapalsingh you should expand the pseudo code I wrote above and write a sample.

satendrapalsingh:
> @satendrapalsingh you should expand the pseudo code I wrote above and write a sample.

Yes, added it to issue #299

src/core/runtime.cpp
```diff
@@ -1500,22 +1503,19 @@ SHTypeInfo deriveTypeInfo(const SHVar &value, const SHInstanceData &data, bool *
...
   } break;
   case SHType::ContextVar: {
-    if (containsVariables) {
-      // containsVariables is used by Const shard mostly
-      *containsVariables = true;
+    if (expInfo) {
       const auto varName = value.payload.stringValue;
       for (auto info : data.shared) {
         if (strcmp(info.name, varName) == 0) {
+          expInfo->push_back(SHExposedTypeInfo{.name = info.name, .exposedType = info.exposedType});
```
guusw, line 1515: wrap around  `if(expInfo)` here, so the value is still returned when it's null

src/core/shards/flow.cpp
```diff
@@ -834,8 +834,9 @@ struct Match {
     for (auto &case_ : _pcases) {
       if (case_.valueType != None) {
         // must compare deeply
-        bool hasVariables = false;
-        TypeInfo cinfo(case_, data, &hasVariables);
+        // also pass vector to ensure we get context var info (even if we don't use it)
+        std::vector<SHExposedTypeInfo> types;
+        TypeInfo cinfo(case_, data, &types);
```
guusw, line 839: &types can just be null if you apply my comment in `deriveTypeInfo`

src/core/shards/wires.cpp
```diff
@@ -631,6 +633,28 @@ struct Recur : public WireBase {
...
+
+    if (capturing) {
+      // build the list of variables to capture and inject into spawned chain
+      _vars.clear();
+      for (auto &require : wireValidation.requiredInfo) {
+        if (!require.global) {
+          SHVar ctxVar{};
+          ctxVar.valueType = ContextVar;
+          ctxVar.payload.stringValue = require.name;
+          auto &p = _vars.emplace_back();
```
guusw, line 649: Can you change it to `_vars.emplace_back(ctxVar);`?

guusw (commented):
Some questions/suggestions

src/core/shards/wires.cpp
sinkingsugar: No, it's on purpose to trigger `operator=` clone, I think.

guusw: Constructor with SHVar& has clone too

sinkingsugar: did you test it? hmm can swap to vector if so

guusw: I didn't but it should work on both I think. 
If it doesn't work because the `TParamVar(SHVar initialValue)` is explicit, just keep this.

sinkingsugar:
> I feel like this should work to in the way you described spawn, although it fails on the second run.
> 
> ```clojure
> (defwire spawned-wire
>   .consumed-variable
>   (Log "============ Spawned wire")
>   (ToInt)
>   (Math.Add 1))
> 
> (defwire test-capture
>   "2" = .consumed-variable
>   (Spawn spawned-wire) = .spawned-wire-0
>   (Spawn spawned-wire) = .spawned-wire-1
>   (Wait .spawned-wire-0)
>   (Assert.Is 3 true)
>   (Wait .spawned-wire-1)
>   (Assert.Is 3 true))
> ```

For some reasons only first (Spawn gets composed...

src/core/runtime.cpp
```diff
@@ -1377,12 +1377,16 @@ SHComposeResult composeWire(const SHWire *wire, SHValidationCallback callback, v
     // If first shard is a plain None, mark this wire has None input
     // But make sure we have no (Input) shards
     auto inTypes = wire->shards[0]->inputTypes(wire->shards[0]);
-    if (inTypes.len == 1 && inTypes.elements[0].basicType == None)
+    if (inTypes.len == 1 && inTypes.elements[0].basicType == None) {
       wire->inputType = SHTypeInfo{};
-    else
+      wire->inputTypeForceNone = true;
```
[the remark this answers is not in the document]
sinkingsugar: check other usage, basically we want to know if we did this fix-up here or not in wires.cpp

src/core/runtime.cpp
sinkingsugar: Good point but I'm afraid sometimes we actually want to return as it is ContextVar as result/type... Wonder if we have coverage for that.
For some reasons the previous version was doing the same (with a bool switch)

guusw: Hmm yeah I see it was like that already. okay fine.

sinkingsugar (commented):
I did the thing where I did not submit and lost them too...

src/core/shards/wires.cpp
```diff
@@ -165,7 +166,7 @@ struct WireBase {
     } else {
       SHLOG_TRACE("Skipping {} compose", wire->name);
       // verify input type
-      if (!passthrough && mode != Stepped && data.inputType != wire->inputType) {
+      if (!passthrough && mode != Stepped && data.inputType != wire->inputType && !wire->inputTypeForceNone) {
```
sinkingsugar, line 169: @guusw check this for why `inputTypeForceNone` is needed. As None could be legit and not match.

guusw (approved):

PR #297, opened 2022-07-04
egui input

guusw:
Add input backend for egui

### DIFF
--- a/src/gfx/egui/src/input.rs
+++ b/src/gfx/egui/src/input.rs
@@ -136,9 +136,29 @@ pub fn to_egui_cursor_icon(i: CursorIcon) -> egui_CursorIcon {
     }
 }
 
+#[derive(Debug)]
+pub enum TranslationError {
+    Text(std::str::Utf8Error),
+}
+
+impl From<std::str::Utf8Error> for TranslationError {
+    fn from(e: std::str::Utf8Error) -> Self {
+        TranslationError::Text(e)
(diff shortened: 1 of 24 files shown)


src/gfx/egui/src/input.rs
```diff
@@ -0,0 +1,267 @@
...
+        CursorIcon::ResizeColumn => egui_CursorIcon_ResizeColumn,
+        CursorIcon::ResizeRow => egui_CursorIcon_ResizeRow,
+        CursorIcon::ZoomIn => egui_CursorIcon_ZoomIn,
+        CursorIcon::ZoomOut => egui_CursorIcon_ZoomOut,
+    }
+}
+
+pub fn translate_raw_input(
+    input: &egui_Input,
+) -> Result<egui::RawInput, Box<dyn std::error::Error>> {
```
Kryptos-FR, line 141:
```suggestion
) -> Result<egui::RawInput, &str> {
```

Kryptos-FR (changes requested):

src/gfx/egui/src/input.rs
guusw: I changed it to 
```
pub enum TranslationError {
    Text(std::str::Utf8Error),
}
```
you should be able to silently ignore that one in the UI shard though, maybe just a log

Kryptos-FR (approved):

PR #306, opened 2022-07-07
Use boost source archive

guusw:
Bundled boost into a source archive that doesn't contain tests/docs/unused modules (9.62 MB)

Check the dep branch for the build/package script.
Pushing a tag to that repo wil automatically create a release that can be linked to inside FetchContent_Declare
https://github.com/chainblocks/boost/commit/e11da3be380e29c5021912816e293c73127960f3

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -82,6 +82,3 @@
 [submodule "deps/imgui"]
 	path = deps/imgui
 	url = https://github.com/ocornut/imgui.git
-[submodule "deps/boost"]
-	path = deps/boost
-	url = https://github.com/chainblocks/boost.git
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -81,7 +81,4 @@
 	url = https://github.com/chainblocks/wgpu-native.git
 [submodule "deps/imgui"]
 	path = deps/imgui
-	url = https://github.com/ocornut/imgui.git
-[submodule "deps/boost"]
-	path = deps/boost
-	url = https://github.com/chainblocks/boost.git
+	url = https://github.com/ocornut/imgui.git
```
Kryptos-FR, line 84: Missing end of line

Kryptos-FR:
I think the chainblocks/boost repo should also update a `latest` tag that point to the last packaged release.

guusw:
Don't know if it'll be much use, I decided to just link to specific tags since you have to specify the SHA when downloading the external archive in cmake anyway.

Always linking to latest doesn't seem like a good idead since it can break older commits if they suddenly compile against newer dependencies

PR #310, opened 2022-07-12
Rust str fixes

sinkingsugar:
removed `impl<'a> From<&'a str> for Var`

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1205,31 +1205,6 @@ impl From<String> for &str {
   }
 }
 
-impl<'a> From<&'a str> for Var {
-  #[inline(always)]
-  fn from(v: &'a str) -> Self {
-    let len = v.len();
-    let p = if len > 0 {
-      v.as_ptr()
-    } else {
-      core::ptr::null()
-    };
-    SHVar {
-      valueType: SHType_String,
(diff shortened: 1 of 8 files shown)


rust/src/types.rs
```diff
@@ -1733,8 +1708,34 @@ impl From<&[u8]> for Var {
...
+            stringLen: len as u32,
+            stringCapacity: 0,
+          },
+        },
+      },
+      ..Default::default()
+    }
+  }
+
+  pub fn ephemeral_context_variable(name: &'static str) -> Var {
```
sinkingsugar, line 1736: seems not used, removing

Kryptos-FR:
So if I get this right, it is to prevent doing conversions from `&str` to `Var` in places where it isn't safe and force us to explicitly use `Var::ephemeral_string()` so that it is clearer in the code when we do it?

Kryptos-FR (approved):

sinkingsugar:
> So if I get this right, it is to prevent doing conversions from `&str` to `Var` in places where it isn't safe and force us to explicitly use `Var::ephemeral_string()` so that it is clearer in the code when we do it?

Exactly, which creates a `Var` that is meant to be `cloned` at some point in the scope. (this will fix `\0`) as well)

PR #312, opened 2022-07-12
Egui text input

Kryptos-FR:
Implement `UI.TextInput` (single line for now).

It supports having a mutable or immutable initial value. In case a non-existing variable is given, it will expose it.

```clj
;; const string
(UI.TextInput "You can't edit this!")

;; immutable variable
"Hello World!" = .imm
(UI.TextInput .imm)

;; mutable variable
"Lorem ipsum" >= .mut
(UI.TextInput .mut)

;; exposing a new string variable
(UI.TextInput .new)
```

Manipulating the `dyn TextBuffer` is optimized to only clone when the string capacity is not sufficient. Otherwise it does pointer arithmetic to move slices of memory around.

### DIFF
--- a/rust/src/shards/gui/widgets/text_input.rs
+++ b/rust/src/shards/gui/widgets/text_input.rs
@@ -19,18 +19,27 @@ use crate::types::Parameters;
 use crate::types::Type;
 use crate::types::Types;
 use crate::types::Var;
+use crate::types::BOOL_TYPES_SLICE;
 use crate::types::NONE_TYPES;
 use crate::types::STRING_TYPES;
 use std::cmp::Ordering;
 use std::ffi::CStr;
 
 lazy_static! {
-  static ref TEXTINPUT_PARAMETERS: Parameters = vec![(
-    cstr!("Variable"),
-    cstr!("The variable that holds the input value."),
-    STRING_VAR_SLICE,
-  )
-    .into(),];
+  static ref TEXTINPUT_PARAMETERS: Parameters = vec![
+    (
+      cstr!("Variable"),
+      cstr!("The variable that holds the input value."),
+      STRING_VAR_SLICE,
+    )
+      .into(),
+    (
+      cstr!("Multiline"),
+      cstr!("Support multiple lines."),
+      BOOL_TYPES_SLICE,
+    )
+      .into(),
+  ];
 }
 
 impl Default for TextInput {
@@ -41,6 +50,7 @@ impl Default for TextInput {
       parents,
       requiring: Vec::new(),
       variable: ParamVar::default(),
+      multiline: ParamVar::new(false.into()),
       exposing: Vec::new(),
       should_expose: false,
       mutable_text: true,
@@ -55,7 +65,6 @@ enum VarTextBuffer<'a> {
 
 impl AsRef<str> for VarTextBuffer<'_> {
   fn as_ref(&self) -> &str {
-
     let var = match self {
       VarTextBuffer::Editable(var) => var.as_ref(),
       VarTextBuffer::ReadOnly(var) => var,
@@ -239,6 +248,7 @@ impl Shard for TextInput {
   fn setParam(&mut self, index: i32, value: &Var) -> Result<(), &str> {
     match index {
       0 => Ok(self.variable.set_param(value)),
+      1 => Ok(self.multiline.set_param(value)),
       _ => Err("Invalid parameter index"),
     }
   }
@@ -246,6 +256,7 @@ impl Shard for TextInput {
   fn getParam(&mut self, index: i32) -> Var {
     match index {
       0 => self.variable.get_param(),
+      1 => self.multiline.get_param(),
       _ => Var::default(),
     }
   }
@@ -318,11 +329,13 @@ impl Shard for TextInput {
   fn warmup(&mut self, ctx: &Context) -> Result<(), &str> {
     self.parents.warmup(ctx);
     self.variable.warmup(ctx);
+    self.multiline.warmup(ctx);
 
     Ok(())
   }
 
   fn cleanup(&mut self) -> Result<(), &str> {
+    self.multiline.cleanup();
     self.variable.cleanup();
     self.parents.cleanup();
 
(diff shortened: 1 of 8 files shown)


rust/src/shards/gui/widgets/text_input.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+      }
+    } else {
+      self.mutable_text = false;
+    }
+
+    Ok(common_type::string)
+  }
+
+  fn exposedVariables(&mut self) -> Option<&ExposedTypes> {
```
guusw, line 241: Ah so you can write `(TextInput .variable)` without having to assign it before?

Kryptos-FR: Yes it's a supported scenario.

rust/src/shards/gui/widgets/text_input.rs
```diff
@@ -0,0 +1,309 @@
...
+      requiring: Vec::new(),
+      variable: ParamVar::default(),
+      exposing: Vec::new(),
+      should_expose: false,
+      mutable_text: true,
+    }
+  }
+}
+
+impl AsRef<str> for Var {
```
guusw, line 51: I feel like this might be better as a dedicated function: `as_str` or something.
Also prevents needing the turbofish in http.rs

rust/src/shards/gui/widgets/text_input.rs
```diff
@@ -0,0 +1,309 @@
...
+          .__bindgen_anon_2
+          .stringCapacity as usize,
+      )
+    };
+
+    if current_cap == 0usize {
+      // new string
+      let var = Var::ephemeral_string(text);
+      cloneVar(self, &var);
+    } else if (current_cap - current_len) >= text_len {
```
guusw, line 103: Doesn't there need to be a +/- 1 here since you add the null terminator?

rust/src/shards/gui/widgets/text_input.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+}
+
+impl egui::TextBuffer for Var {
+  fn is_mutable(&self) -> bool {
+    true
+  }
+
+  fn insert_text(&mut self, text: &str, char_index: usize) -> usize {
```
guusw, line 80: Could maybe clarify the variable names by clearly naming `text` => `inserted_text`/`inserted_text_length_` etc.

guusw (changes requested):
Minor notes/questions. otherwise looks good.

rust/src/shards/gui/widgets/text_input.rs
sinkingsugar: No cos clone does this:
```c++
   auto srcSize = src.payload.stringLen > 0 || src.payload.stringValue == nullptr ? src.payload.stringLen
                                                                                   : uint32_t(strlen(src.payload.stringValue));
    if ((dst.valueType != String && dst.valueType != ContextVar) || dst.payload.stringCapacity < srcSize) {
      destroyVar(dst);
      // allocate a 0 terminator too
      dst.payload.stringValue = new char[srcSize + 1];
      dst.payload.stringCapacity = srcSize;
    } else {
      if (src.payload.stringValue == dst.payload.stringValue)
        return;
    }

    dst.valueType = src.valueType;
    memcpy((void *)dst.payload.stringValue, (void *)src.payload.stringValue, srcSize);
    ((char *)dst.payload.stringValue)[srcSize] = 0;
    // fill the optional len field
    dst.payload.stringLen = srcSize;
```

guusw: As long as this value is always initialized with cloneVar, should be fine then

sinkingsugar: it needs that yes

rust/src/shards/gui/widgets/text_input.rs
Kryptos-FR: It's required by `egui`. Nothing we can do.

See the definition of the `TextBuffer` trait:
```rust
pub trait TextBuffer: AsRef<str>
```

rust/src/shards/gui/widgets/text_input.rs
Kryptos-FR: Usually I like to keep the same name as defined in the trait to avoid confusion when reading the doc.

rust/src/shards/gui/widgets/text_input.rs
Kryptos-FR: I opened https://github.com/emilk/egui/pull/1824

rust/src/shards/gui/widgets/text_input.rs
guusw: It's a bit of a nitpick but it makes it harder to follow when you just bump into this function IMO, since you're dealing with 2 sets of pointers+lengths

rust/src/shards/gui/widgets/text_input.rs
guusw: Can you make a type just for the purpose of implementing TextBuffer inside that function?
Something like: 
```rs
enum VarTextBuffer {
 Editable(&mut Var),
 ReadOnly(&Var),
}
```

Since it only really has to live inside that shard's activate function

rust/src/shards/gui/widgets/text_input.rs
```diff
@@ -0,0 +1,368 @@
...
+        var.payload.__bindgen_anon_1.__bindgen_anon_2.stringLen = new_len as u32;
+        // fixup null-terminator
+        *base_ptr.add(new_len) = 0;
+      }
+    }
+  }
+}
+
+impl VarTextBuffer<'_> {
+  fn as_str_from_mut(var: &mut Var) -> &str {
```
guusw, line 172: Since it's seems exactly the same I seem to be able to replace it with
```rs
fn as_str_from_mut(var: &mut Var) -> &str {
    VarTextBuffer::as_str(var)
  }
```

guusw (changes requested):

rust/src/shards/gui/widgets/text_input.rs
guusw: Maybe even call it directly with `&mut Var` but maybe it needs a function in between

guusw (approved):

PR #315, opened 2022-07-13
ParamVar get returns a reference

Kryptos-FR:
`ParamVar::get()` now returns a `&Var` instead of `Var` which created a copy.

This should make it a bit more efficient since a lot of time we get it in order to convert it into another value:
```rust
let text: &str = self.label.get().as_ref().try_into()?;
// can be simplified to
let text: &str = self.label.get().try_into()?;
```

In the other cases where we do need to get a `Var` we can make an explicit copy by dereferencing:
```rust
let simulation: Var = *self.simulation.get();
```


sinkingsugar (approved):
LGTM

PR #331, opened 2022-07-26
Rust-crates-bump

sinkingsugar:
Bumping dependencies


Kryptos-FR:
Do we always bump by running `cargo update` or are there some packages that we don't always update?

Wondering if having a CI job do that for us would make sense (on a schedule or on-demand).

PR #336, opened 2022-07-27
Android/iOS fixes

guusw:
Various fixes made while working on the product demo / app template.

- Cleaned up how the UI scaling factor is retrieved from the OS and separate it from an application defined scaling factor that can be passed into egui separately

- Fix SDLmain not being linked against correctly for android

- Embed all direct wgpu dependencies hosted on chainblocks repos 

- Additionally I made some changes in how the android package is built, it is now managed through CMake too in order to have a bit more flexibility and a smoother workflow without having go through the gradle process.\
The script is included through a submodule in [cmake/android-cmake/Android.cmake](https://github.com/guusw/android-cmake/blob/main/Android.cmake)


deps/libressl.patch
```diff
@@ -0,0 +1,40 @@
+diff --git a/crypto/CMakeLists.txt b/crypto/CMakeLists.txt
```
sinkingsugar, line 1: what's this?

guusw: There's an issue when building for iOS where it doesn't link against this because it defines and empty static lib file, this is a fix for that.

sinkingsugar: you don't need I think, you can force linking. try `--force_load` or something please.

sinkingsugar: `-Wl,-force_load lib.a`

guusw: Let me try that

guusw: It's a different issue and unique to XCode where it doesn't even generate the lib<name>.a file in this case

sinkingsugar: what's this crypto library anyway? It's kinda a while hack prone to break :(

guusw: It's because some boost libraries reference it (asio/beast). Also rust uses it through openssl-sys on other non-apple platforms

guusw: We're using a fixed version btw. (because rust requires it), so I don't think having a fix like this hurts.
We can have it included in a later version of the library and then we can remove the patch

sinkingsugar: 👍👍

src/gfx/renderer.cpp
```diff
@@ -224,7 +224,14 @@ struct RendererImpl final : public ContextData {
   bool mainOutputWrittenTo = false;
   bool depthStencilWrittenTo = false;
 
-  std::shared_ptr<ViewTexture> depthTexture = std::make_shared<ViewTexture>(WGPUTextureFormat_Depth24Plus, "Depth Buffer");
+  static constexpr WGPUTextureFormat defaultDepthTextureFormat =
+#if GFX_APPLE
+      WGPUTextureFormat_Depth32Float
+#else
+      WGPUTextureFormat_Depth32Float
+#endif
```
sinkingsugar, line 232: looks equal?

sinkingsugar (approved):
LGTM but noticed a code dupe

Kryptos-FR (approved):
LGTM

sinkingsugar:
@guusw merge this?

PR #342, opened 2022-08-04
Daily fixes

sinkingsugar:
fix fullscreen functionality and drop warning

### DIFF
--- a/src/extra/gfx/main_window.cpp
+++ b/src/extra/gfx/main_window.cpp
@@ -22,8 +22,8 @@ SDL_Window *Base::getSdlWindow() { return getWindow().window; }
 struct MainWindow : public Base {
   static inline Parameters params{
       {"Title", SHCCSTR("The title of the window to create."), {CoreInfo::StringType}},
-      {"Width", SHCCSTR("The width of the window to create. In pixels and DPI aware."), {CoreInfo::IntType}},
-      {"Height", SHCCSTR("The height of the window to create. In pixels and DPI aware."), {CoreInfo::IntType}},
+      {"Width", SHCCSTR("The width of the window to create. In pixels and DPI aware."), {CoreInfo::IntOrNone}},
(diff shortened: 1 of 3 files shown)


sinkingsugar:
I'm making full screen (borderless) default for practical reasons @guusw , have a look please

sinkingsugar:
I think fits also mobile devices where stuff is full screen anyway

sinkingsugar:
Actually I'm not so convinced, maybe we need a fullscreen parameter like before?

src/extra/gfx/main_window.cpp
```diff
@@ -126,6 +138,11 @@ struct MainWindow : public Base {
     WindowCreationOptions windowOptions = {};
     windowOptions.width = _pwidth;
     windowOptions.height = _pheight;
+    if (_pwidth == 0 || _pheight == 0) {
+      windowOptions.fullscreen = true;
+    } else {
+      windowOptions.fullscreen = false;
+    }
```
Kryptos-FR, line 145: One issue with this logic is not being able to display in fullscreen at a lower resolution than the screen. Sometimes for performance reason, you want to be fullscreen at `1920*1200` while your screen native resolution is `2560*1440`.

sinkingsugar: In modern rendering you don't do that anymore actually @Kryptos-FR , just use render targets at lower resolution and render all to texture (which you always do due to post effects) 😉

guusw:
I held of on this since having a fixed parameter is not that useful without being able to switch on demand, but I'm fine with adding it like this.

guusw (approved):

sinkingsugar:
> I held off on this since having a fixed parameter is not that useful without being able to switch on demand, but I'm fine with adding it like this.

But I mean, it's resizable when not fullscreen (windowed) anyway no?

sinkingsugar:
The issue is we make no distinction without maximixed and full screen with this PR

sinkingsugar:
We might want a flag

PR #344, opened 2022-08-07
Fix documentation for shards `Hashed`, `Await`, and `Once`

satendrapalsingh:


docs/details/shards/General/Await.md
```diff
@@ -0,0 +1,6 @@
+`(Await)` triggers the execution of its shard (or sequence of shards) and passes the control back to the main program. Once the `(Await)` shard(s) have finished their execution the result of their computation is made available to the prgoram by `(Await)`.
```
Kryptos-FR, line 1: Typo "prgoram"

docs/details/shards/General/Await.md
```diff
@@ -0,0 +1,6 @@
...
+
+This is called asynchronous computation and is used to prevent resource intensive processing (like downloading a large file data from an http server) from holding up the execution of the rest of the shards.
+
+!!! note
+    `(Await)` has an alias `(||)` which is more convenient to use. `||` also removes the need to use `(->)` as, unlike `(Hashed)`, it doesn't require the parameter shards to be grouped together.
```
Kryptos-FR, line 6:
```suggestion
    `(Await)` has an alias `(||)` which is more convenient to use. `||` also removes the need to use `(->)` as, unlike `(Await)`, it doesn't require the parameter shards to be grouped together.
```

docs/samples/shards/General/Await/1.edn
```diff
@@ -0,0 +1,21 @@
+;; Asynchronous execution of shards (using keyword `Await`)
+(defmesh mesh)
+
+(defloop await-wire
+(Await
+  ;:Shards
+  (->
+    (Msg "Await worker1")
+    ;; printed 1st => [info] [2022-08-07 16:35:31.636] [T-29916] [logging.cpp::98] [await-wire] Await worker1
```
Kryptos-FR, line 9: I wouldn't put the log here since every time we generate the doc, it would be different.
Just keep "printed 1st".

docs/samples/shards/General/Await/2.edn
```diff
@@ -0,0 +1,19 @@
...
+  ))
+
+(defloop my-wire
+    (Msg "Log without waiting for Await shards to complete")
+    ;; printed 2nd => [info] [2022-08-07 16:39:45.783] [T-5736] [logging.cpp::98] [my-wire] Log without waiting for Await shards to complete
+    )
+
+(schedule mesh await-wire)
+(schedule mesh my-wire)
+(run mesh 1 1)
```
Kryptos-FR, line 19: Missing empty line at end of file.

docs/details/shards/General/Hashed.md
```diff
@@ -0,0 +1,6 @@
+`(Hashed)` takes a single shard or a sequence of shards and hashes the output of the last shard (or the only shard) in this sequence. So this shard can be used to group together certain shards when you want their final output to be hashed.
+
+Like [(`Sub`)](../sub)), `(Hashed)` too uses `(->)` but unlike `(Sub)`, `(Hashed)` doesn't pass its input to all its shards in parallel. The input to `(Hashed)` becomes input to the first shard in the hashed sequence and the output of this shard becomes the input for the next shard in the hashed sequence (following the regular Shards data flow pattern).
```
Kryptos-FR, line 3: This is confusing. It is the regular flow of any shards to give the input to the first inner shard. `Sub` also doesn't do anything in parallel, certainly not giving the same input to each inner shards.

```clj
5
(| (Math.Add 1)
   (Log) ;; this won't print 5 but 6 as expected
)
;; however, output will be 5 here
```
```

**Review by Kryptos-FR — CHANGES_REQUESTED** [review]

File: docs/details/shards/General/Await.md (hunk 1 above)
Comment (sinkingsugar) [reply docs/details/shards/General/Await.md:1]: Await actually runs the inner shards in a task that is sent to the thread pool specifically

Comment (satendrapalsingh) [reply docs/details/shards/General/Await.md]: Fixed/rewritten.

File: docs/details/shards/General/Await.md (hunk 2 above)
Comment (satendrapalsingh) [reply docs/details/shards/General/Await.md]: Fixed.

File: docs/samples/shards/General/Await/1.edn (hunk 3 above)
Comment (satendrapalsingh) [reply docs/samples/shards/General/Await/1.edn]: Comments rewritten to better explain the expected order of logging (of messages) and the reason for this.

File: docs/samples/shards/General/Await/2.edn (hunk 4 above)
Comment (satendrapalsingh) [reply docs/samples/shards/General/Await/2.edn]: Fixed.

File: docs/details/shards/General/Hashed.md (hunk 5 above)
Comment (satendrapalsingh) [reply docs/details/shards/General/Hashed.md]: Statement corrected.

File: docs/details/shards/General/Hashed.md [hunk 6]
```diff
@@ -1,6 +1,6 @@
 `(Hashed)` takes a single shard or a sequence of shards and hashes the output of the last shard (or the only shard) in this sequence. So this shard can be used to group together certain shards when you want their final output to be hashed.
 
-Like [(`Sub`)](../sub)), `(Hashed)` too uses `(->)` but unlike `(Sub)`, `(Hashed)` doesn't pass its input to all its shards in parallel. The input to `(Hashed)` becomes input to the first shard in the hashed sequence and the output of this shard becomes the input for the next shard in the hashed sequence (following the regular Shards data flow pattern).
+Like [(`Sub`)](../sub)), `(Hashed)` too uses `(->)` to group multiple shards. The input to `(Hashed)` becomes input to the first shard in the hashed sequence and the output of this shard becomes the input for the next shard in the hashed sequence (following the regular Shards data flow pattern).
```
Kryptos-FR, line 3:
```suggestion
Like [(`Sub`)](../sub)), `(Hashed)` also uses `(->)` to group multiple shards. The input to `(Hashed)` becomes input to the first shard in the hashed sequence and the output of this shard becomes the input for the next shard in the hashed sequence (following the regular Shards data flow pattern).
```

Or just remove it. But "too" is usually for other uses.

satendrapalsingh: I simplified the sentence by removing the `Sub` reference altogether.
It was not really adding anything to the meaning.

Kryptos-FR (approved):

Issue #272, opened 2022-06-13
Add a runtime option to set the log level


Kryptos-FR:
Currently the logging level is set based on whether the build is in RELEASE or DEBUG (see function `setupSpdLog()` in  `runtime.cpp`). In DEBUG we might want to enable or disable the trace logs.

Add a runtime switch to enable/disable trace logs and use that value during setup.

Note: this will only affect DEBUG builds. In RELEASE, debug and trace logs should remain disabled.

guusw:
`Add a runtime switch to enable/disable trace logs and use that value during setup.`
I'd add a way to set the level, e.g. `SHARDS_LOG=debug` or `SHARDS_LOG=trace` instead of just switching on/off

guusw:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/shards/272?planning-poker) with ZenHub @inabe-a @Kryptos-FR @sinkingsugar

sinkingsugar:
I'm against this as default.
It is intentional to completely omit the log levels at **compile-time** for obvious performance reasons.
I don't really want a runtime option in release mode.

sinkingsugar:
Also apart from my disapproval. Such kind of task is not something we should do now.

PR #346, opened 2022-08-08
Move logging setup code. Add runtime log levels

guusw:
- Move some of the logger setup code to a separate "module"
- Automatically grab runtime log level from a `LOG_<logger_name>=debug/trace/info/etc.` environment variable
- Examples:
  - `LOG_GFX=debug` debug output from graphics enabled
  - `LOG_shards=trace` trace logging from shards enabled
- The behavior of compiling out DEBUG/TRACE logs in release builds is still intact so trying to enabled them does nothing unless the log calls were outside of macros

Seems to close https://github.com/fragcolor-xyz/shards/issues/272 too

### DIFF
--- a/cmake/Root.cmake
+++ b/cmake/Root.cmake
@@ -27,6 +27,7 @@ set_property(GLOBAL PROPERTY CTEST_TARGETS_ADDED 1)
 
 add_subdirectory(${SHARDS_DIR}/deps deps)
 
+add_subdirectory(${SHARDS_DIR}/src/log src/log)
 add_subdirectory(${SHARDS_DIR}/src/core src/core)
 add_subdirectory(${SHARDS_DIR}/src/mal src/mal)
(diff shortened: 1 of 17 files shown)


Kryptos-FR:
Shouldn't LOG_shards be defaulted to at least DEBUG in debug mode? I'm ok with having TRACE disabled by default but it is rather unexpected to not have DEBUG logs when running a debug build.

sinkingsugar:
> Shouldn't `LOG_shards` be defaulted to at least `DEBUG` in debug mode? I'm ok with having `TRACE` disabled by default but it is rather unexpected to not have `DEBUG` logs when running a debug build.

I'd make it TRACE please, I don't want to remember to pass right flags.

guusw:
Changed the shards default log level to what is was previously (whatever SPDLOG_ACTIVE_LEVEL is set to)

Kryptos-FR (approved):

PR #357, opened 2022-08-10
Fix consistency on CMake target names

guusw:
Some where named `target-name` and some `target_name`, changed all of them to use dashes.
Renamed test targets as a result.


sinkingsugar (approved):
👍👍

Kryptos-FR (approved):

PR #364, opened 2022-08-16
UI copy paste

### LINKED ISSUE #359: Copy/paste with system clipboard is not working in the UI
Copying/pasting from within the UI should interact with the system clipboard. Currently, only pasting from the clipboard seems to work.

**Additional context**
In standard egui implementations, there is a specific "clipboard" feature.
See:
* https://github.com/emilk/egui/blob/master/egui_glium/Cargo.toml#L32
* https://github.com/emilk/egui/blob/master/egui-winit/src/clipboard.rs

Kryptos-FR:
Copied text was not transferred to the system clipboard.

Fixes #359.


Kryptos-FR:
**Open question**
The standalone test also sets the text cursor position:
```rust
app.input_translator
  .update_text_cursor_position(window, full_output.platform_output.text_cursor_pos);
```
@guusw do we need to do it as well? If so, how should we get a pointer to the gfx context from rust (or ideally not need to get it and just set the cursor position)?

rust/src/shards/gui/misc/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+struct Reset {
+  parents: ParamVar,
+  requiring: ExposedTypes,
+}
+
+mod clipboard;
+mod reset;
+
+pub fn registerShards() {
+  // registerShard::<GetClipboard>();
```
guusw, line 27: Meant to be disabled?

rust/src/shards/gui/misc/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+struct Reset {
+  parents: ParamVar,
+  requiring: ExposedTypes,
+}
+
+mod clipboard;
+mod reset;
+
+pub fn registerShards() {
```
guusw, line 26: Wouldn't it be better if we just added clipboard shards that interact with the system clipboard (probably though SDL_Get/SetClipboardText or rust crate) that way they can also be used outside of a UI context and would work the same way

rust/src/shards/gui/misc/mod.rs
Kryptos-FR: Yes because it doesn't work. Unfortunately, in `egui` the `context.copied_text` is cleared each frame so we can't use that to "read" the clipboard. We would have to request a native call to get the system clipboard. Since we might never need that shard, I won't spend time on it until we do.

rust/src/shards/gui/misc/mod.rs
Kryptos-FR: Probably better yes.

sinkingsugar:
Close for now ? or ? @Kryptos-FR @guusw ?

Kryptos-FR:
@sinkingsugar I would say -> Icebox?

sinkingsugar:
It's a PR, we either wrap it up quickly as it seems already going somewhere (hack solution ok) or we close it cos it will be outdated. (Also it can be reopened as well)
Also notice that I'm not mentioning much zenhub which we might or might not use next year.

Kryptos-FR:
@sinkingsugar I'll close it for now. It needs some rework.

PR #369, opened 2022-08-18
Rust shards enum

### LINKED ISSUE #366: enum support for shards in rust

Kryptos-FR:
Add support for declaring shards enum in rust.

There are two macros: `shenum!`  and `shenum_types`. The first one is very similar to what `bitflags!` was doing, except it doesn't generate operators for now. The second one declares the various `Type` that are required if we want to use the enum as input/output or parameter type.

Fixes #366.


sinkingsugar (approved):
LGTM

PR #372, opened 2022-08-23
Texture shards

guusw:
Add shards to create texture objects from images and adds texture parameters to Drawable and Material


src/extra/gfx/material_utils.hpp
```diff
@@ -11,6 +12,25 @@
 #include <spdlog/spdlog.h>
 
 namespace gfx {
+inline void varToTexture(const SHVar &var, TexturePtr &outVariant) {
+  switch (var.valueType) {
+  case SHType::Object: {
+    shards::Type type =
+        SHTypeInfo{SHType::Object, {.object = {.vendorId = var.payload.objectVendorId, .typeId = var.payload.objectTypeId}}};
```
Kryptos-FR, line 19: Opportunity to use `Type::Object(var.payload.objectVendorId, var.payload.objectTypeId)` once #284 is merged.

src/extra/gfx/material_utils.hpp
```diff
@@ -76,15 +96,46 @@ inline void initConstantShaderParams(MaterialParameters &out, SHTable &paramsTab
   }
 }
 
-inline void initReferencedShaderParams(SHContext *shContext, SHShaderParameters &shShaderParameters, SHTable &paramsTable) {
+inline void initConstantTextureParams(MaterialParameters &out, SHTable &texturesTable) {
```
Kryptos-FR, line 99: I would have the *out* parameter always last (or always first) if possible for consistency across multiple functions.

In addition, shouldn't `texturesTable` be ` const &`? But maybe that's not good in C++ (I think it can impact perf?).

src/extra/gfx/material_utils.hpp
```diff
@@ -76,15 +96,46 @@ inline void initConstantShaderParams(MaterialParameters &out, SHTable &paramsTab
...
-  while (paramsTable.api->tableNext(paramsTable, &it, &key, &value)) {
+  texturesTable.api->tableGetIterator(texturesTable, &it);
+  while (texturesTable.api->tableNext(texturesTable, &it, &key, &value)) {
+    TexturePtr texture;
+    varToTexture(value, texture);
+    out.set(key, texture);
+  }
+}
+
+inline void initReferencedShaderParams(SHContext *shContext, std::vector<SHBasicShaderParameter> &outParams, SHTable &inTable) {
```
Kryptos-FR, line 111: I would have the *out* parameter always last (or always first) if possible for consistency across multiple functions.

src/extra/gfx/texture.cpp
```diff
@@ -0,0 +1,169 @@
...
+  SHVar activate(SHContext *shContext, const SHVar &input) {
+    auto &image = input.payload.imageValue;
+
+    textureVar.valueType = SHType::Object;
+    textureVar.payload.objectTypeId = Types::TextureTypeId;
+    textureVar.payload.objectVendorId = gfx::VendorId;
+    textureVar.payload.objectValue = &texture;
+
+    if (!texture)
+      texture = std::make_shared<Texture>();
```
Kryptos-FR, line 50: I would move those 2 lines above for better readability. It feels a bit weird to pass a pointer to `&texture` before it is actually initialized (is it actually correct anyway?).

Kryptos-FR (changes requested):

src/extra/gfx/material_utils.hpp
guusw: Good catch, should be const yes (It's just a compiler annotation, no perf impact as far as I know)

guusw: Moved out params to last

src/extra/gfx/texture.cpp
guusw: It's a pointer to the shared_ptr (which is owned by the shard) so it's correct but I can move it above

Kryptos-FR (approved):

src/extra/gfx/texture.cpp
```diff
@@ -0,0 +1,170 @@
...
+  static inline shards::Types InputTypes{{CoreInfo::ImageType}};
+  static SHTypesInfo inputTypes() { return InputTypes; }
+  static SHTypesInfo outputTypes() { return Types::Texture; }
+  static SHOptionalString help() { return SHCCSTR("Creates a texture from an image"); }
+
+  TexturePtr texture;
+  OwnedVar textureVar;
+
+  PARAM_VAR(asType_, "Type", "Type to interpret image data as. (Default: UNormSRGB for RGBA8 images, UNorm for other formats)",
+            {Types::TextureType, CoreInfo::NoneType});
```
sinkingsugar, line 36: Do we need none here? Just a texturetype default should be ok no?

src/extra/gfx/texture.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    if (format.pixelFormat == WGPUTextureFormat_Undefined)
+      throw TextureFormatException(componentType, asType);
+
+    auto &inputFormat = Texture::getInputFormat(format.pixelFormat);
+    size_t imageSize = inputFormat.pixelSize * image.width * image.height;
+
+    // Copy the data since we can't keep a reference to the image variable
+    ImmutableSharedBuffer isb(image.data, imageSize);
+    texture->init(format, int2(image.width, image.height), SamplerState(), std::move(isb));
```
sinkingsugar, line 163: Does this support loading a new image every frame efficiently? Like if I wanna render a video etc.
It should ideally

guusw: Yes, only thing I couldn't avoid at the moment is copying out from the image since I didn't find a way to keep that memory alive until when I issue the GPU commands that need that memory.

src/extra/gfx/texture.cpp
guusw: That's the default in that case, right?
In previous PRs it was suggested that I should add NoneType if it can be left empty.

sinkingsugar: I think with enums we don't need None. it's an enum right?
Default will be the first element of enum or 0 (actually you decide C++ side) so far for enums that's what I did
UX wise is better too

src/extra/gfx/texture.cpp
sinkingsugar: It's fine cos it likely can be multithreaded or so

src/extra/gfx/texture.cpp
guusw: Fixed

PR #375, opened 2022-08-31
Stop when detaching and other misc

sinkingsugar:
Wip

### DIFF
--- a/include/utility.hpp
+++ b/include/utility.hpp
@@ -262,10 +262,12 @@ public:
   }
 
   void cleanup() {
-    if (_v.valueType == ContextVar) {
-      SH_CORE::releaseVariable(_cp);
+    if (_cp) {
+      if (_v.valueType == ContextVar) {
+        SH_CORE::releaseVariable(_cp);
+      }
+      _cp = nullptr;
     }
-    _cp = nullptr;
   }
 
   SHVar &operator=(const SHVar &value) {
--- a/src/core/runtime.hpp
+++ b/src/core/runtime.hpp
@@ -267,7 +267,7 @@ inline void prepare(SHWire *wire, SHFlow *flow) {
 
 inline void start(SHWire *wire, SHVar input = {}) {
   if (wire->state != SHWire::State::Prepared) {
-    SHLOG_ERROR("Attempted to start a wire not ready for running!");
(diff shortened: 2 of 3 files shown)


sinkingsugar:
@Kryptos-FR @guusw for knowledge and review please 🙏

src/core/shards/casting.cpp
```diff
@@ -560,6 +560,25 @@ struct StringToBytes {
   }
 };
 
+struct ImageToBytes {
+  static SHTypesInfo inputTypes() { return CoreInfo::ImageType; }
+  static SHTypesInfo outputTypes() { return CoreInfo::BytesType; }
+
+  SHVar activate(SHContext *context, const SHVar &input) {
+    auto pixsize = 1;
+    if ((input.payload.imageValue.flags & SHIMAGE_FLAGS_16BITS_INT) == SHIMAGE_FLAGS_16BITS_INT)
```
guusw, line 569: Can you move this to a helper function, the same code is also used below in Resize::activate

src/core/shards/serialization.cpp
```diff
@@ -341,11 +341,110 @@ struct LoadImage : public FileBase {
   }
 };
 
+struct BytesToImage {
```
guusw, line 344: I wouldn't name this BytesToImage, since ImageToBytes also exists but the can not be used in reverse looking this implementation:
- ImageToBytes outputs the raw pixel data into a byte variable
- BytesToImage reads any image using stbi_load (PNG/JPEG/other container formats, etc.)

Maybe merge is with LoadImage, in a similiar way that WritePNG now also has the ability to write to memory

guusw (changes requested):

src/core/shards/casting.cpp
sinkingsugar: sadly it's a bigger refactor as it's done in many places actually 😅

src/core/shards/serialization.cpp
sinkingsugar: Good point, doing it now

src/core/shards/casting.cpp
sinkingsugar: done but will need to catch more locations

guusw (approved):

PR #385, opened 2022-09-07
Update egui to 0.19.0

Kryptos-FR:
Implement `TextBuffer` trait directly on `Var`.
Remove workaround that was using the `VarTextBuffer` enum.
Fix code following egui update.

### DIFF
--- a/rust/src/shards/gui/widgets/image.rs
+++ b/rust/src/shards/gui/widgets/image.rs
@@ -160,12 +160,12 @@ impl Shard for Image {
 impl Image {
   fn activateImage(&mut self, _context: &Context, input: &Var) -> Result<Var, &str> {
     if let Some(ui) = util::get_current_parent(*self.parents.get())? {
-      let image: &SHImage = input.try_into()?;
-      let ptr = image.data;
+      let shimage: &SHImage = input.try_into()?;
+      let ptr = shimage.data;
       let texture = if ptr != self.prev_ptr {
-        let image: egui::ColorImage = image.into();
+        let image: egui::ColorImage = shimage.into();
         self.prev_ptr = ptr;
-        self.texture.insert(ui.ctx().load_texture("example", image, Default::default())) // FIXME name
+        self.texture.insert(ui.ctx().load_texture(format!("UI.Image: {:p}", shimage.data), image, Default::default()))
       } else {
         self.texture.as_ref().unwrap()
       };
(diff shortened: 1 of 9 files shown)


src/gfx/egui/src/input.rs
```diff
@@ -237,6 +237,7 @@ pub fn translate_raw_input(input: &egui_Input) -> Result<egui::RawInput, Transla
         screen_rect: Some(input.screenRect.into()),
         pixels_per_point: Some(input.pixelsPerPoint),
         max_texture_side: None,
+        has_focus: true, // FIXME
```
Kryptos-FR, line 240: @guusw need to check whether it's correct to set `has_focus` to `true` all the time. We might need some other logic when we have multiple UIs in the same GFX scene.

Note: it is a new field on this struct introduced by egui in this release.

guusw: Good to keep true for now.
We'll filter input events once we have multiple instances but we can set this too then.

guusw (approved):

rust/src/shards/gui/widgets/image.rs
```diff
@@ -165,7 +165,7 @@ impl Image {
       let texture = if ptr != self.prev_ptr {
         let image: egui::ColorImage = image.into();
         self.prev_ptr = ptr;
-        self.texture.insert(ui.ctx().load_texture("example", image))
+        self.texture.insert(ui.ctx().load_texture("example", image, Default::default())) // FIXME name
```
sinkingsugar, line 168: fixme? seems broken

rust/src/shards/gui/widgets/image_button.rs
```diff
@@ -229,7 +229,7 @@ impl Shard for ImageButton {
         let image: &SHImage = input.try_into().unwrap();
         let image: egui::ColorImage = image.into();
 
-        ui.ctx().load_texture("example", image) // FIXME name
+        ui.ctx().load_texture("example", image, Default::default()) // FIXME name
```
sinkingsugar, line 232: fixme? seems broken

sinkingsugar (changes requested):
FIX me to fix

Kryptos-FR:
@sinkingsugar Not sure how to fix that for now as it would need some kind of name generator based on the shard identity. It looks like the name is only used for debugging purpose in `egui` so for now it just makes it less friendly but doesn't break at runtime. I could give it a better name though: like "UI.Image image", or something similar.

sinkingsugar:
> @sinkingsugar Not sure how to fix that for now as it would need some kind of name generator based on the shard identity. It looks like the name is only used for debugging purpose in `egui` so for now it just makes it less friendly but doesn't break at runtime (able to load different images in separate instance of `UI.Image` without trouble). I could give it a better name though: like "UI.Image image", or something similar.

`println!("{:p}", self);` ?

sinkingsugar:
Well maybe format it `Shard: {:p}` etc

PR #388, opened 2022-09-12
UI fixes & improvements

Kryptos-FR:
Some fixes and improvements found while working on claymore samples.

- texture for `ImageButton`
- size and other constraints for panels, window and some widgets
- open hyperlink in browser

----
*Note:* based on #385 which should be merged first.


guusw (approved):
Had a quick look, LGTM

sinkingsugar (approved):
We need rust coverage, can check with @aang114 and @alebaffa on how they did it on clamor. I think won't be easy cos we use gcc...

PR #390, opened 2022-09-13
UI Context stack & 4 character constants

guusw:
Makes the egui context exposed variable a stack.
This way we can have claymore UI host user scripts containing it's own UI.
Additionally I want to make some additional things available within the new `Context` struct, mainly to be able to map a region inside egui back into window coordinates

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -43,6 +43,14 @@ use crate::types::Var;
 use std::convert::TryInto;
 use std::ffi::CString;
 
+pub const fn fourCharacterCode(val: [u8; 4]) -> i32 {
+  let unsigned_val = ((val[0] as u32) << 24 & 0xff000000)
+    | ((val[1] as u32) << 16 & 0x00ff0000)
+    | ((val[2] as u32) << 8 & 0x0000ff00)
+    | ((val[3] as u32) & 0x000000ff);
+  unsigned_val as i32
+}
+
 pub trait IntoVar {
   fn into_var(self) -> Var;
 }
--- a/rust/src/shards/gui/containers/area.rs
+++ b/rust/src/shards/gui/containers/area.rs
@@ -217,11 +217,7 @@ impl Shard for Area {
   }
 
   fn activate(&mut self, context: &Context, input: &Var) -> Result<Var, &str> {
-    let gui_ctx = {
-      let ctx_ptr: &mut EguiNativeContext =
-        Var::from_object_ptr_mut_ref(*self.instance.get(), &EGUI_CTX_TYPE)?;
-      &*ctx_ptr
-    };
+    let gui_ctx = util::get_current_context(&self.instance)?;
 
     let mut failed = false;
     if !self.contents.is_empty() {
--- a/rust/src/shards/gui/containers/panels.rs
+++ b/rust/src/shards/gui/containers/panels.rs
@@ -200,11 +200,7 @@ macro_rules! impl_panel {
             })
             .inner?;
         } else {
-          let gui_ctx = {
-            let ctx_ptr: &mut EguiNativeContext =
-              Var::from_object_ptr_mut_ref(*self.instance.get(), &EGUI_CTX_TYPE)?;
-            &*ctx_ptr
-          };
+          let gui_ctx = util::get_current_context(&self.instance)?;
           $egui_func(EguiId::new(self, 0))
             .show(gui_ctx, |ui| {
               util::activate_ui_contents(context, input, ui, &mut self.parents, &mut self.contents)
@@ -397,11 +393,7 @@ impl Shard for CentralPanel {
       return Ok(*input);
     }
 
-    let gui_ctx = {
-      let ctx_ptr: &mut EguiNativeContext =
-        Var::from_object_ptr_mut_ref(*self.instance.get(), &EGUI_CTX_TYPE)?;
-      &*ctx_ptr
-    };
+    let gui_ctx = util::get_current_context(&self.instance)?;
 
     if let Some(ui) = util::get_current_parent(*self.parents.get())? {
       egui::CentralPanel::default()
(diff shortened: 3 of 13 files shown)


rust/src/shards/gui/mod.rs
```diff
@@ -66,6 +66,9 @@ impl EguiId {
     }
   }
 }
+pub struct Context<'a> {
```
Kryptos-FR, line 69: Missing new line

```suggestion

pub struct Context<'a> {
```

In addition, does it need to be publicly available from outside (e.g. third-party code)?
I would also rename it to `InternalContext` or `EguiContextWrapper` directly here, to not be confused with all the other context types we have 😅.

rust/src/shards/gui/mod.rs
```diff
@@ -79,6 +82,7 @@ struct EguiContext {
   input_translator: egui_gfx::InputTranslator,
 }
 
+
```
Kryptos-FR, line 85: Unnecessary new line.
```suggestion
```

rust/src/shards/gui/containers/area.rs
```diff
@@ -6,6 +6,7 @@ use super::Area;
 use crate::shard::Shard;
 use crate::shards::gui::containers::ANCHOR_TYPES;
 use crate::shards::gui::util;
+use crate::shards::gui::util::get_current_context;
```
Kryptos-FR, line 9: Just a personal preference, but I like to keep the namespace for those helper functions. Same as with `util::activate_ui_contents(context, input, ui...)` we could have `util::get_current_context(&self.instance)`.

rust/src/shards/gui/containers/area.rs
```diff
@@ -217,11 +218,7 @@ impl Shard for Area {
   }
 
   fn activate(&mut self, context: &Context, input: &Var) -> Result<Var, &str> {
-    let gui_ctx = {
-      let ctx_ptr: &mut EguiNativeContext =
-        Var::from_object_ptr_mut_ref(*self.instance.get(), &EGUI_CTX_TYPE)?;
-      &*ctx_ptr
-    };
+    let gui_ctx = get_current_context(&self.instance)?;
```
Kryptos-FR, line 221: See other comment.

```suggestion
    let gui_ctx = util::get_current_context(&self.instance)?;
```

rust/src/shards/gui/containers/panels.rs
```diff
@@ -8,6 +8,7 @@ use super::RightPanel;
 use super::TopPanel;
 use crate::shard::Shard;
 use crate::shards::gui::util;
+use crate::shards::gui::util::get_current_context;
```
Kryptos-FR, line 11: Same.

rust/src/shards/gui/containers/window.rs
```diff
@@ -7,6 +7,7 @@ use crate::shard::Shard;
 use crate::shards::gui::containers::SEQ_OF_WINDOW_FLAGS;
 use crate::shards::gui::containers::WINDOW_FLAGS_TYPE;
 use crate::shards::gui::util;
+use crate::shards::gui::util::get_current_context;
```
Kryptos-FR, line 10: Same.

rust/src/shards/gui/context.rs
```diff
@@ -1,6 +1,8 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /* Copyright © 2022 Fragcolor Pte. Ltd. */
 
+use super::util::with_object_stack_var;
```
Kryptos-FR, line 4: Same.

```suggestion
use super::util;
```

Kryptos-FR (changes requested):
Overall LGTM. A few suggestions.

rust/src/lib.rs
```diff
@@ -43,6 +43,14 @@ use crate::types::Var;
 use std::convert::TryInto;
 use std::ffi::CString;
 
+pub const fn fourCharacterCode(val: [u8; 4]) -> i32 {
```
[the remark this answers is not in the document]
guusw: I believe so, it didn't compile without so I think it's similar to constexpr in c++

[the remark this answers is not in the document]
guusw: yep https://doc.rust-lang.org/reference/const_eval.html

rust/src/shards/gui/util.rs
```diff
@@ -9,36 +10,56 @@ use crate::types::ExposedTypes;
...
-pub(crate) fn activate_ui_contents<'a>(
-  context: &Context,
-  input: &Var,
-  ui: &mut egui::Ui,
-  parents: &mut ParamVar,
-  contents: &mut ShardsVar,
-) -> Result<Var, &'a str> {
-  // pass the ui parent to the inner shards
-  let var = parents.get();
+use super::EGUI_CTX_TYPE;
```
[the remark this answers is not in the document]
guusw: I'm not gonna change all of them here, since a bunch of other gui files also use `use super::<something>`

guusw:
Added a test `src/tests/ui-nested.edn` and also fixed a case where the nested UI would take UI parents from the outer UI (in case of CentralPanel for example)

Also I believe there was a potential leak/corruption issue with the stack variables because they:
- retrieved the seq (by value)
- pushed 
- set it on ParamVar
- run some other code
- pop from the **SAME** seq as in the first step (here might be an issue since the memory could've been reallocated/resized, but it won't be reflected here)
- set it on ParamVar again

Should be fixed by retrieving from the ParamVar again after running the inner code

rust/src/shards/gui/context.rs
```diff
@@ -174,11 +175,18 @@ impl Shard for EguiContext {
     self.contents.warmup(ctx)?;
     self.main_window_globals.warmup(ctx);
     self.parents.warmup(ctx);
+
     // Initialize the parents stack in the root UI.
     // Every other UI elements will reference it and push or pop UIs to it.
-    self.parents.set(Seq::new().as_ref().into());
+    if !self.parents.get().is_seq() {
```
guusw, line 180: I'm not sure if this was an issue but I expect this might clear the stack if some nested UI were to be warmed up from within a (Setup) block in another UI

sinkingsugar (approved):

PR #395, opened 2022-09-15
UI code editor

Kryptos-FR:
A `TextInput` with support for syntax highlighting.

It is currently behind a `"code_editor"` feature flag. We might decide to have it enabled by default (as part of `"shards"`) or only when building an app such as claymore.

Uses the `Syntech` crate and the current implementation is mainly based on the `egui_demo_lib` from the `egui` repository for the layouting part.

---

Part of fragcolor-xyz/management#31

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -43,6 +43,7 @@ chrono = { version = "0.4", optional = true }
 chacha20poly1305 = { version = "0.9.0", optional = true }
 egui = { version = "0.19.0", optional = true }
 egui-gfx = { path = "../src/gfx/egui", optional = true }
+syntect = { version = "5.0.0", default-features = false, features = ["default-fancy"], optional = true }
 half = { version = "2.0.0" }
 tract-onnx = { version = "0.17.7", optional = true }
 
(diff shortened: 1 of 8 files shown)


rust/Cargo.toml
```diff
@@ -84,6 +85,7 @@ shards = ["reqwest",
           "chrono",
           "egui",
           "egui-gfx"]
+code_editor = ["syntect"]
```
sinkingsugar, line 88: fine to be on by default for now.

guusw (approved):
Looked quickly, good for me.

One issue I have though is that the code widget doesn't fill the window height.
Tried playing around a bit and couldn't get the window with a code editor inside of it to resize beyond a maximum size.

Kryptos-FR:
@guusw god point. I have a hardcoded `.desired_rows(10)` which was copied from the egui sample. I will try play with it (eventually expose it) to see how it impacts the layout.

We might need to have the code editor inside a scroll area. Either the user would need to do it manually (using `(UI.ScrollArea)` or it could also be built-in the code editor shard.

PR #398, opened 2022-09-17
Output rust binaries to build/target dir. Argument error checking

guusw:
Wanted to unify the location of build artifacts.
This way if someone has issues we can tell them to clear the build folder.
Right now rust still outputs into `target`, `src/gfx/wpgu/target`, etc.

Might need some adjustments for the CI rust cache to work


sinkingsugar:
close @guusw ?

PR #400, opened 2022-09-19
Render Target with UI Input

guusw:
This adds basic render targets along with some changes to allow UI input to work when it's rendered inside of a render target.


rust/src/shards/gui/widgets/image.rs
```diff
@@ -159,6 +162,11 @@ impl Shard for Image {
 }
 
 impl Image {
+  fn get_scale(scale_var: &ParamVar, ui: &Ui) -> Result<egui::Vec2, &'static str> {
```
guusw, line 165: Decided to draw images at the original size, so if an image is 512x512 it'll be drawn at that size regardless of the UI scale.
Maybe we need a flag for this though, where by default it does automatically scale but can be disabled.

rust/src/shards/gui/widgets/render_target.rs
```diff
@@ -0,0 +1,202 @@
...
+  static ref TEXTURE_TYPES: Vec<Type> = vec![*TEXTURE_TYPE];
+  static ref RENDER_TARGET_PARAMETERS: Parameters = vec![(
+    cstr!("Scale"),
+    cstr!("Scaling to apply to the source texture"),
+    FLOAT2_VAR_SLICE,
+  )
+    .into(),];
+}
+
+impl Default for RenderTarget {
```
guusw, line 48: Need to de-duplicate some shared code with image.rs

src/gfx/renderer_types.hpp
```diff
@@ -1,12 +1,38 @@
-#ifndef GFX_RENDERER_TYPES
-#define GFX_RENDERER_TYPES
+#ifndef CA95E36A_DF18_4EE1_B394_4094F976B20E
```
guusw, line 1: tl;dr Moved internal structs used within renderer.cpp to this header.
Also added some new structs namely:
- PipelineDrawables
- RenderTargetData
- RenderGraphNode/RenderGraph/RenderGraphEvaluator

src/gfx/view_texture.hpp
```diff
@@ -1,69 +0,0 @@
-#ifndef GFX_VIEW_TEXTURE
-#define GFX_VIEW_TEXTURE
-
-#include "context_data.hpp"
-#include "gfx_wgpu.hpp"
-#include "linalg.hpp"
-
-namespace gfx {
-
-struct ViewTexture : public ContextData {
```
guusw, line 10: Merged functionality into Texture by adding the conditional resize function

Kryptos-FR (changes requested):
Overall LGTM. A few remarks and one refactor suggestion.

rust/src/shards/gui/widgets/image.rs
sinkingsugar: There was scaling params in the wrong imgui implmentation, why we don't follow that? please...

rust/src/shards/gui/widgets/image_util.rs
```diff
@@ -0,0 +1,111 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright © 2022 Fragcolor Pte. Ltd. */
+
+use egui::TextureHandle;
+use egui::TextureId;
+use egui::Ui;
+use egui::Vec2;
+use std::ptr::null_mut;
```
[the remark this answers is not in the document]
sinkingsugar: rust_analyzer formatter should do this automatically also...

rust/src/shards/gui/widgets/image.rs
```diff
@@ -156,23 +156,10 @@ impl Shard for Image {
...
-          Default::default(),
-        ))
-      } else {
-        self.texture.as_ref().unwrap()
-      };
-
-      let scale: (f32, f32) = self.scale.get().try_into()?;
-      let scale: egui::Vec2 = scale.into();
-      ui.image(texture, scale * texture.size_vec2());
+      let texture = image_util::ui_image_cached(&mut self.cached_ui_image, input, ui)?;
```
sinkingsugar, line 159: what's cached?

rust/src/shards/gui/widgets/image.rs
guusw: Ah sorry with `UI Scale` I meant the scale the UI is drawn at relative to how many pixels it covers, e.g. 2x on mac displays, 1.5x if you set display scaling to 150% in windows settings.

The scale parameter is still applied, just that scale of 1.0 now corresponds to 1:1 image to screen pixels

rust/src/shards/gui/widgets/image.rs
guusw: The loaded egui texture if you create it from a shards image, same as previous behavior but I wrapped it so it could be reused for image_button, etc.

rust/src/shards/gui/widgets/image_util.rs
```diff
@@ -0,0 +1,111 @@
...
+    cached_image.prev_ptr = ptr;
+    cached_image.texture_handle.insert(ui.ctx().load_texture(
+      format!("UI.Image: {:p}", shimage.data),
+      image,
+      Default::default(),
+    ))
+  } else {
+    cached_image.texture_handle.as_ref().unwrap()
+  })
+}
```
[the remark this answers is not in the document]
guusw: Did that and also renamed them a bit

rust/src/shards/gui/widgets/image_util.rs
```diff
@@ -0,0 +1,111 @@
...
+
+pub fn ui_image_texture(input: &Var) -> Result<(TextureId, Vec2), &'static str> {
+  let ptr = unsafe { input.payload.__bindgen_anon_1.__bindgen_anon_1.objectValue as u64 };
```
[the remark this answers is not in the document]
guusw: Actually didn't need that cast, can just use the ptr below and cast it to u64.
That one is already behind an assert

rust/src/shards/gui/widgets/image_util.rs
```diff
@@ -0,0 +1,111 @@
...
+
+lazy_static! {
+  pub static ref TEXTURE_TYPE: Type = Type::object(FRAG_CC, TextureCC);
+  pub static ref TEXTURE_OR_IMAGE_TYPES: Vec<Type> = vec![common_type::image, *TEXTURE_TYPE];
+}
+
+pub fn get_scale(scale_var: &ParamVar, ui: &Ui) -> Result<egui::Vec2, &'static str> {
+  let scale: (f32, f32) = scale_var.get().try_into()?;
+  Ok(egui::vec2(scale.0, scale.1) / ui.ctx().pixels_per_point())
+}
```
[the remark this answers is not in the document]
guusw: I removed the scaling, I only really needed it for the render_target widget, which now does it by itself.
So this function is now just there to turn Var => egui::vec2 just so it's a bit less code duplicated over the three widgets

rust/src/shards/gui/widgets/image.rs
guusw: Undid that scaling change btw, so it's back to what it was as it makes more sense to just scale images with the UI by default. Only needed unscaled images for the render target widget which now does that by itself

rust/src/shards/gui/widgets/image_util.rs
```diff
@@ -25,6 +19,8 @@ use crate::types::Type;
 use crate::types::Types;
 use crate::types::Var;
 use crate::types::FRAG_CC;
+use std::intrinsics::transmute;
```
Kryptos-FR, line 22: Doesn't seem required.

rust/src/shards/gui/widgets/render_target.rs
```diff
@@ -31,6 +26,7 @@ use crate::types::Type;
 use crate::types::Types;
 use crate::types::Var;
 use crate::types::FRAG_CC;
+use std::borrow::BorrowMut;
```
Kryptos-FR, line 29: Doesn't seem required.

Sometimes while typing, rust-analyzer adds useless imports.

Kryptos-FR (approved):
LGTM

rust/src/shards/gui/widgets/image_util.rs
```diff
@@ -0,0 +1,104 @@
...
+      )
+    };
+
+    if image.flags & SHIMAGE_FLAGS_PREMULTIPLIED_ALPHA as u8
+      == SHIMAGE_FLAGS_PREMULTIPLIED_ALPHA as u8
+    {
+      let pixels = rgba
+        .chunks_exact(4)
+        .map(|p| egui::Color32::from_rgba_premultiplied(p[0], p[1], p[2], p[3]))
+        .collect();
```
sinkingsugar, line 98: Allocation, write it simple.
Those things are like `linq` sadly...

src/extra/gfx/view.cpp
```diff
@@ -0,0 +1,167 @@
...
+    viewStack.pop();
+    inputStack.pop();
+
+    return Var::Empty;
+  }
+};
+
+void registerViewShards() {
+  REGISTER_SHARD("GFX.View", ViewShard);
+  REGISTER_SHARD("GFX.PushView", PushViewShard);
```
sinkingsugar, line 164: I don't like Push/Pop with shards.. it's complicated cos you can resume other coroutines and what not in shards. Hard to keep track. One Wire can stop (where you push) while other resumes (where you pop).
If possible I prefer (-> contexts.

src/gfx/checked_stack.hpp
```diff
@@ -0,0 +1,37 @@
...
+  // throws on unmatched push/pop
+  // if caught stack will be restored to the passed marker
+  void restoreMarkerChecked(Marker marker) {
+    size_t numItemsBeforeFixup = size();
+
+    // Fixup before throw in case error is handled
+    resize(marker);
+
+    if (numItemsBeforeFixup != marker) {
+      throw std::runtime_error(fmt::format("Stack imbalance ({} elements, expected {})", numItemsBeforeFixup, marker));
```
sinkingsugar, line 27: Meh, and we give this complexity to our users?

sinkingsugar (changes requested):
Few things to discuss so far. Looks great though of course!

rust/src/shards/gui/widgets/image_util.rs
Kryptos-FR: I thought rust compiler was optimizing them away. I saw a few posts on reddit where it showed that the resulting assembly in release was optimal in most cases.

src/extra/gfx/view.cpp
guusw: Yeah fair, I'm not sure why I didn't do that from the start, will change it.

src/gfx/checked_stack.hpp
Kryptos-FR: Should be a debug assert only maybe? @guusw was it to catch issue when developing the feature or can it still happen at runtime?

rust/src/shards/gui/widgets/image_util.rs
guusw: Not too much point in rewriting it, I think we can merge it later with what's being done by GFX.Texture (less image conversions/copies)

src/gfx/checked_stack.hpp
guusw: I'll change it to `(View (-> content...))` so don't need the checks anymore

rust/src/shards/gui/widgets/image_util.rs
sinkingsugar: I think it's good preventive optimisation here. We should have some hard rules, e.g. no `collect`.
Btw not sure how compiler should optimise.. it needs a dynamic array.
You should have a Vec in the self struct you use as cache, so you can clear (cheap) and refill every frame.

sinkingsugar: That said if it's a one off.. and not supposed to run every frame we can keep it but add a comment about why you allow this `collect`

Kryptos-FR: > Btw not sure how compiler should optimise.. it needs a dynamic array.

`chunks_exact` returns an iterator that implement both `TrustedLen` and `ExactSizeIterator` traits. With that knowledge, `collect()` can use specialization and pre-allocate the memory knowing the exact size.

src/gfx/checked_stack.hpp
guusw: I renamed the shard to GFX.Region to avoid confusion with the existing GFX.View (also want to rename that one to Camera later)

guusw: It also wraps content instead of being separate push/pop

src/extra/gfx/texture.cpp
```diff
@@ -166,11 +164,91 @@ struct TextureShard {
...
 
     return textureVar;
   }
 };
 
-void registerTextureShards() { REGISTER_SHARD("GFX.Texture", TextureShard); }
+void registerTextureShards() {
+  REGISTER_SHARD("GFX.Texture", TextureShard);
+  REGISTER_SHARD("GFX.RenderTarget", RenderTargetShard);
+  REGISTER_SHARD("GFX.RenderTargetTexture", RenderTargetTextureShard);
```
sinkingsugar, line 252: Whats the difference between RenderTarget and RenderTargetTexture ? This is weird, isn't a render target just a texture you can bind to normal draw calls just like a texture?

src/extra/gfx/view.cpp
```diff
@@ -0,0 +1,157 @@
...
+      inputItem.windowMapping = input::WindowSubRegion{
+          .region = region,
+      };
+
+      // Safe to automatically set reference size based on region
+      float2 regionSize = float2(region.z - region.x, region.w - region.y);
+      viewItem.referenceSize = int2(linalg::floor(float2(regionSize)));
+    }
+
+    auto &mwg = getMainWindowGlobals();
```
sinkingsugar, line 136: do we only have mainwindow?

src/gfx/checked_stack.hpp
sinkingsugar: > I'll change it to `(View (-> content...))` so don't need the checks anymore

did that happened?

sinkingsugar: why we need still stack checks ?

src/gfx/egui/input.cpp
```diff
@@ -65,143 +65,161 @@ static egui::ModifierKeys translateModifierKeys(SDL_Keymod flags) {
...
+    deferQueue.emplace_back([this, eventIdx = events.size() - 1, stringIdx = strings.size() - 1]() {
+      events[eventIdx].compositionUpdate.text = strings[stringIdx].c_str();
+    });
+    break;
+  }
+  case SDL_TEXTINPUT: {
+    auto &ievent = sdlEvent.text;
 
-      newEvent(InputEventType::CompositionUpdate);
+    if (imeComposing) {
```
sinkingsugar, line 176: oh wow nice

rust/src/shards/gui/widgets/image_util.rs
guusw: #420

src/extra/gfx/texture.cpp
guusw: Ah yeah I will change it later but RenderTarget groups multiple attachments (color,depth,etc.) and RenderTargetTexture retrieves a single one as a texture

src/extra/gfx/view.cpp
guusw: Yes, but it's a member from the base consumer so you could have a separate wire with another window if you really wanted

src/gfx/checked_stack.hpp
guusw: Removed scoped checks, but kept one at the end of frame to catch implementation errors

src/extra/gfx/texture.cpp
sinkingsugar: I see hmm ok leave like this for now

sinkingsugar:
ping when green I will finalize a review.

src/extra/gfx/texture.cpp
```diff
@@ -166,11 +164,91 @@ struct TextureShard {
...
+    }
+
+    return renderTargetVar;
+  }
+};
+
+struct RenderTargetTextureShard {
+  static SHTypesInfo inputTypes() { return Types::RenderTarget; }
+  static SHTypesInfo outputTypes() { return Types::Texture; }
+  static SHOptionalString help() { return SHCCSTR("Retrieve a named attachment from a render target"); }
```
sinkingsugar, line 209: named attachment so far are hardcoded main and depth no?

src/extra/gfx/view.cpp
```diff
@@ -0,0 +1,157 @@
...
+using namespace shards;
+
+namespace gfx {
+void SHView::updateVariables() {
+  if (viewTransformVar.isVariable()) {
+    view->view = shards::Mat4(viewTransformVar.get());
+  }
+}
+
+struct ViewShard {
```
sinkingsugar, line 19: add help

src/gfx/renderer.cpp
```diff
@@ -16,228 +18,47 @@
...
+#include "render_target.hpp"
 #include "log.hpp"
+#include "geom.hpp"
 #include <algorithm>
 #include <magic_enum.hpp>
 #include <spdlog/spdlog.h>
 
 #define GFX_RENDERER_MAX_BUFFERED_FRAMES (2)
 
-namespace gfx {
```
sinkingsugar, line 27: I see a lot of rewriting which is good but we need an architecture document soon. can use gitbook

sinkingsugar (changes requested):
few points

src/extra/gfx/texture.cpp
guusw: Yeah in this PR they are. In the other PR #436 you can define whatever you want, although there's still a hidden contract that they're called "color" and "depth" so passes can communicate

src/gfx/renderer.cpp
guusw: Yeah I can try something

src/extra/gfx/view.cpp
guusw: Added

Kryptos-FR (approved):

sinkingsugar (approved):
Approving but please fix coverage and please before merge can you add somewhere a big comment block with some architecture remarks? Just in case, we need to be able to navigate a bit :) doesn't have to be deep 🙏

guusw:
I started a document, I'll add some documentation on the internals there too

sinkingsugar:
Any idea what's going on with Codecov?

Kryptos-FR:
> Any idea what's going on with Codecov?

It looks like GitHub CI workers are having a lot of networking issue recently. I had to restart a bunch of failed jobs because of that. And similarly sometimes it fails to upload the coverage to codecov. Or more specifically, codecov is not able to ping back GitHub's API to get some data:
> {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

It's possible we need to review the way we use the codecov action. Instead of `bash <(curl -s https://codecov.io/bash)` we should look into using their officially supported action: https://github.com/codecov/codecov-action.

PR #404, opened 2022-09-21
Clip/Scissor rectangles

guusw:
Certain UI elements need scissor rectangles to render correctly, added it to drawables


Kryptos-FR:
Is it related to clipping issue with scrollbars for instance?

Kryptos-FR (approved):

PR #410, opened 2022-09-27
add a flowIndex var to shards

sinkingsugar:


guusw:
What are these going to be used for?

sinkingsugar:
> What are these going to be used for?

debug info/ui etc, currently we are not tracking properly things. still wip btw

PR #412, opened 2022-09-27
Reenable most rust warnings

guusw:
Removes the allow attributes from the root module.

Also tries to address most of the generated warnings.
Applied `cargo fix` automatically, leaving the manual renaming for later.


sinkingsugar:
close @guusw ?

PR #418, opened 2022-10-04
UI style fixes

Kryptos-FR:
**Description**
Some fixes mostly around styling:
- fix an issue with unset position on `UI.Area` and `UI.Window`
- add a `:Style` parameter on `UI.Label`, `UI.Button`, `UI.RadioButton`, sliders, `UI.Combo`, `UI.Hyperlink`
- fix `UI.Style` not working at top level

### DIFF
--- a/rust/src/shards/gui/misc/style.rs
+++ b/rust/src/shards/gui/misc/style.rs
@@ -4,10 +4,10 @@
 use super::Style;
 use crate::shard::Shard;
 use crate::shards::gui::util;
+use crate::shards::gui::ANY_TABLE_SLICE;
 use crate::shards::gui::EGUI_UI_SEQ_TYPE;
 use crate::shards::gui::PARENTS_UI_NAME;
 use crate::shardsc::SHColor;
-use crate::types::common_type;
 use crate::types::Context;
 use crate::types::ExposedInfo;
 use crate::types::ExposedTypes;
@@ -19,7 +19,7 @@ use crate::types::Var;
 use crate::types::ANY_TYPES;
 
 lazy_static! {
-  static ref INPUT_TYPES: Types = vec![common_type::any_table, common_type::any_table_var];
+  static ref INPUT_TYPES: Types = ANY_TABLE_SLICE.into();
 }
 
 macro_rules! apply_style {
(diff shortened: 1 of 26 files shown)


rust/src/shards/gui/widgets/link.rs
```diff
@@ -0,0 +1,194 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright © 2022 Fragcolor Pte. Ltd. */
+
+use super::Link;
```
guusw, line 4: Maybe we can just have link and simplify it to work for either url's if you pass :Url or otherwise by passing :Action
The only difference in behavior is what happens when clicking

guusw (approved):

guusw:
Some feedback, other than that looks good

guusw:
ui.edn test fails though

Kryptos-FR:
> ui.edn test fails though

And others samples that uses a `UI.Button`. I will fix them.

rust/src/shards/gui/misc/style.rs
```diff
@@ -91,343 +95,353 @@ impl Shard for Style {
...
+          &mut style.visuals.widgets.inactive,
+          &widgets,
+          cstr!("inactive"),
+        )?;
+        Style::applyWidgetVisuals(
+          &mut style.visuals.widgets.hovered,
+          &widgets,
+          cstr!("hovered"),
+        )?;
+        Style::applyWidgetVisuals(&mut style.visuals.widgets.active, &widgets, cstr!("active"))?;
```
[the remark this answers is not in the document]
Kryptos-FR: Good point. I actually already made that change in the next branch `ui/console` where I moved most of those functions into a separate `style_util.rs` file.

rust/src/shards/gui/widgets/link.rs
Kryptos-FR: Hyperlink and Link have other differences:

|  | Input type | Output type |
|---|---|---|
| Hyperlink | String | String (input passthrough) |
| Link | Any | bool (whether clicked this frame)l |

What happens if both `:Url` and `:Action` are set: should we open the link AND execute the action?

PR #419, opened 2022-10-04
UI console

Kryptos-FR:
Usage:

```clj
(CaptureLog
 :Size 24
 :MinLevel "info"
 :Pattern "%^[%l]%$ [%Y-%m-%d %T.%e] %v")
(UI.Console)
```


Kryptos-FR:
Currently it is rather unoptimized (with the exception of a cache of strings).

Ideas for further optimizations:
* return a Seq of string instead of a single concatenated string.
* return the Seq along the number of new logs (as keys/values in a table)
* the console can then recompute the layout only for the new logs (and drop the old ones), while shifting the ones already existing (and kept)

docs/samples/shards/General/CaptureLog/1.edn
```diff
@@ -0,0 +1,28 @@
...
+     .ui-draw-queue
+     (->
+      (UI.CentralPanel
+       (->
+        (UI.Button "Say Hello" (Msg "Hello!"))
+        (CaptureLog
+         :Size 24
+         :MinLevel "info"
+         :Pattern "%^[%l]%$ [%Y-%m-%d %T.%e] %v")
+        (UI.Console)))))
```
sinkingsugar, line 22: how does this work? Shouldn't CaptureLog be BLOCKING (via coroutines)?

src/core/shards/logging.cpp
```diff
@@ -100,30 +103,166 @@ struct Msg : public LoggingBase {
...
+      : q_{n_items}, min_level_(min_level) {}
+
+  inline size_t generation() { return _generation; }
+
+  inline bool is_dirty() { return _last_generation < _generation; }
+
+  std::vector<std::string> last_formatted() {
+    std::lock_guard<Mutex> lock(spdlog::sinks::base_sink<Mutex>::mutex_);
+
+    if (is_dirty()) {
```
sinkingsugar, line 121: can't dirty be atomic so to avoid locking at all?

src/core/shards/logging.cpp
```diff
@@ -100,30 +103,166 @@ struct Msg : public LoggingBase {
...
+    }
+  }
+
+  SHVar activate(SHContext *context, const SHVar &input) {
+    if (_ring) {
+      if (_ring->is_dirty()) {
+        auto msgs = _ring->last_formatted();
+
+        _s.clear();
+        for (auto i = msgs.begin(); i != msgs.end(); ++i) {
```
sinkingsugar, line 238: HMM I wonder if we should implement it instead using suspends but I can see this being ok too

docs/samples/shards/General/CaptureLog/1.edn
Kryptos-FR: Why should it be blocking? It uses a ring buffer internally.

src/core/shards/logging.cpp
Kryptos-FR: We need locking because `q_` could otherwise be written to. I should move the lock inside the branch maybe.

Though you are right that `is_dirty()` isn't atomic so the lock might also be needed for it. I could add a `bool` that is recomputed when necessary (same as previous commit), which makes it atomic at the expense of one additional field.

sinkingsugar: what I mean is lock only if dirty basically, as dirty can be an atomic bool

docs/samples/shards/General/CaptureLog/1.edn
Kryptos-FR: Added an option to suspend.

src/core/shards/logging.cpp
Kryptos-FR: Use a simple boolean for now which should be atomic.

src/core/shards/logging.cpp
Kryptos-FR: Generates a sequence and uses suspend.

src/core/shards/logging.cpp
```diff
@@ -100,30 +103,166 @@ struct Msg : public LoggingBase {
...
+    }
+    return res;
+  }
+
+  void warmup(SHContext *context) {
+    auto logger = spdlog::default_logger();
+    assert(logger);
+    auto sink = logger->sinks().at(0);
+    assert(sink);
+    auto dist_sink = std::dynamic_pointer_cast<spdlog::sinks::dist_sink_mt>(sink);
```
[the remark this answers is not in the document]
Kryptos-FR: You can't safely add/remove a sink after the logging has started. It is only possible through the `dist_sink` (which was made for this purpose) which I need to retrieve from `logger->sinks()`.

docs/samples/shards/General/CaptureLog/1.edn
```diff
@@ -0,0 +1,28 @@
...
+    .ui-draw-queue (GFX.ClearQueue)
+
+    (UI
+     .ui-draw-queue
+     (->
+      (UI.CentralPanel
+       (->
+        (UI.Button "Say Hello" (Msg "Hello!"))
+        (CaptureLog
+         :Size 24
```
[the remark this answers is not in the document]
Kryptos-FR: I made `CaptureLog` output a sequence.

For the moment `(UI.Console)` still takes a single string as input. That's because `egui::TextEdit` only works with a single string entry. We could made our own widget and deal with a list of entries instead but that's a lot of work (though we could partially copy-paste `egui::TextEdit` implementation. So for now, I'll keep it simple.

src/core/shards/logging.cpp
```diff
@@ -100,30 +103,162 @@ struct Msg : public LoggingBase {
...
+    }
+  }
+
+  SHVar activate(SHContext *context, const SHVar &input) {
+    if (likely((bool)_ring)) {
+      while (_suspend && !_ring->is_dirty()) {
+        SH_SUSPEND(context, 0);
+      }
+
+      if (_ring->is_dirty()) {
```
sinkingsugar, line 227: I think we can avoid checking dirty flag twice here (as it potentially is atomic it will introduce barriers so better do once)

Kryptos-FR: Good point.

Kryptos-FR: Actually it is needed to do so when  `_suspend` is `false`.

However, since the `:Suspend` parameter doesn't change after it is assigned, maybe it should use two different `activate` overrides.

sinkingsugar: yeah

sinkingsugar:
can you add a headless test for the capture one and add it to valgrind build please?

PR #429, opened 2022-10-14
Fix Issues with the `Erase` shard #316

### LINKED ISSUE #316: Issues with the `Erase` shard
**Describe the bug**
- Issue 1: `Erase` gives an error if the input to it doesn't match the data structure of its `:Name` entity or the indices/keys passed to it.
- Issue 2: When erasing elements from a sequence `Erase` gives an error if the indices sequence is not in descending order
 

**To Reproduce**

Issue 1
```clj
{:k1 100 :k2 200 :k3 300} >= .tab
"Hello World"
(Erase ["k3" "k1"] .tab)
;; => [error] [2022-07-13 21:36:34.538] [T-10164] [runtime.cpp::1090] Error composing shard: Erase, invalid indices or malformed input., wire: main
```
Issue 2: case 1 (indices sequence sorted in ascending order)
```clj
[10 20 30 40 50] >= .seq
(Erase [1 2 3] .seq) ;; indices sorted in ascending order => error
;;[error] [2022-07-13 21:46:33.026] [T-16264] [runtime.cpp::791] Shard activation error, failed shard: Erase, error: Comparison < between two different value types
```

Issue 2: case 2 (indices sequence randomly sorted)
```clj
[10 20 30 40 50] >= .seq
(Erase [3 1 2] .seq) ;; indices sequence randomly sorted
.seq (Log) ;; .seq incorrectly filtered => [10, 30] ; expected .seq => [10, 50]
```

**Expected behavior**
- Issue 1: `Erase` should not validate the input as it does not use the input value for processing.
- Issue 2: `Erase` should work correctly for  indices sequence sorted by ascending order (case 1) as well as sorted randomly (case 2)

sinkingsugar:
Fix #316

### DIFF
--- a/src/core/shards/core.cpp
+++ b/src/core/shards/core.cpp
@@ -861,7 +861,7 @@ private:
 
 struct Erase : SeqUser {
   static SHOptionalString help() {
-    return SHCCSTR("Deletes identfied element(s) from a sequence or key-value pair(s) from a table.");
+    return SHCCSTR("Deletes identified element(s) from a sequence or key-value pair(s) from a table.");
   }
 
   static SHTypesInfo inputTypes() { return CoreInfo::AnyType; }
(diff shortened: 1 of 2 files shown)


src/core/shards/core.cpp
```diff
@@ -942,7 +961,7 @@ struct Erase : SeqUser {
 
   SHVar activate(SHContext *context, const SHVar &input) {
     const auto &indices = _indices.get();
-    if (unlikely(_isTable && _target->valueType == Table)) {
+    if (unlikely(_target->valueType == Table)) {
```
Kryptos-FR, line 964: Should we keep `unlikely` here?

If I get it correctly, it means that this shard is optimized (fast path) for sequence, and will pay the cost of a bad branch prediction for a table. Am I right?

Kryptos-FR (approved):
LGTM.

Just a remark about the use of `unlikely` which I think might not be optimal here, but it was already there in the previous code, so I'm fine with keeping it for now.

guusw (approved):
LGTM

PR #433, opened 2022-10-18
Mark GUI shards as deprecated

Kryptos-FR:
**Description**
`ImGui`-based shards (prefixed by `GUI`) have mostly be replaced by `egui`-based one (prefixed by `UI`). Mark them as obsolete and display a warning in the API documentation.


ChelseaChX (approved):
Looks good! Serves as a nice way to let people know which shards are depreciated, especially if they are not in the loop for new updates or changes.

PR #435, opened 2022-10-18
UI HexViewer

Kryptos-FR:
**Description**
We had an hex viewer implemented with `ImGui`. Replace with a similar one implemented with `egui`.

**Test plan**
- [x] should have the same features

**Usage**
*See sample*

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -43,7 +43,7 @@ chrono = { version = "0.4", optional = true }
 chacha20poly1305 = { version = "0.9.0", optional = true }
 egui = { version = "0.19.0", optional = true }
 egui-gfx = { path = "../src/gfx/egui", optional = true }
-egui_memory_editor = { version = "0.2.0", optional = true }
+egui_memory_editor = { version = "0.2.1", optional = true }
 syntect = { version = "5.0.0", default-features = false, features = ["default-fancy"], optional = true }
 half = { version = "2.0.0" }
 tract-onnx = { version = "0.17.7", optional = true }
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -43,6 +43,7 @@ chrono = { version = "0.4", optional = true }
 chacha20poly1305 = { version = "0.9.0", optional = true }
 egui = { version = "0.19.0", optional = true }
 egui-gfx = { path = "../src/gfx/egui", optional = true }
+egui_memory_editor = { version = "0.2.0", optional = true }
 syntect = { version = "5.0.0", default-features = false, features = ["default-fancy"], optional = true }
 half = { version = "2.0.0" }
 tract-onnx = { version = "0.17.7", optional = true }
@@ -88,9 +89,11 @@ shards = ["reqwest",
           "egui",
           "egui-gfx",
           "code_editor",
+          "hex_viewer",
           "tract-onnx",
           "enum-map"]
 code_editor = ["syntect"]
+hex_viewer = ["egui_memory_editor"]
 dllshard = ["dlopen"]
 scripting = []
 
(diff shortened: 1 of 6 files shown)


rust/src/shards/gui/widgets/hex_viewer.rs
```diff
@@ -0,0 +1,192 @@
...
+        std::slice::from_raw_parts_mut(data, len)
+      };
+
+      let mut mem_editor = MemoryEditor::new().with_address_range("All", 0..mem.len());
+
+      if let Some(options) = &self.editor_options {
+        mem_editor = mem_editor.with_options(options.clone());
+      }
+      mem_editor.draw_editor_contents_read_only(ui, mem, |mem, address| mem[address].into());
+      self.editor_options = Some(mem_editor.options.clone());
```
guusw, line 185: I think you can keep the MemoryEditor instance on this shard, that way you also don't have to get/restore the options every frame.

guusw: Just update the address range potentially

guusw (changes requested):

rust/Cargo.toml
```diff
@@ -88,9 +89,11 @@ shards = ["reqwest",
           "egui",
           "egui-gfx",
           "code_editor",
+          "hex_viewer",
           "tract-onnx",
           "enum-map"]
 code_editor = ["syntect"]
+hex_viewer = ["egui_memory_editor"]
```
sinkingsugar, line 96: is it building by default?

rust/src/shards/gui/widgets/hex_viewer.rs
```diff
@@ -0,0 +1,192 @@
...
+          SHType_String => (
+            input.payload.__bindgen_anon_1.__bindgen_anon_2.stringValue as *mut u8,
+            input.payload.__bindgen_anon_1.__bindgen_anon_2.stringLen as usize,
+          ),
+          _ => unreachable!(),
+        };
+        std::slice::from_raw_parts_mut(data, len)
+      };
+
+      let mut mem_editor = MemoryEditor::new().with_address_range("All", 0..mem.len());
```
sinkingsugar, line 179: anything new is hyper misleading, this is just the way UI api works (badly) right? doesn't mean heavy allocations...

rust/src/shards/gui/widgets/hex_viewer.rs
sinkingsugar: yeah.... I think that needs to be done

sinkingsugar (changes requested):

rust/src/shards/gui/widgets/hex_viewer.rs
Kryptos-FR: Unfortunately it can't be done. The address range is not writeable, except during construction.

sinkingsugar: I mean all the `new` around egui

rust/src/shards/gui/widgets/hex_viewer.rs
Kryptos-FR: `with_address_range` only available during construction: it moves `Self` (doesn't return a reference), so if we keep a reference to the `MemoryEditor`, we can't modify the range afterwards. It's a limitation of this library currently.

rust/Cargo.toml
Kryptos-FR: Yes.

rust/src/shards/gui/widgets/hex_viewer.rs
Kryptos-FR: Issue opened: https://github.com/Hirtol/egui_memory_editor/issues/5

Kryptos-FR: PR created: Hirtol/egui_memory_editor#6

rust/src/shards/gui/widgets/hex_viewer.rs
sinkingsugar: I think we discussed this on discord

sinkingsugar (approved):

Issue #172, opened 2022-03-02
Render graph & post effect passes


guusw:
Implement functionality as demonstrated by
https://github.com/fragcolor-xyz/FIPs/blob/FIP-1/FIPs/FIP-1.md#post-effects
and
https://github.com/fragcolor-xyz/FIPs/blob/FIP-1/FIPs/FIP-1.md#velocity-pass

TODO: 
- [x] Add EffectPass render step
- [x] Add definition of frame buffers to render steps
- [x] Add input/output color targets to effect passes
- [x] Add output color targets to drawable passes

guusw:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/chainblocks/172?planning-poker) with ZenHub @Kryptos-FR @satendrapalsingh @sinkingsugar

PR #436, opened 2022-10-18
Render graph shards

guusw:
Implementation of #172  into the existing and new shards


src/extra/gfx/shader/core_shards.hpp
```diff
@@ -17,99 +17,116 @@ namespace gfx {
...
+    if (shard->_indices.valueType == SHType::Int) {
+      result = componentNames[shard->_indices.payload.intValue];
+    } else if (shard->_indices.valueType == SHType::Seq) {
+      SHSeq indices = shard->_indices.payload.seqValue;
+      for (size_t i = 0; i < indices.len; i++) {
+        auto &elem = indices.elements[i];
+        if (elem.valueType != SHType::Int)
+          throw ShaderComposeError("Take indices should be integers");
+        int index = elem.payload.intValue;
+        result += componentNames[index];
```
guusw, line 103: Should add index out of range check here

sinkingsugar (approved):
Can you add more comments to the examples, really need more instructions on how to use this

Kryptos-FR:
Why is it not targeting `devel`?

PR #439, opened 2022-10-19
Switch to rust stable

guusw:


sinkingsugar:
I think emscripten-mt will never work this PR is pointless 😭😭😭

guusw:
Yeah I'm not sure what to make of the rust standard library error in emscripten-mt. Probably just not tested outside of nightly

sinkingsugar:
I'd close this

PR #440, opened 2022-10-20
UI grid and UI table

Kryptos-FR:
**Description**
Adds `(UI.Grid)` and `(UI.Table)`.

**Details**
`(UI.Grid)` is a simple manual layout where widgets are set in a grid and a new row is created whenever `(UI.RowEnd)` is used.

`(UI.Table)` on the other end deals with sequence inputs and produces a row for each item in the sequence. Shards callback are given for each column of a row, and the columns' header and size are configurable.

### DIFF
--- a/docs/samples/shards/UI/Grid/1.edn
+++ b/docs/samples/shards/UI/Grid/1.edn
@@ -1,0 +1,32 @@
+(defloop main-wire
+  (GFX.MainWindow
+   :Contents
+   (->
+    (Setup
+     (GFX.DrawQueue) >= .ui-draw-queue
+     (GFX.UIPass .ui-draw-queue) >> .render-steps)
+    .ui-draw-queue (GFX.ClearQueue)
+
+    (UI
+     .ui-draw-queue
+     (UI.CentralPanel
+      (UI.Grid
+       :Striped true
+       :Contents
+       (->
+        "Col 1" (UI.Label)
+        "Col 2" (UI.Label)
+        "Col 3" (UI.Label)
+        (UI.NextRow)
+        "Hello" (UI.Label)
+        "shards" (UI.Label)
+        "!" (UI.Label)
+        (UI.NextRow)
+        "Lorem ipsum dolor sit amet" (UI.Label)
+        "consectetur adipiscing elit" (UI.Label)
+        "sed do eiusmod tempor" (UI.Label)))))
+
+    (GFX.Render :Steps .render-steps))))
+(defmesh root)
+(schedule root main-wire)
+(run root 0.1 10)
--- a/docs/samples/shards/UI/Table/1.edn
+++ b/docs/samples/shards/UI/Table/1.edn
@@ -1,0 +1,37 @@
+(defloop main-wire
+  (GFX.MainWindow
+   :Contents
+   (->
+    (Setup
+     (GFX.DrawQueue) >= .ui-draw-queue
+     (GFX.UIPass .ui-draw-queue) >> .render-steps)
+    .ui-draw-queue (GFX.ClearQueue)
+
+    (UI
+     .ui-draw-queue
+     (UI.CentralPanel
+      :Contents
+      (->
+       [{:Name "Doe" :Surname "John"}
+        {:Name "Dough" :Surname "Jane"}
+        {:Name "Smith" :Surname "Dick"}]
+       (UI.Table
+        :Resizable true
+        :Striped true
+        :RowIndex .index
+        :Columns
+        [{:Initial (float 20)}
+         {:Header "Surname"
+          :Initial (float 100) :AtLeast (float 60) :AtMost (float 160)}
+         {:Header "Name"
+          :Initial (float 80) :AtLeast (float 60) :AtMost (float 160)}]
+        :Rows
+        [(-> .index (ToString) (UI.Label))
+         (-> (ExpectTable) (Take "Surname") (ToString) (UI.Label))
+         (-> (ExpectTable)
+             (Take "Name") (ToString) (UI.Label))]))))
+
+    (GFX.Render :Steps .render-steps))))
+(defmesh root)
+(schedule root main-wire)
+(run root 0.1 10)
--- a/docs/samples/shards/UI/Table/1.edn
+++ b/docs/samples/shards/UI/Table/1.edn
@@ -27,9 +27,8 @@
           :Initial (float 80) :AtLeast (float 60) :AtMost (float 160)}]
         :Rows
         [(-> .index (ToString) (UI.Label))
-         (-> (ExpectTable) (Take "Surname") (ToString) (UI.Label))
-         (-> (ExpectTable)
-             (Take "Name") (ToString) (UI.Label))]))))
+         (-> (Take "Surname") (UI.Label))
+         (-> (Take "Name") (UI.Label))]))))
 
     (GFX.Render :Steps .render-steps))))
 (defmesh root)
--- a/docs/samples/shards/UI/Table/3.edn
+++ b/docs/samples/shards/UI/Table/3.edn
@@ -1,0 +1,36 @@
+(defloop main-wire
+  (GFX.MainWindow
+   :Contents
+   (->
+    (Setup
+     (GFX.DrawQueue) >= .ui-draw-queue
+     (GFX.UIPass .ui-draw-queue) >> .render-steps)
+    .ui-draw-queue (GFX.ClearQueue)
+
+    (UI
+     .ui-draw-queue
+     (UI.CentralPanel
+      :Contents
+      (->
+       [{:Name "Doe" :Surname "John"}
+        {:Name "Dough" :Surname "Jane"}
+        {:Name "Smith" :Surname "Dick"}]
+       (UI.Table
+        :Resizable true
+        :Striped true
+        :RowIndex .index
+        :Columns
+        [{:Initial (float 20)}
+         {:Header "Surname"
+          :Initial (float 100) :AtLeast (float 60) :AtMost (float 160)}
+         {:Header (-> "Name" (UI.Label) (UI.Button "Up" (Msg "Clicked Up")) (UI.Button "Down" (Msg "Clicked Down")))
+          :Initial (float 120) :AtLeast (float 100) :AtMost (float 160)}]
+        :Rows
+        [(-> .index (ToString) (UI.Label))
+         (-> (Take "Surname") (UI.Label))
+         (-> (Take "Name") (UI.Label))]))))
+
+    (GFX.Render :Steps .render-steps))))
+(defmesh root)
+(schedule root main-wire)
+(run root 0.1 10)
--- a/docs/samples/shards/UI/Table/3.edn
+++ b/docs/samples/shards/UI/Table/3.edn
@@ -23,7 +23,10 @@
         [{:Initial (float 20)}
          {:Header "Surname"
           :Initial (float 100) :AtLeast (float 60) :AtMost (float 160)}
-         {:Header (-> "Name" (UI.Label) (UI.Button "Up" (Msg "Clicked Up")) (UI.Button "Down" (Msg "Clicked Down")))
+         {:Header
+          (-> "Name" (UI.Label :Style {:text_style "Heading"})
+              (UI.Button "Up" (Msg "Clicked Up") :Style {:text_style "Small"})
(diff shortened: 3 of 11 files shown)


docs/samples/shards/UI/Grid/1.edn
```diff
@@ -0,0 +1,32 @@
...
+     .ui-draw-queue
+     (UI.CentralPanel
+      (UI.Grid
+       :Striped true
+       :Contents
+       (->
+        "Col 1" (UI.Label)
+        "Col 2" (UI.Label)
+        "Col 3" (UI.Label)
+        (UI.RowEnd)
```
guusw, line 20: I'd call it EndRow (or NextRow)

docs/samples/shards/UI/Table/1.edn
```diff
@@ -0,0 +1,37 @@
...
+      (->
+       [{:Name "Doe" :Surname "John"}
+        {:Name "Dough" :Surname "Jane"}
+        {:Name "Smith" :Surname "Dick"}]
+       (UI.Table
+        :Resizable true
+        :Striped true
+        :RowIndex .index
+        :Columns
+        [{:Initial (float 20)}
```
guusw, line 23: Maybe the Row callback can even be part of this table, keeps them tied closer together
like 
```clj
    {:Header "Surname" :Value (-> ...)}
```

docs/samples/shards/UI/Table/1.edn
```diff
@@ -0,0 +1,37 @@
...
+         {:Header "Surname"
```
guusw, line 24: Also maybe header should be a callback too, sometimes you might want different styling, maybe an image/icon.

rust/src/shards/gui/layouts/table.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+  }
+
+  fn hasCompose() -> bool {
+    true
+  }
+
+  fn compose(&mut self, data: &InstanceData) -> Result<Type, &str> {
+    for s in &mut self.shards {
+      s.compose(data)?;
```
guusw, line 191: I think you have to modify the input type here to match the sequence elements

docs/samples/shards/UI/Table/1.edn
Kryptos-FR: I was thinking about it but I find it easier to read when it's separate.

docs/samples/shards/UI/Table/1.edn
Kryptos-FR: Header callback sounds good. I will make it support either a string (keep current behavior) or a shard.

rust/src/shards/gui/layouts/table.rs
Kryptos-FR: I will try.

Kryptos-FR: I tried to modify the `inputType` before composing but I'm hit with an error:
> [trace] [2022-10-20 14:14:37.665] [T-42112] [runtime.cpp::2955] composeShards failed: basic_string: construction from null is not valid
[info] [2022-10-20 14:14:37.665] [T-42112] [runtime.cpp::3007] Compose failed with error basic_string: construction from null is not valid
[error] [2022-10-20 14:14:37.665] [T-42112] [runtime.cpp::1031] Error composing shard: Composition failed., wire: main-wire
[trace] [2022-10-20 14:14:37.665] [T-42112] [runtime.cpp::2955] composeShards failed: Composition failed.

```rust
fn compose(&mut self, data: &InstanceData) -> Result<Type, &str> {
  let mut data = *data;
  //...
  let input_type = data.inputType;
  let slice = unsafe {
    let ptr = input_type.details.seqTypes.elements;
    std::slice::from_raw_parts(ptr, input_type.details.seqTypes.len as usize)
  };
  let element_type = if !slice.is_empty() {
    slice[0]
  } else {
    common_type::any
  };

  for s in &mut self.shards {
    data.inputType = element_type;
    s.compose(&data)?;
  }

  // Always passthrough the input
  Ok(input_type)
}
```

Not sure why I get an error about a string, while I'm only manipulating types.

docs/samples/shards/UI/Grid/1.edn
Kryptos-FR: I will choose `NextRow` since this creates empty rows:
```clj
(UI.Grid
 :Striped true
 :Contents
 (->
  "Col 1" (UI.Label)
  "Col 2" (UI.Label)
  "Col 3" (UI.Label)
  (UI.NextRow)
  (UI.NextRow)
  (UI.NextRow)
  "Hello" (UI.Label)
  "shards" (UI.Label)
  "!" (UI.Label)))
```

docs/samples/shards/UI/Table/1.edn
Kryptos-FR: It's also easier to compose.

rust/src/shards/gui/layouts/table.rs
```diff
@@ -0,0 +1,421 @@
...
+    };
+
+    for s in &mut self.shards {
+      data.inputType = element_type;
+      s.compose(&data)?;
+    }
+
+    // Always passthrough the input
+    Ok(input_type)
+  }
```
guusw, line 267: Shouldn't the header_shards be composed too?

guusw (changes requested):

rust/src/shards/gui/layouts/table.rs
Kryptos-FR: They should. I just forgot 😉

Kryptos-FR:
@sinkingsugar please review

guusw (approved):

docs/samples/shards/UI/Table/1.edn
```diff
@@ -0,0 +1,36 @@
...
+     (UI.CentralPanel
+      :Contents
+      (->
+       [{:Name "Doe" :Surname "John"}
+        {:Name "Dough" :Surname "Jane"}
+        {:Name "Smith" :Surname "Dick"}]
+       (UI.Table
+        :Resizable true
+        :Striped true
+        :RowIndex .index
```
sinkingsugar, line 21: what's this for?

docs/samples/shards/UI/Table/1.edn
```diff
@@ -0,0 +1,36 @@
...
+     .ui-draw-queue
+     (UI.CentralPanel
+      :Contents
+      (->
+       [{:Name "Doe" :Surname "John"}
+        {:Name "Dough" :Surname "Jane"}
+        {:Name "Smith" :Surname "Dick"}]
+       (UI.Table
+        :Resizable true
+        :Striped true
```
sinkingsugar, line 20: can't this be abstracted based on data?

docs/samples/shards/UI/Table/3.edn
```diff
@@ -0,0 +1,39 @@
...
+        :RowIndex .index
+        :Columns
+        [{:Initial (float 20)}
+         {:Header "Surname"
+          :Initial (float 100) :AtLeast (float 60) :AtMost (float 160)}
+         {:Header
+          (-> "Name" (UI.Label :Style {:text_style "Heading"})
+              (UI.Button "Up" (Msg "Clicked Up") :Style {:text_style "Small"})
+              (UI.Button "Down" (Msg "Clicked Down") :Style {:text_style "Small"}))
+          :Initial (float 120) :AtLeast (float 100) :AtMost (float 160)}]
```
sinkingsugar, line 30: a bit complex, are defaults decent?

rust/src/shards/gui/layouts/grid.rs
```diff
@@ -0,0 +1,318 @@
...
+    Ok(())
+  }
+
+  fn activate(&mut self, context: &Context, input: &Var) -> Result<Var, &str> {
+    if self.contents.is_empty() {
+      return Ok(*input);
+    }
+
+    if let Some(ui) = util::get_current_parent(*self.parents.get())? {
+      let mut grid = egui::Grid::new(EguiId::new(self, 0));
```
sinkingsugar, line 211: all those `new` in egui.. can't they by easily cached inside Shards memory? instead of wasting stack or hell even allocations potentially

rust/src/shards/gui/layouts/table.rs
```diff
@@ -0,0 +1,429 @@
...
+      .into(),
+    (
+      cstr!("Columns"),
+      cstr!("Configuration of the columns."),
+      &SEQ_OF_ANY_TABLE_TYPES[..]
+    )
+      .into(),
+    (
+      cstr!("Striped"),
+      cstr!("Whether to alternate a subtle background color to every other row."),
```
sinkingsugar, line 52: ah I see

rust/src/shards/gui/layouts/table.rs
```diff
@@ -0,0 +1,429 @@
...
+      .into(),
+    (
+      cstr!("Resizable"),
+      cstr!("Whether columns can be resized within their specified range."),
+      BOOL_VAR_OR_NONE_SLICE,
+    )
+      .into(),
+    (
+      cstr!("RowIndex"),
+      cstr!("Variable to hold the row index, to be used within Rows."),
```
sinkingsugar, line 64: seems added complexity, can it be avoided somehow?

rust/src/shards/gui/layouts/table.rs
```diff
@@ -0,0 +1,429 @@
...
+      if let Some(header_shard) = header_shard {
+        header_shard.compose(data)?;
+      }
+    }
+
+    // compose row shards
+    let mut data = *data;
+    let mut shared: ExposedTypes;
+
+    // we need to inject the row index to the inner shards
```
sinkingsugar, line 218: if you really really really need a variable you could inject one like I think `Map` or `ForEach` does which is a special character, to check in `core.cpp`

rust/src/shards/gui/layouts/table.rs
```diff
@@ -0,0 +1,429 @@
...
+  fn activate(&mut self, context: &Context, input: &Var) -> Result<Var, &str> {
+    if let Some(ui) = util::get_current_parent(*self.parents.get())? {
+      ui.push_id(EguiId::new(self, 0), |ui| {
+        use egui_extras::{Size, TableBuilder};
+
+        let seq = Seq::try_from(input)?;
+        let text_height = egui::TextStyle::Body.resolve(ui.style()).size;
+        // start building a table
+        let mut builder =
+          TableBuilder::new(ui).cell_layout(egui::Layout::left_to_right(egui::Align::Center));
```
sinkingsugar, line 330: applies to all egui but within Shards all those can be easily cached

sinkingsugar (changes requested):
looks good but I had some remark

docs/samples/shards/UI/Table/1.edn
Kryptos-FR: What can be abstracted?

rust/src/shards/gui/layouts/table.rs
Kryptos-FR: It's optional. But there is no other way to tell a row calback in which index it is when it is called.

rust/src/shards/gui/layouts/table.rs
Kryptos-FR: The way egui works, most of the time they can't be cached: they use the "builder pattern" which means they return a moved `Self`, not a reference. So after being cached, nothing can be changed again (some calls can't be made anymore because of moving `Self`).

rust/src/shards/gui/layouts/table.rs
Kryptos-FR: I don't see anything there that has the same functionality.

rust/src/shards/gui/layouts/grid.rs
Kryptos-FR: See comment below.

docs/samples/shards/UI/Table/3.edn
Kryptos-FR: By default it is set to `egui_extras::Size::Remainder` which basically splits equally the columns. This can be seen in practice in the second sample `docs/samples/shards/UI/Table/2.edn`.

sinkingsugar:
Can we have some memory leaks checking in tests? I wonder, ASAN on linux maybe? cc @guusw , not sure if rust supports such cases. Asking cos everytime I implement something table related there are many leaks, also rust is not leak immune due to a lot of FFI.

PR #441, opened 2022-10-21
add Pure and defpure

sinkingsugar:
At long last

### DIFF
--- a/src/core/runtime.cpp
+++ b/src/core/runtime.cpp
@@ -550,8 +550,9 @@ SHVar *referenceVariable(SHContext *ctx, const char *name) {
     auto rit = ctx->wireStack.rbegin();
     for (; rit != ctx->wireStack.rend(); ++rit) {
       // prioritize local variables
-      auto it = (*rit)->variables.find(name);
-      if (it != (*rit)->variables.end()) {
+      auto wire = *rit;
+      auto it = wire->variables.find(name);
+      if (it != wire->variables.end()) {
         // found, lets get out here
         SHVar &cv = it->second;
         cv.refcount++;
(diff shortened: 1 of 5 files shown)


Kryptos-FR:
I tried it to replace some `defshards` in the snake tutorials. I was able to do it for `get-index` and `populate-grid`. However, I need to specify some type expectations during composition.

```diff
// snake-logic.edn
-(defshards get-index []
+(defpure get-index
  (| (Take 0) >= .x)
  (| (Take 1) >= .y)
  .y (Math.Multiply grid-cols) (Math.Add .x))

-(defshards populate-grid [fruit snake]
-  >= .tmp-grid
+(defpure populate-grid
+  (| (Take 0) (ExpectInt2) = .fruit)
+  (| (Take 1) (ExpectInt2Seq) = .snake)
+  (| (Take 2) (ExpectIntSeq) >= .grid)

```

So I think we should add a few more ExpectXXX in this PR as they would be needed:

```diff
// casting.cpp
  using ExpectFloatSeq = ExpectXComplex<CoreInfo::FloatSeqType>;
  REGISTER_SHARD("ExpectFloatSeq", ExpectFloatSeq);
+  using ExpectFloat2Seq = ExpectXComplex<CoreInfo::Float2SeqType>;
+  REGISTER_SHARD("ExpectFloat2Seq", ExpectFloat2Seq);
+  using ExpectFloat3Seq = ExpectXComplex<CoreInfo::Float3SeqType>;
+  REGISTER_SHARD("ExpectFloat3Seq", ExpectFloat3Seq);
+  using ExpectFloat4Seq = ExpectXComplex<CoreInfo::Float4SeqType>;
+  REGISTER_SHARD("ExpectFloat4Seq", ExpectFloat4Seq);
  using ExpectIntSeq = ExpectXComplex<CoreInfo::IntSeqType>;
  REGISTER_SHARD("ExpectIntSeq", ExpectIntSeq);
+  using ExpectInt2Seq = ExpectXComplex<CoreInfo::Int2SeqType>;
+  REGISTER_SHARD("ExpectInt2Seq", ExpectInt2Seq);
+  using ExpectInt3Seq = ExpectXComplex<CoreInfo::Int3SeqType>;
+  REGISTER_SHARD("ExpectInt3Seq", ExpectInt3Seq);
+  using ExpectInt4Seq = ExpectXComplex<CoreInfo::Int4SeqType>;
+  REGISTER_SHARD("ExpectInt4Seq", ExpectInt4Seq);
+  using ExpectInt8Seq = ExpectXComplex<CoreInfo::Int8SeqType>;
+  REGISTER_SHARD("ExpectInt8Seq", ExpectInt8Seq);
+  using ExpectInt16Seq = ExpectXComplex<CoreInfo::Int16SeqType>;
+  REGISTER_SHARD("ExpectInt16Seq", ExpectInt16Seq);
```

src/core/runtime.cpp
```diff
@@ -1016,10 +1024,12 @@ void validateConnection(ValidationContext &ctx) {
         shards::arrayPush(data.shared, type);
       }
     }
-    // and inherited
-    for (auto &info : ctx.inherited) {
-      for (auto &type : info.second) {
-        shards::arrayPush(data.shared, type);
+    if (!ctx.wire->pure) {
```
guusw, line 1027: Chan you change this to `if (!ctx.wire || !ctx.wire->pure) {` so data.wire can be null when composing?

I have a case like this in the shader translator where I take a vector of shards and have to compose them but they don't belong to a wire so it's set to null in SHInstanceData (also why it crashes on CI)

guusw (changes requested):

sinkingsugar:
> `(defpure)` is (roughly) equivalent to `(defwire :Pure)`. Do we need another macro for `(defloop :Pure)`?
> 
> In other words, is there a way to adapt the existing `(defloop)` macro to have an optional `:Pure` parameter?

Actually I think adding `:Pure` to `defwire` and `defloop` works already indeed.

sinkingsugar:
> > `(defpure)` is (roughly) equivalent to `(defwire :Pure)`. Do we need another macro for `(defloop :Pure)`?
> > In other words, is there a way to adapt the existing `(defloop)` macro to have an optional `:Pure` parameter?
> 
> Actually I think adding `:Pure` to `defwire` and `defloop` works already indeed.

This does not work due to the way the macro works.

Let's just leave `defpure` for now as it will be the most common.
can always write `(def my-wire (Wire "my-wire" :Looped :Pure :LStack ...))` if needed

Kryptos-FR:
I did not finish reviewing it. I'd prefer we don't merge during the weekend.

PR #442, opened 2022-10-25
UI plot

Kryptos-FR:
Plot widgets using `egui` instead of `ImGui`.

Similarly to what we had with the `ImGui` integration, it supports lines, points and bars. There are more features we can integrate later. However, the goal of this PR is to have a par implementation so that `ImGui` can be removed.

### DIFF
--- a/rust/src/shards/gui/widgets/plots/plot_bar.rs
+++ b/rust/src/shards/gui/widgets/plots/plot_bar.rs
@@ -1,0 +1,201 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright © 2022 Fragcolor Pte. Ltd. */
+
+use super::PlotBar;
+use super::EGUI_PLOT_UI_TYPE;
+use super::PLOT_UI_NAME;
+use super::SEQ_OF_FLOAT2_TYPES;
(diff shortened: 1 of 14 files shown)


rust/src/shards/gui/widgets/plots/plot_bar.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    Ok(())
+  }
+
+  fn activate(&mut self, _context: &Context, input: &Var) -> Result<Var, &str> {
+    let plot_ui: &mut egui::plot::PlotUi =
+      Var::from_object_ptr_mut_ref(*self.plot_ui.get(), &EGUI_PLOT_UI_TYPE)?;
+
+    let seq: Seq = input.try_into()?;
+    let points = seq.iter().map(|x| {
```
sinkingsugar, line 153: hell a lot of allocations here, compared to imgui plot where I avoided allocations completely :(

Kryptos-FR: Can't be totally avoided, as `egui::plot::BarChart::new()` takes a `Vec<egui::plot::Bar>` as parameter.

PR #447, opened 2022-10-28
Codecov

Kryptos-FR:
Improve CI jobs for code coverage.

- merge `Linux-test` and `Linux-test-extra` jobs to save one machine during build and a bit of time
- upload coverage data from `Linux-test` and `Linux-gpu` as artifacts
- add a separate job to collect all coverage artifacts and send them at once to codecov
- uses codecov official action instead of the bash script

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -207,13 +207,17 @@ jobs:
         if: ${{ needs.Linux.outputs.build-type == 'Debug' }}
         run: |
           mkdir coverage
+          # capture
           lcov \
             --capture \
             --directory build/src \
             --output-file coverage/coverage.info
+          # remove external dependencies
           lcov \
             --remove coverage/coverage.info "*/c++/*" "*/boost/*" "*/usr/*" "*/deps/*" "*/src/mal/*" \
             --output-file coverage/coverage.linux.info
+          # convert absolute path to relative path
+          sed -i s/${PWD////\\/}/./g coverage/coverage.linux.info
       - name: Upload coverage (Debug)
         if: ${{ needs.Linux.outputs.build-type == 'Debug' }}
         uses: actions/upload-artifact@v3
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -158,7 +158,7 @@ jobs:
           ./shards ../src/tests/genetic.clj
           ./shards ../src/tests/imaging.clj
           ./shards ../src/tests/http.clj
-          ./shards ../src/tests/ws.edn
+          # ./shards ../src/tests/ws.edn
           ./shards ../src/tests/bigint.clj
           ./shards ../src/tests/brotli.clj
           ./shards ../src/tests/snappy.clj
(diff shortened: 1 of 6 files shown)


.github/workflows/codecov.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Download Linux coverage
+        uses: actions/download-artifact@v3
+        with:
+          name: shards-linux-coverage
+          path: coverage
+      - name: Download Linux GPU coverage
+        uses: actions/download-artifact@v3
+        with:
+          name: shards-linux-gpu-coverage
+          path: coverage
```
guusw, line 42: Collecting the coverage in one step is one way to do it, nice solution 😄

.github/workflows/test-linux-gpu.yml
```diff
@@ -121,19 +121,29 @@ jobs:
...
         if: ${{ steps.setup.outputs.build-type == 'Debug' }}
         uses: actions/upload-artifact@v3
         with:
           name: shards-linux-gpu-coverage
           path: |
-            coverage
-            test-gfx.xml
+            coverage/coverage.linux-gpu.info
           if-no-files-found: error
+      # - name: Test results (Debug)
```
guusw, line 139: Remove comment

.github/workflows/build-linux.yml
```diff
@@ -120,31 +119,43 @@ jobs:
...
+
+          cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCODE_COVERAGE=1 -DSHARDS_BUILD_TESTS=1 ..
+
+          if [ "${{ needs.Linux.outputs.run-tests }}" == "true" ]
+          then
+            ninja shards
+          fi
+          if [ "${{ needs.Linux.outputs.run-extra-tests }}" == "true" ]
+          then
+            ninja test-runtime
```
guusw, line 157: I would even just remove `run-extra-tests` and just run this under `run-tests`, no point in disabling only test-runtime. It's technically not even testing any of the 'extra' shards.

guusw (changes requested):

.github/workflows/build-linux.yml
Kryptos-FR: Ok. It was separate before so I thought I had to keep it that way. But yeah let's test everything.

.github/workflows/test-linux-gpu.yml
Kryptos-FR: I kept it there on purpose. It was uploading a file `test-gfx.xml` but I didn't want to mess up the coverage artifacts so I disabled it.

If you say we don't need it anymore, I can indeed remove this commented part.

guusw: Ah yeah if it's just the xml, you can add it to the `if: failure()` step below, it's more useful there anyway.

In the case a graphics comparison tests fails it'll upload the images there and then the xml can be there too containing a list of which tests failed/passed

sinkingsugar (approved):

PR #457, opened 2022-11-14
Vector constructors & more generic math implementations

guusw:
Some changes I needed while working on the shader translator.

This make Float2/3/4 accept variables, e.g.:
```clojure
(Float4 0.0 0.0 .z .w) >= .my-float-4
(Float4 .all-4-values) >= .my-float-4-2
```

Some internal changes to make more combinations of math operations possible and remove duplication of applying operations between sequences

Performance note: everything is templated and the result is the same as when writing out the switch statements as done previously.

### DIFF
--- a/include/number_types.hpp
+++ b/include/number_types.hpp
@@ -26,6 +26,8 @@ enum class NumberType {
 std::map<SHType, NumberType> getSHTypeToNumberTypeMap();
 const std::vector<NumberType> &getSHTypeToNumberTypeArrayMap();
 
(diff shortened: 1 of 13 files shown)


sinkingsugar (approved):
Very nice, approved with the condition that we can have very high coverage as there are a lot of changes!

guusw:
@sinkingsugar I'll go over the coverage and add cases where it's missing

Kryptos-FR (approved):

PR #464, opened 2022-11-21
Shader generator improvements

guusw:
Various work on shader translator.

Composition now happens whenever the renderer requests a shader variant, this allows composition based on actual parameter, input/output types and available fields.
It also removes the need to explicitly specify the expected type and dimension on some shader shards.

This PR also adds a bunch of shader features (If, ForEach, Comparisons) and some fixes for exisiting shader shards.

NOTE: some shards have been moved to header files with public fields so they can be read by the shader translator (flow.hpp, wires.hpp)


guusw:
Adding some points since it involves some additional work besides just the issue that was linked

src/core/shards/wires.cpp
```diff
@@ -11,254 +10,197 @@
 #endif
 
 namespace shards {
-enum RunWireMode { Inline, Detached, Stepped };
 
-struct WireBase {
```
guusw, line 16: Moved into header

src/core/shards/flow.cpp
```diff
@@ -1,1134 +1,9 @@
...
 /* Copyright © 2019 Fragcolor Pte. Ltd. */
 
-#include "shardwrapper.hpp"
-#include "shards.h"
-#include "foundation.hpp"
-#include "shared.hpp"
+#include "flow.hpp"
 
 namespace shards {
-static Type condShardSeqs = Type::SeqOf(CoreInfo::ShardsOrNone);
```
guusw, line 10: Moved into header

guusw:
Have to squash some WIP commits but otherwise ready

sinkingsugar (approved):
I think we need a lot of tests,
I could not see if there were major changes in flow or wires, I hope not.

guusw:
@sinkingsugar Not changes in flow/wire shards besides making some fields public.

Kryptos-FR:
Why is it not targeting `devel`?

guusw:
I will make a new branch with doc additions (shard help, examples, small explanations, etc.)

PR #465, opened 2022-11-21
Fix OpenSSL selection

guusw:
Changes:
- Use system OpenSSL on system that have it (linux-based)

This fixes OpenSSL not finding the root CA certificates on linux


Kryptos-FR (approved):
Checked and worked for me both on Linux Ubuntu 22.04 (yesterday) and now Ubuntu 22.10 (since upgrade).

PR #466, opened 2022-11-22
Test (HTTP) test failure

guusw:


.github/workflows/build-linux.yml
```diff
@@ -58,87 +58,37 @@ jobs:
...
           ninja shards
           ninja test-runtime
+      - name: Build (Release)
+        if: ${{ steps.setup.outputs.build-type == 'Release' }}
+        run: |
+          mkdir -p build
+          cd build
+
+          cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..
+          ninja shards
```
Kryptos-FR, line 91: Would `test-runtime` work in release? If so maybe we could also build and run it.

cmake/Utils.cmake
```diff
@@ -6,7 +6,7 @@ if(NOT CMAKE_PROPERTIES_TO_DUPLICATE)
   string(REGEX REPLACE ";" "\\\\;" CMAKE_PROPERTIES_TO_DUPLICATE "${CMAKE_PROPERTIES_TO_DUPLICATE}")
   string(REGEX REPLACE "\n" ";" CMAKE_PROPERTIES_TO_DUPLICATE "${CMAKE_PROPERTIES_TO_DUPLICATE}")
 
-  set(IGNORED_PROPERTIES "IMPORTED_GLOBAL" "NAME" "TYPE" "SOURCES" "HEADER_SETS" "INTERFACE_HEADER_SETS")
+  set(IGNORED_PROPERTIES "IMPORTED_GLOBAL" "NAME" "TYPE" "SOURCES" "HEADER_SETS" "INTERFACE_HEADER_SETS" "CXX_MODULE_HEADER_UNIT_SETS")
```
Kryptos-FR, line 9: I also had the issue with `CXX_MODULE_SETS`, `INTERFACE_CXX_MODULE_HEADER_UNIT_SETS` and `INTERFACE_CXX_MODULE_SETS`.

guusw: I'll add those

PR #472, opened 2022-12-01
GFX & Shader Documentation

guusw:
## [Preview](https://tdrz.nl/shards_docs/)

Document GFX shards:
- [x] BuiltinFeature
- [x] Draw
- [x] Drawable
- [x] DrawablePass
- [x] DrawQueue
- [x] Feature
- [x] glTF
- [x] Render
- [x] UIPass

PR #473, opened 2022-12-03
Move enum registration / conditional shard Inlining

guusw:
Couldn't help it, I spent some extra time on improving build iteration times & cleaning up some of the enum registration.

It ties in a bit to my #472 branch too because I can move the enum help strings the source file too.


guusw:
Some statistics before/after

## Full shards source

With

```bash
rm -rf src/core src/extra
time cmake --build . --target shards
```

### devel (Debug/Windows)

```
cmake --build . --target shards  0.00s user 0.01s system 0% cpu 1:46.59 total
```

### this branch (Debug/Windows)

```
cmake --build . --target shards  0.00s user 0.01s system 0% cpu 52.644 total
cmake --build . --target shards  0.01s user 0.00s system 0% cpu 52.190 total
```

## Partial rebuild (gfx/gizmos)

With

```bash
pushd src/extra/CMakeFiles/shards-extra.dir; rm -rf gfx gizmos; popd
time cmake --build . --target shards
```

### devel (Debug/Windows)

```
cmake --build . --target shards  0.00s user 0.00s system 0% cpu 46.512 total
cmake --build . --target shards  0.00s user 0.00s system 0% cpu 46.027 total
```

### this branch (Debug/Windows)

```
cmake --build . --target shards  0.00s user 0.00s system 0% cpu 20.348 total
```

src/core/shards/inlined.cpp
```diff
@@ -0,0 +1,154 @@
...
+    shard->inlineShardId = InlineShard::MathMod;
+  } else if (name == "Math.LShift") {
+    shard->inlineShardId = InlineShard::MathLShift;
+  } else if (name == "Math.RShift") {
+    shard->inlineShardId = InlineShard::MathRShift;
+  }
+  // Unary math is dealt inside math.hpp compose
+}
+
+SHARDS_CONDITIONAL_INLINE FLATTEN bool activateShardInline(Shard *blk, SHContext *context, const SHVar &input, SHVar &output) {
```
sinkingsugar, line 83: well this won't work, cos it's in a .cpp file

sinkingsugar:
Well first of all I don't think this was a priority at all, second while nice on paper, I don't think shards will be inlined, C++ works in a very rudimentary way, inline functions have to be headers, anything .cpp gets it's own compilation unit. LTO might work but that's another bet.
Check assembly, I'm pretty sure this doesn't work.

Also you could have split the work.. enum stuff easy to approve, core runtime nope.

sinkingsugar:
Maybe you should revert just the inline shards bit, that require valgrind measuring (that's how I implemented them...) before any refactor. 👍

src/core/shards/inlined.cpp
guusw: So when it's enabled, it'll compile [this file](https://github.com/fragcolor-xyz/shards/blob/478a33940040c9d06dfc3febc49e839c319ee984/src/core/runtime_inlined.cpp)
which groups everything into a single compilation unit, also checked the assembly to make sure it still does.

sinkingsugar: ah ok good job

sinkingsugar:
Alright let's review it as is since you checked asm and get it out of the way then 👍
Btw you are already in SG timezone I can see @guusw :D

sinkingsugar:
Let me know when ready to review, add me 🙏

guusw:
Yeah should be good, ran some test script as well
```clojure
(defpure inner
  = .i
  .i (Math.Add 2))

(defwire test
  0 >= .v
  [] >= .seq
  (ForRange 0 10000000 (->
                        >= .av
                        .v (Math.Add .av) > .v
                        .v (Do inner) > .v
                        .v >> .seq))
  .v)

(defmesh root)
(schedule root test)
(if (run root) nil (throw "Root tick failed"))
```

Same timings comparing to devel, there's not that inline shards and this should at least touch some to indicate that they're still inlined

sinkingsugar (approved):
Looks good 👍

PR #474, opened 2022-12-04
Propagate Do failures

guusw:
Changed Do behaviour to propagate failures to the parent wire

Due to this I uncovered a failure in an IPFS upload in flows.edn, moved it out of there for now.


sinkingsugar (approved):
Correct fix 👍

guusw:
Test fail was http against ipfs again, rest seems to pass

PR #476, opened 2022-12-06
Documentation Improvements

guusw:
## [Preview](https://tdrz.nl/shards_docs/)

As requested, split #472 into just the structure/technical changes, which are included here.

- [x] Improved search (terms like ">=", ">>", "UI Context", "UI.Context", "Context" all work)
- [x] Compact shard IO/Parameter overview
- [x] Generate enum documentation and per-value help text
- [x] Add section about how to read the shard documentation (docs\reference\shards)
- [x] Type section cleanup to use the new type format
- [x] Type section cleanup to use the term "Sequence" instead of "Array"
- [x] Type section cleanup to use the term "Shard" and "[ Shard ]" instead of ShardRef
- [x] Move shards to "Reference" section
- [x] Move "shards introduction" to the "Learn" section
- [x] Add image outputs to samples, e.g. in [GFX.Render](http://tdrz.nl/shards_docs/reference/shards/GFX/Render/)


guusw:
@Kryptos-FR To address the comments from the other PR.

I moved the pages to more clearly separate where the reference is and which pages go into depth, I think it's good to have them separate. Even if it's a work in progress, it should be fine to move I believe.

For the parameter tables, the only HTML I added is to add a `<div class="sh-parameters" markdown="1">` outer class for css, I looked but there didn't seem to be a way to add a class to a markdown table directly ([Python-Markdown#312](https://github.com/Python-Markdown/markdown/issues/312))

Heres the table without being able to apply custom css to it, I think it's necessary to make it look reasonable
TableDiff

sinkingsugar (approved):
So far so good, I think it was needed to have enum descriptions.
CI failing though.

guusw:
One thing to note is that I didn't implement attaching descriptions for rust enums, I think we can parse annotations in the decl macros but for now I left them empty from rust

docs/docs/reference/shards/UI/index.md
```diff
@@ -0,0 +1,9 @@
+# UI
```
Kryptos-FR, line 1: That file shouldn't exist since we generate it for `(UI)` aka the context.

guusw: I don't think we should have it as a top level page if UI is a shard itself and has children.
I let it generate in the default path (General/UI.md) and linked to there since that avoids having to manually copy it

guusw:
I see you commented on the other PR for things that are now split into this PR, but I'll comment over there (#472)

src/core/ops_internal.cpp
```diff
@@ -16,22 +17,27 @@ std::ostream &operator<<(std::ostream &os, const SHVar &var) {
...
+      os << "<Wire: None>";
     }
   } break;
   case SHType::Bytes:
-    os << "Bytes: 0x" << std::hex << reinterpret_cast<uintptr_t>(var.payload.bytesValue) << " size: " << std::dec
-       << var.payload.bytesSize;
+    os << "<" << var.payload.bytesSize << " Bytes>" << std::dec;
     break;
   case SHType::Array:
     os << "Array: 0x" << std::hex << reinterpret_cast<uintptr_t>(var.payload.arrayValue.elements) << " size: " << std::dec
```
guusw, line 43: Format enums

include/utility.hpp
```diff
@@ -310,21 +316,8 @@ struct TEnumInfo {
...
-
-  static constexpr auto enum_values() {
-    if constexpr (IsFlags) {
-      return magic_enum::flags::enum_values<E>();
-    } else {
-      return magic_enum::enum_values<E>();
-    }
-  }
+  static constexpr auto enum_names() { return magic_enum::enum_names<E>(); }
+  static constexpr auto enum_values() { return magic_enum::enum_values<E>(); }
```
Kryptos-FR, line 320: Are you sure about that?
I remember I had issues with it hence why this code was changed.

Kryptos-FR: It was added for enum flags with ImGui which we don't have anymore. So we will see once we have a need for new ones.

Kryptos-FR (approved):

PR #477, opened 2022-12-06
Add Math.Lerp

guusw:
- [x] Add `Math.Lerp` & tests
- [x] Add shader translation

Usage:

```clojure
  (Float3 0.0 1.0 2.0) >= .first-3
  (Float3 4.0 -1.0 0.0) >= .second-3
  0.5 (Math.Lerp .first-3 .second-3) (Log "Lerped value")
  
  (Color 0x10 0x20 0x40 0xff) >= .first-c
  (Color 0x25 0xff 0x20 0xff) >= .second-c
  0.5 (Math.Lerp .first-c .second-c) (Log "Lerped color")
```


Kryptos-FR (approved):
LGTM.

Might as well add `Math.InvLerp`  (or `Math.InverseLerp` if we want to be more explicit).

guusw:
What would that do? @Kryptos-FR

PR #479, opened 2022-12-07
Shader stuff & related fixes

guusw:
- Fix :Feature parameter on Drawable/glTF 
- Add shader bitwise operators (L/RShift, And,Xor,Or)
- Fix validation for ShardsVar inside pure & add test cases


src/core/runtime.cpp
```diff
@@ -978,12 +978,10 @@ void validateConnection(ValidationContext &ctx) {
         shards::arrayPush(data.shared, type);
       }
     }
-    if (!ctx.wire || !ctx.wire->pure) {
-      // and inherited
-      for (auto &info : ctx.inherited) {
-        for (auto &type : info.second) {
-          shards::arrayPush(data.shared, type);
-        }
+    // and inherited
```
sinkingsugar, line 981: this likely is no good? are we sure @guusw ?

src/core/shards/wires.cpp
```diff
@@ -133,13 +133,15 @@ SHTypeInfo WireBase::compose(const SHInstanceData &data) {
...
   IterableExposedInfo shared(data.shared);
   IterableExposedInfo sharedCopy;
-  if (mode == RunWireMode::Detached && !capturing) {
-    // keep only globals
-    auto end = std::remove_if(shared.begin(), shared.end(), [](const SHExposedTypeInfo &x) { return !x.global; });
-    sharedCopy = IterableExposedInfo(shared.begin(), end);
-  } else {
-    // we allow Detached but they need to be referenced during warmup
-    sharedCopy = shared;
+  if (!wire->pure) {
```
sinkingsugar, line 136: rationale?

sinkingsugar (changes requested):
Can you explain a bit more? I have a bad feeling about the changes but I saw you added more tests.... the thing is that even if pure works we risk to break all other things

sinkingsugar (approved):

src/core/runtime.cpp
guusw: Will add comment to clarify

src/tests/pure.edn
```diff
@@ -1,15 +1,45 @@
...
+    4 >= .value
+    (Do pure))
+  (prepare main))"
+  (EDN.Eval))
 
-(schedule root main)
-(if (run root 0.1) nil (throw "Root tick failed"))
\ No newline at end of file
+(schedule root failure-case)
+(if (run root 0.1) (throw "Expected failure") nil)
```
Kryptos-FR, line 45: Smart

PR #480, opened 2022-12-08
UI documentation

Kryptos-FR:
Add missing text in `help()` functions for UI shards.


Kryptos-FR:
@ChelseaChX I added you as a reviewer. I suggest building shards and the doc and reviewing the final documents, if that's easier for you than browsing the source code.

Kryptos-FR:
@guusw I am making some changes to simplify UI samples. Basically I added a `--looped` param to `run-sample.edn` to be used with `UI` and `GFX` samples.

I noticed that the single GFX sample was using (run root (/ 1.0 120.0) 200)`. Is it important to run for slightly more than a second and at 120 FPS for that test? If not, I will also simplify it and it will use the default timestep of UI samples which is `(run main 0.02 50)` (i.e. 1 second at 50 FPS).

rust/src/shards/gui/menus/menu.rs
```diff
@@ -1,399 +1,413 @@
...
+  }
+
+  fn name(&mut self) -> &str {
+    "UI.Menu"
+  }
+
+  fn help(&mut self) -> OptionalString {
+    OptionalString(shccstr!(
+      "Creates a menu button that when clicked will show the given menu. \r\n
+      If called from within a menu this will instead create a button for a sub-menu."
```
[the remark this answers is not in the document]
Kryptos-FR: oh I think it interprets the leading space as meaning it's code. That's an issue to fix.

rust/src/shards/gui/menus/menu.rs
```diff
@@ -1,399 +1,413 @@
...
+  {
+    compile_time_crc32::crc32!("UI.CloseMenu-rust-0x20200101")
+  }
+
+  fn name(&mut self) -> &str {
+    "UI.CloseMenu"
+  }
+
+  fn help(&mut self) -> OptionalString {
+    OptionalString(shccstr!("Closes the currently opened menu."))
```
ChelseaChX, line 82: Untitled

The text is not showing for some reason.

Kryptos-FR: It's showing for me. Did you rebuild `shards` in debug before generating the doc?

ChelseaChX: Ah yes that was the problem. I see it now, thanks!

Kryptos-FR:
Thanks for the feedback. I will improve based upon it.

>     * When referring to parameters, try to keep the capitalization the same, keep the colon before it, and place it within a code block. e.g. the `:Contents`'s shards VS the contents shards. Mostly seen in the  description (e.g. UI.TopPanel).

As discussed, we will not include the `:` as it is a feature of textual shard.

PR #498, opened 2022-12-16
Shards flow docs

ChelseaChX:
- Reworked the Do/Dispatch example to better reflect the differences between the two
- Removed incorrect explanations of defn and Wire usage
- Added the segment on Step
- Added a summary table


docs/docs/learn/shards/.pages
```diff
@@ -1,6 +1,6 @@
 nav:
   - index.md
-  - introduction.md
   - what-is-shards.md
   - the-flow.md
   - working-with-data.md
+
```
Kryptos-FR, line 6: new line should be empty (here it has a space which make it not being considered a new line)

docs/docs/learn/shards/.pages
```diff
@@ -1,6 +1,6 @@
 nav:
   - index.md
-  - introduction.md
```
Kryptos-FR, line 3: File still exists. Should it be removed as well?

Kryptos-FR (approved):
LGTM.

PR #501, opened 2022-12-16
Tracy profiler suport

guusw:
Worked on this a while back but never finished it, let's merge it.

The way you use it:
- Grab the profiler UI from https://github.com/wolfpld/tracy/releases (Use 0.9 for the currently library version)
- Build shards with TRACY_ENABLE=ON (-D on CMake configure or edit cache)
- Use the tracy API from c++ (`#include <tracy/Tracy.hpp>`)

It works from rust too, though for now only from the `gfx` crate, because it needs the initialization to be done from rust. After initialization C++ & rust share the state though.

### DIFF
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -221,3 +221,11 @@ target_include_directories(utf8.h INTERFACE utf8.h)
 
 add_library(pdqsort INTERFACE)
 target_include_directories(pdqsort INTERFACE pdqsort)
+
+# Tracy profiler, default off
+option(TRACY_ENABLE "Enable profiling" OFF)
+
+# These settings are required to make tracy from C++ work together with rust's tracy-client
+set(TRACY_MANUAL_LIFETIME ON CACHE BOOL "" FORCE)
+set(TRACY_DELAYED_INIT ON CACHE BOOL "" FORCE)
+add_subdirectory(tracy)
--- a/src/gfx/context.cpp
+++ b/src/gfx/context.cpp
@@ -5,6 +5,7 @@
 #include "platform_surface.hpp"
 #include "window.hpp"
 #include "log.hpp"
+#include <tracy/Tracy.hpp>
 #include <magic_enum.hpp>
 #include <spdlog/fmt/fmt.h>
 #include <spdlog/spdlog.h>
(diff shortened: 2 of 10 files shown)


src/gfx/context.cpp
```diff
@@ -5,6 +5,7 @@
 #include "platform_surface.hpp"
 #include "window.hpp"
 #include "log.hpp"
+#include <tracy/Tracy.hpp>
```
Kryptos-FR, line 8: Also guard it with a `#ifdef `

deps/CMakeLists.txt
```diff
@@ -221,3 +221,11 @@ target_include_directories(utf8.h INTERFACE utf8.h)
...
 add_library(pdqsort INTERFACE)
 target_include_directories(pdqsort INTERFACE pdqsort)
+
+# Tracy profiler, default off
+option(TRACY_ENABLE "Enable profiling" OFF)
+
+# These settings are required to make tracy from C++ work together with rust's tracy-client
+set(TRACY_MANUAL_LIFETIME ON CACHE BOOL "" FORCE)
+set(TRACY_DELAYED_INIT ON CACHE BOOL "" FORCE)
+add_subdirectory(tracy)
```
Kryptos-FR, line 231: Nit: missing empty line

src/gfx/renderer.cpp
```diff
@@ -24,6 +24,7 @@
 #include "render_target.hpp"
 #include "log.hpp"
 #include "geom.hpp"
+#include <tracy/Tracy.hpp>
```
Kryptos-FR, line 27: Also guard it with a `#ifdef `

src/gfx/texture.cpp
```diff
@@ -1,6 +1,7 @@
 #include "texture.hpp"
 #include "context.hpp"
 #include "error_utils.hpp"
+#include <tracy/Tracy.hpp>
```
Kryptos-FR, line 4: Also guard it with a `#ifdef `

src/gfx/context.cpp
guusw: Doesn't need to since it does that in the header itself.
It allows you to inserts zones like `ZoneScoped;` without having to wrap those in ifdefs as well

src/gfx/renderer.cpp
guusw: See prev comment

src/gfx/texture.cpp
guusw: See prev comment

sinkingsugar (approved):
Love it but should be exposed also in the runtime, shards activations, chain activations etc!

sinkingsugar:
Add a story to add hooks into mesh tick, wire run, and shards activations.

PR #503, opened 2022-12-19
Compare with DBL_EPSILON

Kryptos-FR:
`FLT_EPSILON` isn't precise enough

- [x] add 2 test cases


Kryptos-FR:
A good reference document that I checked while looking for digits precision of `Pi`: http://www.mimirgames.com/articles/programming/digits-of-pi-needed-for-floating-point-numbers/

guusw (approved):

guusw:
I think we can just add a `Assert.IsAlmost` & `IsAlmost` where you can specify epsilon/threshold values.
Then we can replace the asserts that fail with those, I think in only expects the value to be roughly the same / doesn't care about precision loss from double => float

sinkingsugar (approved):

src/core/shards/assert.cpp
```diff
@@ -92,9 +92,146 @@ struct IsNot : public Base {
...
+    }
+    default:
+      return false;
+    }
+  }
+
+  bool isAlmost(double lhs, double rhs) { return __builtin_fabs(lhs - rhs) <= _epsilon; }
+
+  static inline Types MathTypes{{
+      CoreInfo::IntType,
```
sinkingsugar, line 209: I would remove Int types cos if not ideally epsilon could be `2` and you'd have to implement the logic.

sinkingsugar (changes requested):

src/core/shards/assert.cpp
Kryptos-FR: I added Int because there was a sample with a sequence of numbers containing both integers and decimals.

Arguably, I could remove it from the list of supported types but still support it as fallback when from inside a sequence. In fact, it could just be described in the documentation that epsilon only affects decimal number and otherwise acts like `Is`.

Kryptos-FR: I could (should) also adda check that epsilon must be strictly between 0 and 1 (both ends excluded).

sinkingsugar (commented):
So we decided not to do the Ints? @guusw cc.
I kinda like that idea tbh,

Kryptos-FR:
It complicates the implementation especially regarding sequence of values that could have both integers and decimals (or even other types such as string, color, etc.). `IsAlmost` is meant as "solving" the precision issue with comparing floating point numbers. I feel like if someone wants to compare integers they will use `Math.Abs`and `Math.Subtract` themselves.

src/core/ops_internal.cpp
```diff
@@ -610,4 +610,88 @@ bool operator==(const SHTypeInfo &a, const SHTypeInfo &b) {
...
+            _almostEqual(lhs.payload.float2Value[1], rhs.payload.float2Value[1], e));
+  case SHType::Float3:
+    return (_almostEqual(lhs.payload.float3Value[0], rhs.payload.float3Value[0], e) &&
+            _almostEqual(lhs.payload.float3Value[1], rhs.payload.float3Value[1], e) &&
+            _almostEqual(lhs.payload.float3Value[2], rhs.payload.float3Value[2], e));
+  case SHType::Float4:
+    return (_almostEqual(lhs.payload.float4Value[0], rhs.payload.float4Value[0], e) &&
+            _almostEqual(lhs.payload.float4Value[1], rhs.payload.float4Value[1], e) &&
+            _almostEqual(lhs.payload.float4Value[2], rhs.payload.float4Value[2], e) &&
+            _almostEqual(lhs.payload.float4Value[3], rhs.payload.float4Value[3], e));
```
sinkingsugar, line 639: Vectors should use vector math

src/core/ops_internal.cpp
```diff
@@ -610,4 +610,88 @@ bool operator==(const SHTypeInfo &a, const SHTypeInfo &b) {
...
+            _almostEqual(lhs.payload.int16Value[6], rhs.payload.int16Value[6], e) &&
+            _almostEqual(lhs.payload.int16Value[7], rhs.payload.int16Value[7], e) &&
+            _almostEqual(lhs.payload.int16Value[8], rhs.payload.int16Value[8], e) &&
+            _almostEqual(lhs.payload.int16Value[9], rhs.payload.int16Value[9], e) &&
+            _almostEqual(lhs.payload.int16Value[10], rhs.payload.int16Value[10], e) &&
+            _almostEqual(lhs.payload.int16Value[11], rhs.payload.int16Value[11], e) &&
+            _almostEqual(lhs.payload.int16Value[12], rhs.payload.int16Value[12], e) &&
+            _almostEqual(lhs.payload.int16Value[13], rhs.payload.int16Value[13], e) &&
+            _almostEqual(lhs.payload.int16Value[14], rhs.payload.int16Value[14], e) &&
+            _almostEqual(lhs.payload.int16Value[15], rhs.payload.int16Value[15], e));
```
sinkingsugar, line 679: Vectors should use vector math

sinkingsugar (changes requested):

src/core/ops_internal.cpp
Kryptos-FR: What kind of vector math do you want to use here?

Unrolling the loops is what the compiler will do anyway: https://godbolt.org/z/8KTxGncfe

sinkingsugar: We do have actual comparisons operators with vectors implemented for Var. Please check again. 👍

sinkingsugar: Exactly in ops.hpp btw

sinkingsugar: Err I see what you mean. Why this happened? @guusw was you or me? We should use vectors just like `==` compiler is not smart enough. 
Maybe you can refactor that too @Kryptos-FR actually. It's the whole point for us using gcc/clang vectors...

sinkingsugar: But you know what, we can avoid for now if it's annoying. Didn't realize it's so convoluted.

PR #504, opened 2022-12-21
UI markdown viewer

Kryptos-FR:
Adds a markdown viewer as a UI widget. Supports standard commonmark syntax.


sinkingsugar (approved):
Wow was super simple, no parameters at all to customize it?

Kryptos-FR:
There are a few options regarding some size constraints but I haven't found them very useful in our case, since we usually control a widget size by the layout around it.

PR #510, opened 2022-12-27
ENTT & events

sinkingsugar:
Start adding ENTT

For now mostly for events.
But other apps can use it further...

### DIFF
--- a/src/core/runtime.hpp
+++ b/src/core/runtime.hpp
@@ -538,10 +538,11 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
   SHInstanceData instanceData{};
 
 private:
+  SHMesh() = default;
+
   std::list<std::shared_ptr<SHFlow>> _flows;
   std::vector<std::string> _errors;
   std::vector<SHWire *> _failedWires;
-  SHMesh() = default;
 };
 
 namespace shards {
--- a/src/core/runtime.hpp
+++ b/src/core/runtime.hpp
@@ -378,7 +378,7 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
 
   ~SHMesh() { terminate(); }
 
-  void preCompose(const std::shared_ptr<SHWire> &wire, SHVar input = shards::Var::Empty) {
+  void compose(const std::shared_ptr<SHWire> &wire, SHVar input = shards::Var::Empty) {
     SHLOG_TRACE("Pre-composing wire {}", wire->name);
 
     if (wire->warmedUp) {
--- a/src/core/runtime.hpp
+++ b/src/core/runtime.hpp
@@ -378,6 +378,44 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
 
   ~SHMesh() { terminate(); }
 
+  void preCompose(const std::shared_ptr<SHWire> &wire, SHVar input = shards::Var::Empty) {
+    SHLOG_TRACE("Pre-composing wire {}", wire->name);
+
+    if (wire->warmedUp) {
+      SHLOG_ERROR("Attempted to Pre-composing a wire multiple times, wire: {}", wire->name);
+      throw shards::SHException("Multiple wire Pre-composing");
+    }
+
+    // this is to avoid recursion during compose
+    visitedWires.clear();
+
+    wire->mesh = shared_from_this();
+    wire->isRoot = true;
+    // remove when done here
+    DEFER(wire->isRoot = false);
+
+    // compose the wire
+    SHInstanceData data = instanceData;
+    data.wire = wire.get();
+    data.inputType = shards::deriveTypeInfo(input, data);
+    auto validation = shards::composeWire(
+        wire.get(),
+        [](const Shard *errorShard, const char *errorTxt, bool nonfatalWarning, void *userData) {
+          auto blk = const_cast<Shard *>(errorShard);
+          if (!nonfatalWarning) {
+            throw shards::ComposeError(std::string(errorTxt) + ", input shard: " + std::string(blk->name(blk)));
+          } else {
+            SHLOG_INFO("Validation warning: {} input shard: {}", errorTxt, blk->name(blk));
+          }
+        },
+        this, data);
+    shards::arrayFree(validation.exposedInfo);
+    shards::arrayFree(validation.requiredInfo);
+    shards::freeDerivedInfo(data.inputType);
+
+    SHLOG_TRACE("Wire {} Pre-composing", wire->name);
+  }
+
   struct EmptyObserver {
     void before_compose(SHWire *wire) {}
     void before_tick(SHWire *wire) {}
@@ -395,16 +433,17 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
       throw shards::SHException("Multiple wire schedule");
     }
 
-    // this is to avoid recursion during compose
-    visitedWires.clear();
-
     wire->mesh = shared_from_this();
-    wire->isRoot = true;
-    // remove when done here
-    DEFER(wire->isRoot = false);
 
     observer.before_compose(wire.get());
     if (compose) {
+      // this is to avoid recursion during compose
+      visitedWires.clear();
+
+      wire->isRoot = true;
+      // remove when done here
+      DEFER(wire->isRoot = false);
+
       // compose the wire
       SHInstanceData data = instanceData;
       data.wire = wire.get();
--- a/src/core/runtime.hpp
+++ b/src/core/runtime.hpp
@@ -1289,6 +1289,8 @@ template <typename T> struct WireDoppelgangerPool {
   void release(std::shared_ptr<T> wire) { _avail.emplace(wire); }
 
 private:
+  WireDoppelgangerPool() = delete;
+
   struct Writer {
     std::stringstream &stream;
     Writer(std::stringstream &stream) : stream(stream) {}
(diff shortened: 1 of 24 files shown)


src/core/runtime.hpp
```diff
@@ -382,6 +378,44 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
 
   ~SHMesh() { terminate(); }
 
+  void preCompose(const std::shared_ptr<SHWire> &wire, SHVar input = shards::Var::Empty) {
```
guusw, line 381: Why is it called preCompose? It seems to call composeWire, so what's different?

guusw: Can't it be just called composeWire or something

sinkingsugar: Because it implies you pass false to schedule (compose parameter).
But you are right, can just be "compose" actually.

sinkingsugar:
anything against merging? would like to get it in devel asap

guusw:
> anything against merging? would like to get it in devel asap

Let me check quickly

src/core/shards/core.hpp
```diff
@@ -3469,7 +3469,7 @@ struct Once {
         // let's cheat in this case and stop triggering this call
         self->inlineShardId = InlineShard::NoopShard;
       } else {
-        next = next + dsleep;
+        next = (current - (current - next)) + dsleep;
```
guusw, line 3472: This is just the same but more calculations 😄

src/core/shards/events.cpp
```diff
@@ -0,0 +1,113 @@
...
+    switch (index) {
+    case 0:
+      return Var(_eventName);
+    default:
+      return Var::Empty;
+    }
+  }
+
+  SHTypeInfo compose(const SHInstanceData &data) {
+    _dispatcher = GetGlobals().Dispatchers[_eventName];
```
guusw, line 39: I don't think you can keep this as reference, as soon as the collection decides to move stuff around (after adding/removing), the references will become invalid.
Might have to store the values inside a shared/unique_ptr in the map if you want to reference them by address

src/core/shards/events.cpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  void cleanup() {
+    assert(_dispatcher);
+    if(_connection)
+      _connection.release();
+  }
+
+  SHVar activate(SHContext *context, const SHVar &input) {
+    _eventsOut = std::move(_events);
```
guusw, line 90: I'm not sure but explicit clear and then std::swap might be better to avoid allocation, since it doesn't completely reset the moved from object (_events)

src/mal/SHCore.cpp
```diff
@@ -433,12 +433,18 @@ class malSHMesh : public malValue, public malRoot {
 
   SHMesh *value() const { return m_mesh.get(); }
 
-  void schedule(malSHWire *wire) {
+  void schedule(malSHWire *wire, bool compose = true) {
     auto cp = SHWire::sharedFromRef(wire->value());
-    m_mesh->schedule(cp);
+    m_mesh->schedule(cp, Var::Empty, compose);
```
guusw, line 438: Can we remove compose from schedule instead so the caller can decide by either calling:
```
compose(...)
schedule(...)
```
or just
```
schedule(...)
```
?

src/core/runtime.hpp
```diff
@@ -382,6 +378,44 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
 
   ~SHMesh() { terminate(); }
 
+  void compose(const std::shared_ptr<SHWire> &wire, SHVar input = shards::Var::Empty) {
```
guusw, line 381: Check comment in SHCore malMesh as well.
This code seems very similar to that in SHMesh::schedule, can we remove the compose code from schedule and instead have callers call:
```
compose(...)
schedule(...)
```
separately?

src/tests/events.edn
```diff
@@ -0,0 +1,19 @@
...
+
+(defloop updater
+  (Events.Update "RandomInts")
+  (Msg "Updated")
+  (Pause 0.2))
+
+(schedule main send)
+(schedule main receive)
+(schedule main updater)
+(run main 0.05 100)
```
guusw, line 19: Don't forget the (if (throw, etc.

src/tests/timing.edn
```diff
@@ -0,0 +1,12 @@
...
+(defloop
+  time
+  (Once (-> (Msg "1")) :Every 0.1)
+  (Once (-> (Msg "2")) :Every 0.2)
+  (Once (-> (Msg "3")) :Every 0.4)
+  (Once (-> (Msg "4")) :Every 0.8)
+  (Msg "main tick"))
+
+(schedule main time)
+(run main 0.1 20)
```
guusw, line 12: Also add (if .. (throw ...

guusw (changes requested):
Left some comments

src/core/shards/events.cpp
sinkingsugar: it's a `unordered_map` the memory will not be compacted or moved by spec.

src/core/shards/events.cpp
sinkingsugar: `std::move` should do exactly like you said here. (by spec)

src/core/runtime.hpp
sinkingsugar: yeah but what about edn code? keep `schedule` abstraction right?

src/mal/SHCore.cpp
sinkingsugar: yeah but what about edn code? keep schedule abstraction right?

src/core/shards/core.hpp
sinkingsugar: not exactly, the above took no care of `current` at all in fact was broken.
did you check above?

sinkingsugar: `current - next` is the drift which ideally is 0 but in reality never is 0
We subtract it from current to remove how much we drifted and add the sleep.
I think the logic is not flawed here to try to realign to real time

guusw: I mean it's mathematically equivalent: `(current - (current - next)) == next` so it still evaluates to `next + dsleep`

guusw: https://www.wolframalpha.com/input?i=c+-+%28c+-+n%29+%2B+d

src/core/shards/events.cpp
guusw: Ah, never though about that, but it seems that way yes.

src/core/shards/events.cpp
guusw: Yeah but it moves out of _events, leaving it probably empty (0 reserved memory, since it now belongs to _eventsOut)
So when new events come in, _events would have to reallocate

guusw: At least that's what I think happens

src/mal/SHCore.cpp
guusw: Yeah edn could stay the same for now I think, just talking about the cpp code

guusw: If edn turns into (shedule .. true) or false for compose, it'd split it there too.

src/core/shards/core.hpp
sinkingsugar: Oh you mean canceling each other. How do you explain that the above is broken though when used 🤣🤔? That's why I had to fix it.

src/core/shards/events.cpp
sinkingsugar: Good point with that

src/core/shards/events.cpp
sinkingsugar: It's a very very good clever trick you should use, and we use a lot even in channels.
`std::deque` also has this property which is why I use it sometimes vs vector (which invalidates)

src/core/runtime.hpp
sinkingsugar: Leaving this out as requires a bigger refactor that is a waste of time for me sadly @guusw

src/mal/SHCore.cpp
sinkingsugar: Leaving this out as requires a bigger refactor that is a waste of time for me sadly @guusw

src/core/shards/core.hpp
```diff
@@ -3379,6 +3379,7 @@ struct Once {
   SHTime current;
   SHTimeDiff next;
```
guusw, line 3380: Is this supposed to be an SHTimeDiff, next - current doesn't really make sense to me in that case.
next should also be just SHTime, no?

sinkingsugar: std::chrono is a mess, everyone knows :) 
I just know it works lol
btw it's the same in "run" command mal side

sinkingsugar: it's `diff` cos it's result of an operation. Also can't be SHTime in that case.

guusw: Ah it's the same, `SHTimeDiff` is defined as `decltype(SHClock::now() - SHDuration(0.0))` which will be the same as `SHTime`

src/mal/SHCore.cpp
guusw: https://github.com/fragcolor-xyz/shards/issues/526

guusw (approved):
LGTM

PR #512, opened 2023-01-02
UI tabs

Kryptos-FR:
Adds `(UI.DockArea)` and `(UI.Tab)`.

It works a bit differently than other UI shards: it is the parent (i.e. `UI.DockArea`) that collects the tabs so that they are all known before the first activate. This design is due to how the underlying `egui_dock` library works. It might change later if we add more layouting options ; for now, all tabs start aligned on a single tab row.

**Example**

```clj
(UI.DockArea
 :Contents
 (->
  (UI.Tab
   :Title "Tab 1"
   :Contents (-> "Tab 1 contents" (UI.Label)))
  (UI.Tab
   :Title "Tab 2"
   :Contents (-> "Tab 2 contents" (UI.Label)))
  (UI.Tab
   :Title "Tab 3"
   :Contents (-> "Tab 3 contents" (UI.Label)))
  (UI.Tab
   :Title "Tab 4"
   :Contents (-> "Tab 4 contents" (UI.Label)))))
```

tabs_2

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -43,6 +43,7 @@ chrono = { version = "0.4", optional = true }
 chacha20poly1305 = { version = "0.9.0", optional = true }
 egui = { version = "0.20.1", optional = true }
 egui_commonmark = { version = "0.6.0", optional = true }
+egui_dock = { version = "0.3.0", optional = true }
 egui_extras = { version = "0.20.0", optional = true }
 egui-gfx = { path = "../src/gfx/egui", optional = true }
 egui_memory_editor = { version = "0.2.2", optional = true }
@@ -89,13 +90,14 @@ shards = ["reqwest",
           "chacha20poly1305",
           "chrono",
           "egui",
+          "egui_commonmark",
+          "egui_dock",
           "egui_extras",
           "egui-gfx",
           "code_editor",
           "hex_viewer",
           "tract-onnx",
-          "enum-map",
-          "egui_commonmark"]
+          "enum-map"]
 code_editor = ["syntect", "egui_commonmark/syntax_highlighting"]
 hex_viewer = ["egui_memory_editor"]
 dllshard = ["dlopen"]
--- a/rust/src/shards/gui/containers/docking.rs
+++ b/rust/src/shards/gui/containers/docking.rs
@@ -1,0 +1,460 @@
+use super::DockArea;
+use super::Tab;
+use crate::core::registerShard;
+use crate::shard::Shard;
+use crate::shards::gui::util;
+use crate::shards::gui::CONTEXTS_NAME;
+use crate::shards::gui::EGUI_CTX_TYPE;
+use crate::shards::gui::PARENTS_UI_NAME;
+use crate::types::Context;
+use crate::types::ExposedInfo;
+use crate::types::ExposedTypes;
+use crate::types::InstanceData;
+use crate::types::OptionalString;
+use crate::types::ParamVar;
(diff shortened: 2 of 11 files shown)


rust/Cargo.toml
```diff
@@ -89,13 +90,14 @@ shards = ["reqwest",
           "chacha20poly1305",
           "chrono",
           "egui",
+          "egui_commonmark",
+          "egui_dock",
```
sinkingsugar, line 94: these are on by default right?

rust/src/shards/gui/containers/docking.rs
```diff
@@ -0,0 +1,441 @@
...
+
+  fn getParam(&mut self, index: i32) -> Var {
+    match index {
+      0 => self.title.get_param(),
+      1 => self.contents.get_param(),
+      _ => Var::default(),
+    }
+  }
+
+  fn requiredVariables(&mut self) -> Option<&ExposedTypes> {
```
sinkingsugar, line 135: Shouldn't this also expose? or it can be "hidden" / "not execute" conditionally?

Kryptos-FR: Yeah. Likely forgot expose here. Since we can't close tabs for now, they are not executed conditionally.

rust/Cargo.toml
Kryptos-FR: yes

Kryptos-FR:
Outstanding issue filled in `egui_dock` repo: https://github.com/Adanos020/egui_dock/issues/90

Currently our workarounds are:
- either not use a dock area inside of a Window
- use it only in Window with fixed size

rust/src/shards/gui/containers/docking.rs
```diff
@@ -0,0 +1,460 @@
...
+
+  fn parameters(&mut self) -> Option<&Parameters> {
+    Some(&DOCKAREA_PARAMETERS)
+  }
+
+  fn setParam(&mut self, index: i32, value: &Var) -> Result<(), &str> {
+    match index {
+      0 => {
+        let seq = Seq::try_from(value)?;
+        for tab_spec in seq.iter().rev().filter_map(|s| {
```
guusw, line 264: Can you maybe assign the result of filter_map to a variable to make it more readable?

rust/src/shards/gui/containers/docking.rs
```diff
@@ -0,0 +1,460 @@
...
+          if let Ok(r) = ShardRef::try_from(s) {
+            if r.name() == TAB_NAME {
+              let mut title = ParamVar::default();
+              title.set_param(&r.get_parameter(0));
+              let mut shard = ShardsVar::default();
+              match shard.set_param(&s) {
+                Ok(_) => Some(Ok((title, shard))),
```
guusw, line 271: Call it `(title, contents)` like below. Maybe rename the struct member too: shards => contents

guusw (approved):
Looks good, just had some style suggestion.

rust/src/shards/gui/containers/docking.rs
Kryptos-FR: There is already a struct member named contents (which is the parameter).

rust/src/shards/gui/containers/docking.rs
Kryptos-FR: I don't think it makes it more readable since it is an iterator. It's quite idiomatic to use a for loop on an iterator.

PR #514, opened 2023-01-05
MainWindow compose stack size

guusw:
Easy enough, since wire is already exposed during compose, I can just extend the stack size.

### DIFF
--- a/src/extra/gfx/main_window.cpp
+++ b/src/extra/gfx/main_window.cpp
@@ -85,6 +85,9 @@ struct MainWindow : public Base {
   }
 
   SHTypeInfo compose(SHInstanceData &data) {
+    // Require extra stack space for the wire containing the main window
+    data.wire->stackSize = std::max<size_t>(data.wire->stackSize, 4 * 1024 * 1024);
+
     if (data.onWorkerThread) {
       throw ComposeError("GFX Shards cannot be used on a worker thread (e.g. "
                          "within an Await shard)");


src/extra/gfx/main_window.cpp
```diff
@@ -85,6 +85,9 @@ struct MainWindow : public Base {
   }
 
   SHTypeInfo compose(SHInstanceData &data) {
+    // Require extra stack space for the wire containing the main window
+    data.wire->stackSize = std::max<size_t>(data.wire->stackSize, 4 * 1024 * 1024);
```
sinkingsugar, line 89: Not sure this is good. At least should be a constant in `foundation.hpp` or so.

Kryptos-FR: Does it need to be that big, or could it also work with a smaller value (like `512 * 1024`)?

guusw: It doesn't that matter that much how big (this 4MB) since it'll only get applied to the wire containing MainWindow (of which there is only 1 in normal cases)

guusw: I'd rather avoid cryptic bugs than setting it too low and reducing application memory footprint by like 2 megabytes.

sinkingsugar: Yeah ok, and yeah 4MB is ok 👌 
Just consider linker by default has 1MB set on windows. But it's true that some apps (likely game engines etc might go higher.)

Kryptos-FR (approved):

PR #516, opened 2023-01-10
Documentation: enums

Kryptos-FR:
**Description**
Automatically create links to the corresponding enum description page from the API pages.

Also support writing the enum description in rust code.

**Syntax**
```rust
shenum! {
  struct UIProperty {
    [description("Returns the remaining space")]
    const RemainingSpace = 0x0;
  }
  struct UIPropertyInfo {}
}
```

It resembles the macro syntax but without the `#`. This has two advantages:
- clearly show that it is not a dedicated macro: it is just a syntax sugar interpreted by the `shenum!` macro itself
- allow to still attach other macros individually to enum values if needed:
  ```rust
  shenum! {
    struct UIProperty {
      [description("Returns the remaining space")]
      #[MyMacro)]
      const RemainingSpace = 0x0;
    }
    struct UIPropertyInfo {}
  }
  ```

Note: it is also mandatory on all enum values. So at the very least, it needs to have an empty string:

```rust
shenum! {
  struct UIProperty {
    [description("")]
    const RemainingSpace = 0x0;
  }
  struct UIPropertyInfo {}
}
```

### DIFF
--- a/docs/generate.edn
+++ b/docs/generate.edn
@@ -30,6 +30,33 @@
     .f (ToString))
    true))
 
+(defn print-type-info [buffer]
+  (-> (ExpectTable)
+      (| (Take "name") (ExpectString) >= .p_type_name)
+      (| (Take "type") (ExpectInt) >= .p_type_type)
+      "[`" >> buffer .p_type_name >> buffer "`](" >> buffer
+      .p_type_type
+      (Match
+       [2; Enum
+        (-> "../../../enums/" >> buffer
+            .p_type_name >> buffer)
+        15 ; ShardRef
+        (-> "../../types/#shard" >> buffer)
+        54 ; ContextVar
+        (-> "../../types/#contextvar" >> buffer)
+        56 ; Seq
+        (-> "../../types/#seq" >> buffer)
+        57 ; Table
+        (-> "../../types/#table" >> buffer)
+        59 ; Object
+        (-> "../../types/#object" >> buffer)
+        61 ; Set
+        (-> "../../types/#set" >> buffer)
+        nil
+        (-> "../../types/#" >> buffer
+            .p_type_name (String.ToLower) >> buffer)])
+      ")" >> buffer))
+
 ;; dumps the generated content to the final file
 (defn dump-shard [name]
   (Wire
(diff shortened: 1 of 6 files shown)


Kryptos-FR:
@ChelseaChX please review the enum descriptions in all these 3 files:
- `rust/src/shards/gui/containers/mod.rs`
- `rust/src/shards/gui/properties.rs`
- `rust/src/shards/gui/widgets/plots/mod.rs`

ChelseaChX:
> @ChelseaChX please review the enum descriptions in all these 3 files:
> 
> * `rust/src/shards/gui/containers/mod.rs`
> * `rust/src/shards/gui/properties.rs`
> * `rust/src/shards/gui/widgets/plots/mod.rs`

Looks good!

docs/generate.edn
```diff
@@ -30,6 +30,33 @@
...
    true))
 
+(defn print-type-info [buffer]
+  (-> (ExpectTable)
+      (| (Take "name") (ExpectString) >= .p_type_name)
+      (| (Take "type") (ExpectInt) >= .p_type_type)
+      "[`" >> buffer .p_type_name >> buffer "`](" >> buffer
+      .p_type_type
+      (Match
+       [2; Enum
```
guusw, line 40: Don't we have an enum for this? (Type.Enum or something?)
Maybe we can add one?

guusw (approved):
One thing, but for the rest LGTM

rust/src/shards/gui/containers/mod.rs
```diff
@@ -23,14 +23,23 @@ struct Area {
 
 shenum! {
   struct Anchor {
+    [description("Top left corner.")]
```
guusw, line 26: Shouldn't it be like
```rs
#[description("Middle left.")]
const Left = 0x10;
```
to match how attributes look in rust

Kryptos-FR: I tried that but then I also can't allow additional attributes at the same time with the ` $(#[$attr:ident $($args:tt)*])*` rule. The macro system thinks it is ambiguous to have:
```rust
$(
  #[description($desc:literal)]
  $(#[$inner:ident $($args:tt)*])*
  const $EnumValue:ident = $value:expr;
)+
```

Now it's possible we will never need to add other attributes but I preferred to be future-proof. In addition it clearly shows that those are not the usual attributes.

docs/generate.edn
Kryptos-FR: We do have a [`Type`](https://docs.fragcolor.xyz/reference/enums/Type/) enum. I'll try to see if I can use it.

rust/src/shards/gui/containers/mod.rs
guusw: Yeah okay, makes sense

PR #522, opened 2023-01-16
Fix D3D12 warp performance

guusw:
Performance for running the windows tests in the software rasterizer seems to have degraded after #500 (Vulkan seems unaffected)
Comparing job [6658454962](https://github.com/fragcolor-xyz/shards/actions/runs/3898687960/jobs/6658454962) to [6669771237](https://github.com/fragcolor-xyz/shards/actions/runs/3903338722/jobs/6669771237)


sinkingsugar:
Is the actual real GPU performance affected?

guusw:
Not that I noticed, since Vulkan tests on the GPU runner seems to take the same length still.

guusw:
Might be the buffer mapping but I want to change that anyway later so that it's done between frames (as part of a bigger change).

PR #525, opened 2023-01-17
Virtual UI panels

guusw:
MVP for world space UI, tried to keep the PR as small as possible
- Added basic world space UI rendering and input.
- Added some basic documentation (`reference/shards/VUI/Context/`)


src/virtual_ui/context.cpp
```diff
@@ -0,0 +1,277 @@
...
+      PointerInput outEvt;
+      outEvt.iterator = it;
+      outEvt.ray = sizedView.getRay(pointerCoords * inputToViewScale);
+      pointerInputs.push_back(outEvt);
+    } else {
+      otherEvents.push_back(it);
+    }
+  }
+
+  struct PanelInput {
```
guusw, line 43: @guusw  remove

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,257 @@
...
+    // TODO: Move to input context
+    _inputBuffer.clear();
+    for (auto &event : _inputContext->events)
+      _inputBuffer.push_back(event);
+
+    // Evaluate all UI panels
+    _vuiContext.activationContext = shContext;
+    _vuiContext.context.virtualPointScale = _scale.payload.floatValue;
+    withObjectVariable(*_vuiContextVar, &_vuiContext, VUIContext::Type, [&]() {
+      gfx::float2 inputToViewScale{1.0f};
```
guusw, line 108: @guusw This should probably be set based on Window

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+void VUIContextShard::warmup(SHContext *context) {
+  _vuiContextVar = referenceVariable(context, VUIContext::VariableName);
+
+  _inputContext.warmup(context);
+  _graphicsContext.warmup(context);
+
+  withObjectVariable(*_vuiContextVar, &_vuiContext, VUIContext::Type, [&]() { PARAM_WARMUP(context); });
+
+  // Collect VUI.Panel shards similar to UI dock does
```
guusw, line 239: @guusw remove

src/gfx/egui/renderer.cpp
```diff
@@ -166,62 +168,68 @@ struct EguiRendererImpl {
...
+      if (texture) {
+        drawable.parameters.set("color", texture);
+        if (texture->getFormat().pixelFormat == WGPUTextureFormat_R8Unorm) {
+          drawable.parameters.set("flags", uint32_t(0x1));
+        }
       }
-    }
 
-    drawable.clipRect = int4(prim.clipRect.min.x, prim.clipRect.min.y, prim.clipRect.max.x, prim.clipRect.max.y);
+      // drawable.clipRect = int4(prim.clipRect.min.x, prim.clipRect.min.y, prim.clipRect.max.x, prim.clipRect.max.y);
```
guusw, line 205: TODO: Make this switchable for world space UI, also make it work by either rendering to texture first or using stencil

src/virtual_ui/context.cpp
```diff
@@ -0,0 +1,277 @@
...
+}
+
+void Context::renderDebug(gfx::ShapeRenderer &sr) {
+  float uiToWorldScale = 1.0f / virtualPointScale;
+
+  for (const auto &panel : panels) {
+    auto geom = panel->getGeometry().scaled(uiToWorldScale);
+    float4 c = panel == focusedPanel ? float4(0, 1, 0, 1) : float4(0.8, 0.8, 0.8, 1.0);
+    sr.addRect(geom.center, geom.right, geom.up, geom.size, c, 2);
+    // panel->transform
```
guusw, line 100: @guusw Remove

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,257 @@
...
+  PARAM_PARAMVAR(_view, "View", "The view that is being used to render.", {Type::VariableOf(gfx::Types::View)});
+  PARAM(ShardsVar, _contents, "Contents", "The list of UI panels to render.", {CoreInfo::ShardsOrNone});
+  PARAM_VAR(_scale, "Scale", "The scale of how many UI units per world unit.", {CoreInfo::FloatType});
+  PARAM_VAR(_debug, "Debug", "Render debug.", {CoreInfo::BoolType});
+  PARAM_IMPL(VUIContextShard, PARAM_IMPL_FOR(_queue), PARAM_IMPL_FOR(_view), PARAM_IMPL_FOR(_contents), PARAM_IMPL_FOR(_scale),
+             PARAM_IMPL_FOR(_debug));
+
+  VUIContextShard() {
+    _scale = Var(1000.0f);
+    _debug = Var(true);
```
guusw, line 44: @guusw change default to false

src/tests/vui.edn
```diff
@@ -0,0 +1,87 @@
...
+         (UI.LeftPanel
+          :Contents
+          (->
+           "Left Panel" (UI.Label)
+           (UI.Button :Label "Button")))))
+
+    (GFX.Render :Steps .render-steps :View .view))))
+
+(schedule Root test-wire)
+(if (run Root timestep) nil (throw "Root tick failed"))
```
Kryptos-FR, line 87: If we plan to run it during CI, it needs to have a fixed number of frames to exit.

sinkingsugar (changes requested):

rust/src/shards/gui/raw_context.rs
```diff
@@ -0,0 +1,254 @@
+use egui_gfx::egui_FullOutput;
+use egui_gfx::egui_Input;
+use egui_gfx::make_native_full_output;
+use egui_gfx::NativeFullOutput;
+
+use super::util;
```
Kryptos-FR, line 6: I usually group those (i.e. no empty lines) and they end-up at the bottom.

rust/src/shards/gui/raw_context.rs
```diff
@@ -0,0 +1,254 @@
...
+struct UIContext {
+  context: Option<egui::Context>,
+  full_output: Option<NativeFullOutput>,
+  instance: ParamVar,
+  parents: ParamVar,
+  exposed: Vec<ExposedInfo>,
+}
+
+impl UIContext {
+  pub fn default() -> Self {
```
Kryptos-FR, line 46: Shouldn't it implement the `Default` trait instead?

```rust
impl Default for UIContext {
  fn default() -> Self;
}
```

Otherwise, call it `new()` to avoid confusion.

src/extra/egui/context.cpp
```diff
@@ -0,0 +1 @@
+#include "context.cpp"
```
Kryptos-FR, line 1: Infinite import loop? I'm surprised it actually compiles, but probably because the file is not referenced in the corresponding `CMakeLists.txt`.

```suggestion
#include "context.hpp"
```

src/extra/egui/pass.cpp
```diff
@@ -48,5 +48,4 @@ struct UIPassShard {
 };
 
 void registerShards() { REGISTER_SHARD("GFX.UIPass", UIPassShard); }
-} // namespace Gui
-} // namespace shards
\ No newline at end of file
+} // namespace shards::egui
```
Kryptos-FR: Missing EOL

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,243 @@
...
+  PARAM_IMPL(VUIContextShard, PARAM_IMPL_FOR(_queue), PARAM_IMPL_FOR(_view), PARAM_IMPL_FOR(_contents), PARAM_IMPL_FOR(_scale),
+             PARAM_IMPL_FOR(_debug));
+
+  VUIContextShard() {
+    _scale = Var(1000.0f);
+    _debug = Var(false);
+  }
+
+  static SHTypesInfo inputTypes() { return CoreInfo::NoneType; }
+  static SHTypesInfo outputTypes() { return CoreInfo::NoneType; }
```
Kryptos-FR, line 48: Try to add the corresponding `inputHelp()` and `outputHelp()`.

Also the idiomatic way for a shard that doesn't do anything with its input is to return it ("pass through"), in other words, `CoreInfo::AnyType`. Unless there is a reason here to make it a "black-hole"?

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,243 @@
...
+  PARAM_IMPL(VUIPanelShard, PARAM_IMPL_FOR(_transform), PARAM_IMPL_FOR(_size), PARAM_IMPL_FOR(_contents));
+
+  RequiredVUIContext _context;
+  Shard *_uiShard{};
+  EguiContext _eguiContext;
+  ExposedInfo _exposedTypes;
+  std::shared_ptr<Panel> _panel;
+
+  static SHTypesInfo inputTypes() { return CoreInfo::NoneType; }
+  static SHTypesInfo outputTypes() { return CoreInfo::NoneType; }
```
Kryptos-FR, line 144: Similar remark. I would expect the shard to be "pass through".

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,243 @@
...
+    if (_queue->valueType == SHType::None)
+      throw ComposeError("Queue is required");
+
+    if (_view->valueType == SHType::None)
+      throw ComposeError("View is required");
+
+    return shards::CoreInfo::NoneType;
+  }
+
+  SHVar activate(SHContext *shContext, const SHVar &input) {
```
Kryptos-FR, line 92: Is there no way to pass the input of `VUI.Context` as input of each `VUI.Panel`?

Kryptos-FR (changes requested):

docs/samples/shards/VUI/Context/1.edn
```diff
@@ -0,0 +1,55 @@
...
+  ; Camera
+ {:Position (Float3 1 2 8) :Target (Float3 0 0 0)} (Math.LookAt) >= .view-transform
+ (GFX.View :View .view-transform) >= .view)
+
+(GFX.MainWindow
+ :Title "World space UI" :Width 1280 :Height 720
+ :Contents
+ (->
+  .queue (GFX.ClearQueue)
+  (VUI.Context
```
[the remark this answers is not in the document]
Kryptos-FR: I think we might need to differentiate with other UI shards. However since we already have `UI` instead of `UI.Context`, we could name it `VUI`.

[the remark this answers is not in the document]
sinkingsugar: I don't buy the "virtual" yet though. 😄`3UI` 😄?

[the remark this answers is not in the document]
guusw: Yeah I'm open to name suggestions, 3UI probably not the best because it starts with a number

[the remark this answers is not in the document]
guusw: WorldUI maybe?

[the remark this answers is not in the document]
guusw: WUI 😄 ?

src/extra/egui/context.cpp
guusw: Yeah should be the header

src/extra/vui/vui.cpp
guusw: Yeah I can do that I think

guusw: I tried to look at your solution for `UI.DockArea`, does it do that too?

src/extra/egui/context.cpp
guusw: Actually removed

src/extra/vui/vui.cpp
guusw: Did passthrough

src/extra/vui/vui.cpp
guusw: Did passthrough

guusw: Although `VUI.Panel` is not supposed to be used directly so input/output is not entirely accurate. I set InputType to any for now

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+  input::InputBuffer _inputBuffer;
+
+  std::shared_ptr<gfx::GizmoRenderer> _debugRenderer;
+
+  PARAM_PARAMVAR(_queue, "Queue", "The draw queue to insert draw commands into.", {Type::VariableOf(gfx::Types::DrawQueue)});
+  PARAM_PARAMVAR(_view, "View", "The view that is being used to render.", {Type::VariableOf(gfx::Types::View)});
+  PARAM(ShardsVar, _contents, "Contents", "The list of UI panels to render.", {CoreInfo::ShardsOrNone});
+  PARAM_VAR(_scale, "Scale", "The scale of how many UI units per world unit.", {CoreInfo::FloatType});
+  PARAM_VAR(_debug, "Debug", "Render debug.", {CoreInfo::BoolType});
```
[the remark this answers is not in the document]
guusw: Changed

src/extra/vui/vui.cpp
Kryptos-FR: Yes.

It is given to the TabViewer (https://github.com/fragcolor-xyz/shards/blob/devel/rust/src/shards/gui/containers/docking.rs#L404). Then, in turn, passed to the contents children (https://github.com/fragcolor-xyz/shards/blob/devel/rust/src/shards/gui/containers/docking.rs#L453).

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,244 @@
...
+  _vuiContextVar = referenceVariable(context, VUIContext::VariableName);
+
+  _inputContext.warmup(context);
+  _graphicsContext.warmup(context);
+
+  withObjectVariable(*_vuiContextVar, &_vuiContext, VUIContext::Type, [&]() { PARAM_WARMUP(context); });
+}
+
+void registerShards() {
+  REGISTER_SHARD("VUI.Context", VUIContextShard);
```
sinkingsugar, line 241: `Spatial.UI`

src/extra/vui/vui.cpp
```diff
@@ -0,0 +1,244 @@
...
+  REGISTER_SHARD(VUI_PANEL_SHARD_NAME, VUIPanelShard);
```
sinkingsugar, line 242: `Spatial.Panel`

sinkingsugar (changes requested):

guusw:
Renamed to Spatial.UI & Panel

Kryptos-FR:
I have an issue with right-clicking. For some reason it doesn't work in debug but does in release (even after removing the `FreeCamera` which was capturing it to move the view). I don't have an easy reproduction case right now, as it needs to create a shard with ui that responds to the right-click specifically, which we don't have any example at the moment.

```rust
// in UI's shard activate
let mouse = &ui.ctx().input().pointer;
if mouse.secondary_released() {
  // do something
  shlog!("right mouse clicked");
}
```

guusw:
Yes thanks for reminding, I'll fix that @Kryptos-FR

sinkingsugar (approved):

PR #529, opened 2023-01-20
Log slow ticks at most once every 1000

Kryptos-FR:
Log when a frame takes too long but only do it at most every `1000` frames (regardless of actual time rate).

Applies to both the main loop of a wire running on a mesh, and to the specific repeat loop of `Once` shard.


sinkingsugar (approved):
much better logic 👍

PR #535, opened 2023-01-23
Don't use oldtime feature with chrono crate

Kryptos-FR:
Following the dependabot alert, don't use the old version of `time` as a dependency for `chrono`.

References
- https://github.com/fragcolor-xyz/shards/security/dependabot/4
- https://github.com/chronotope/chrono/issues/876
- https://docs.rs/crate/chrono/0.4.20/features (page is not available for 0.4.19).


Kryptos-FR:
@sinkingsugar you might want to look at it since you were solving some semver issue in #331. Possibly, disabling the `oldtime` feature flag would now fix it?

sinkingsugar (approved):

PR #536, opened 2023-01-26
Fix compress-strings script

Kryptos-FR:
shards' `help()` functions were not exported.

We force populating the optional strings by retrieving `(shard-info)` on every shard.


sinkingsugar (approved):
Ah it's automated from
CI now?
I remember I was doing all manually when needed (very rarely)

Kryptos-FR:
@sinkingsugar It is not fully-automated. We still need to trigger the CI job manually, but it could become automated at some point if we want to.

PR #539, opened 2023-01-30
Fixup output type during compose

Kryptos-FR:
Let's the shard declare both as potential output types.

I think this should be the idiomatic way to declare input/output types, list all potential types and only fixup during compose if necessary. And as a bonus, it makes introspection easier (for generating the doc as well as with the new shard editor).

### DIFF
--- a/rust/src/shards/http.rs
+++ b/rust/src/shards/http.rs
@@ -9,6 +9,7 @@ use crate::shard::Shard;
 use crate::types::common_type;
 use crate::types::ClonedVar;
 use crate::types::Context;
+use crate::types::InstanceData;
 use crate::types::ParamVar;
 use crate::types::Parameters;
 use crate::types::RawString;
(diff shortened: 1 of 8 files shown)


Kryptos-FR:
There are a few other shards that have conditional logic in the  `outputTypes()` function. Should they also be changed?

**`flow.hpp`**
- `BaseSubFlow::outputTypes()`: line 333

**`casting.hpp`**
- `ToNumber::outputTypes()`: line 20
- `MakeVector::outputTypes()`: line 222

**`genetic.hpp`**
- `Mutant::outputTypes()`: line 605
- `DShard::outputTypes()`: line 1100

**`http.cpp`**
- `Base::outputTypes()`: line 108 (with emscripten)

**`seqs.cpp`**
- `IndexOf::outputTypes()`: line 203

**`serialization.cpp`**
- `WritePNG::outputTypes()`: line 371

sinkingsugar:
Hmm I see, documentation doesn't pick those right

guusw:
Yeah I've been using that style as well, I think it's better to do it that way, no way to pick it up in the documentation otherwise.

sinkingsugar:
This is not complete @Kryptos-FR I think. Not sure if the list you gave is the full version either. Should be addressed soon 👍

guusw (approved):

PR #540, opened 2023-01-31
GFX.Feature Generated

guusw:
Add generator wires to GFX.Feature. They act as a kind of callback in order to generated precomputed data for the current view being rendered

TODO:
- [ ] Remove currently existing DrawData and replace it with this


guusw:
Closed for #565

PR #541, opened 2023-02-01
Fix local memory reference & logging

guusw:
Fix for something that might have caused a crash in the infos.clj test


src/core/shards/casting.hpp
```diff
@@ -249,7 +249,7 @@ template <SHType ToType> struct MakeVector {
     static Parameters params = []() {
       Parameters params;
       auto vectorType = VectorTypeLookup::getInstance().get(ToType);
-      auto paramType = getCompatibleUnitType(vectorType->numberType);
+      const Type &paramType = getCompatibleUnitType(vectorType->numberType);
```
sinkingsugar, line 252: ouch

sinkingsugar (approved):

src/core/shards/casting.hpp
Kryptos-FR: Just curious: what wouldn't `auto` grab the correct type here? Or should have it been `auto&`?

Kryptos-FR (approved):

src/core/shards/casting.hpp
guusw: auto always tries to remove reference types, so you explicitly need to use auto& when you want to store a reference.
(doesn't remove pointers)

guusw: I think const carries over, so you can assign a const reference to `auto& something = ...`

PR #543, opened 2023-02-01
fix some issues in wires.cpp

sinkingsugar:


Kryptos-FR:
Did you just bump `tract-onnx` version ? I don't have that many changes in Cargo.lock when I do it locally.
Or did you do a full `cargo update`?

On that note, is it something we should automate somehow (or have a manual CI job, just like the string one)?

sinkingsugar:
> Did you just bump `tract-onnx` version ? I don't have that many changes in Cargo.lock when I do it locally. Or did you do a full `cargo update`?
> 
> On that note, is it something we should automate somehow (or have a manual CI job, just like the string one)?

deleted cargo and regenerated yea

Kryptos-FR:
There is another tool we can also consider in such case: `cargo upgrade` (which is part of `cargo-edit`). It updates `Cargo.toml` with latest compatible versions (based on SemVer I think).

Kryptos-FR:
Re: build failures. It looks like it is caused by `webbrowser-rs` version `0.8.4+` which now uses AssocQueryStringW Windows API (cf. https://github.com/amodm/webbrowser-rs/commit/ff7b51e33d8b1b37e3669de3a36907cae2ea856d). We should keep version `0.8.3` in `cargo.lock`

PR #545, opened 2023-02-02
WS c++ -> rust simpler implementation

sinkingsugar:


rust/src/shards/ws.rs
```diff
@@ -0,0 +1,323 @@
...
+    // we need to keep this alive here, otherwise it will be dropped at the end of this function
+    self.ws = Rc::new(Some(ws.0));
+    let var_ws = Var::new_object(&self.ws, &WS_CLIENT_TYPE);
+    Ok(var_ws)
+  }
+}
+
+#[derive(Default)]
+struct ClientUser {
+  ws: Rc<Option<WsClient>>,
```
guusw, line 86: Why not `Option<Rc<WsClient>>` or even just `Option<WsClient>`? It doesn't need to stay alive past the lifetime of the `Client` shard I believe.

rust/src/shards/ws.rs
```diff
@@ -0,0 +1,323 @@
...
+    self.client.warmup(context)
+  }
+
+  fn cleanup(&mut self) -> Result<(), &str> {
+    self.client.cleanup()
+  }
+
+  fn activate(&mut self, _context: &Context, input: &Var) -> Result<Var, &str> {
+    // TODO make this run on a worker thread but not using BlockingShard as would overlap/crash
+    let ws = self.client.activate()?;
```
guusw, line 305: Where is the overlap, is it in the websocket library itself?

rust/src/shards/ws.rs
sinkingsugar: yeah has to be that way so the client could even disappear

rust/src/shards/ws.rs
sinkingsugar: no it was here, both were using await

PR #546, opened 2023-02-02
Remove dispatch

guusw:
Removing Dispatch to simplify the language, I feel like we don't need it since the only difference is passthrough.
If you really didn't want the output you could write:
`(| (Do wire))`


src/tests/general.edn
```diff
@@ -296,11 +296,6 @@
    (Count "tab11")
    (Assert.Is 2 true)
 
-   "My input"
-   (Dispatch inner1)
-   (Assert.Is "My input" true)
-   (Log)
-
```
Kryptos-FR, line 303: I would keep that sample and use `(| (Do inner1))` here. It may be a bit redundant with other tests, but looking at git history later in the future will clearly show the intent of the removal of Dispatch that way.

Kryptos-FR (changes requested):

src/tests/general.edn
sinkingsugar: Yeah might break coverage too maybe

guusw: To me it doesn't make much sense to leave that as a test here if it's `(| (Do inner1))`, it wouldn't tests anything new.
If you wanted to figure out what happened to Dispatch, you'd just check the commit logs

guusw: I added this PR comment to the commit message so people can find it

Kryptos-FR (approved):

PR #547, opened 2023-02-03
Implement Debug trait for SHVar

Kryptos-FR:
**Description**
Print a pretty-log for `Var` to help debugging.

**Rationale**
`Debug` trait on `SHVarPayload` cannot be implemented automatically because it is an `unsafe` union. And transitively, `Debug` is also not automatically implemented for `SHVar`.

**Additional details**
Here are a few examples.

**SHEnum**
```
Var {
    payload: SHEnum {
        value: 1,
        vendorId: 1718772071,
        typeId: 1819240268,
    },
    valueType: 2,
    innerType: 0,
    ..
},
``` 
**SHInt2**
```
Var {
    payload: [1, 2],
    valueType: 5,
    innerType: 0,
    ..
},
```
**SHColor**
```
Var {
    payload: SHColor { r: 255, g: 128, b: 0, a: 128 },
    valueType: 14,
     innerType: 0,
    ..
},
```
**SHBytes**
```
 Var {
    payload: SHBytes {
        value: 0x21efe4e7320,
        size: 3,
        capacity: 3,
    },
    valueType: 51,
    innerType: 0,
    ..
}
```
Note: the bytes value is not interpreted. Instead it displays the pointer address. It will act similarly for all pointer-like struct to not print recursively.

### DIFF
--- a/rust/src/core.rs
+++ b/rust/src/core.rs
@@ -5,16 +5,7 @@
 
 use crate::shard::shard_construct;
 use crate::shard::Shard;
-use crate::shardsc::SHBool;
-use crate::shardsc::SHContext;
-use crate::shardsc::SHCore;
-use crate::shardsc::SHEnumInfo;
-use crate::shardsc::SHOptionalString;
-use crate::shardsc::SHString;
-use crate::shardsc::SHStrings;
-use crate::shardsc::SHVar;
-use crate::shardsc::SHWireState;
-use crate::shardsc::ShardPtr;
+use crate::shardsc::*;
 use crate::types::ClonedVar;
 use crate::types::Context;
 use crate::types::DerivedType;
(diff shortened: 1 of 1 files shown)


sinkingsugar:
It's good but should have called C side formatter somehow to avoid duplicating debug code and consistency. Now if we add or remove another type we got another place to fix (we have many sadly 😜)

sinkingsugar (approved):

guusw (approved):
Looks good.
So core::fmt::Debug is actually what you see in the debugger (GDB/LLDB)? didn't know that.

sinkingsugar:
> Looks good. So core::fmt::Debug is actually what you see in the debugger (GDB/LLDB)? didn't know that.

I don't think so?

Kryptos-FR:
> It's good but should have called C side formatter somehow to avoid duplicating debug code and consistency. Now if we add or remove another type we got another place to fix (we have many sadly 😜)

Both implementations are different about what is printed. The formatter is meant for user consumption and only formats the value. This debug format displays more details such as inner types and other "properties" and also it is not recursive (while the formatter is).

PR #549, opened 2023-02-04
Update build documentation (windows/android)

guusw:
Update build documentation for clang toolchain (windows)


Kryptos-FR:
Need to update the CI workflows as well and to update rust installation steps (if we use msvc instead of gnu).

PR #550, opened 2023-02-06
Any type as input if Expect is the first shard

sinkingsugar:
Allow wires (especially valid for pure ones) to have Any type as input if Expect is the first shard


guusw (approved):
Seems okay.
Still not sure why it actually needs an exception and can't just match `Expect`'s Any input type against whatever input type is passed at runtime (using Do)?

sinkingsugar:
> Seems okay. Still not sure why it actually needs a special bool to disable chekcing and can't just match `Expect`'s Any input type against whatever input type is passed at runtime (using Do)?

well it's simply because there is an explicit check in `wires.cpp` trying to match the input and the wire input type.

PR #553, opened 2023-02-07
refactor "Wire" builtin

sinkingsugar:
split Def and Impl to allow to use wire variable inside self.
It's a kind of magic.

### DIFF
--- a/src/tests/flows.edn
+++ b/src/tests/flows.edn
@@ -442,3 +442,17 @@
 (schedule Root test-worker-1)
 (schedule Root test-worker-2)
 (if (run Root 0.1 20) nil (throw "Root tick failed"))
+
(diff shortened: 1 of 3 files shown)


src/tests/flows.edn
```diff
@@ -442,3 +442,17 @@
 (schedule Root test-worker-1)
 (schedule Root test-worker-2)
 (if (run Root 0.1 20) nil (throw "Root tick failed"))
+
+(Wire "test" (-> (Msg "Hello")))
+
+(defwire a-perfect-circle
+  (Log)
+  (When (Is false) (Stop))
+  false (Spawn a-perfect-circle))
```
Kryptos-FR, line 451: 🤯

Kryptos-FR (approved):

PR #554, opened 2023-02-07
refactor Join

sinkingsugar:
split into Bytes.Join, String.Join

### DIFF
--- a/src/core/shards/strings.cpp
+++ b/src/core/shards/strings.cpp
@@ -146,11 +146,13 @@ struct Replace : public Common {
 };
 
 struct Join {
+  static inline Type InputType = Type::SeqOf(CoreInfo::StringOrBytes);
+
   static SHOptionalString help() {
     return SHCCSTR("Concatenates all the elements of a string sequence, using the specified separator between each element.");
   }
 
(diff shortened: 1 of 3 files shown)


src/core/shards/strings.cpp
```diff
@@ -187,7 +189,8 @@ struct Join {
     _buffer.append(input.payload.seqValue.elements[0].payload.stringValue, SHSTRLEN(input.payload.seqValue.elements[0]));
 
     for (uint32_t i = 1; i < input.payload.seqValue.len; i++) {
-      assert(input.payload.seqValue.elements[i].valueType == SHType::String);
+      assert(input.payload.seqValue.elements[i].valueType == SHType::String ||
+             input.payload.seqValue.elements[i].valueType == SHType::Bytes);
       _buffer.append(_separator);
       _buffer.append(input.payload.seqValue.elements[i].payload.stringValue, SHSTRLEN(input.payload.seqValue.elements[i]));
```
Kryptos-FR, line 195: Here we use `append()` while in `Bytes.Join` we use `insert()`. I assume they are equivalent but I think we should be consistent in both implementations and use only one way of doing it.

PR #555, opened 2023-02-08
Fix Table Serialization

sinkingsugar:
Especially when shards are added into a table + leaking fixes


guusw:
Replaced by other changes

PR #556, opened 2023-02-08
Add reference counting to ShardRef variable type

guusw:
Almost all instances where Shard* is passed around, it will now be reference counted.
This allows shards that have been deserialized to be passed around inside Seq, Table and nested combinations.

Also fixes some incorrect usage of malSHVar with cloned variables.


include/shardwrapper.hpp
```diff
@@ -46,6 +46,8 @@ template <class T> struct ShardWrapper {
   static __cdecl Shard *create() {
     Shard *result = reinterpret_cast<Shard *>(new (std::align_val_t{16}) ShardWrapper<T>());
 
+    result->refCount = 0;
```
sinkingsugar, line 49: Well the allocation above will also zero memory so no need.

include/utility.hpp
```diff
@@ -465,7 +465,7 @@ template <class SH_CORE> class TShardsVar {
         auto msg = "TShardsVar: Error during blocks cleanup: " + std::string(errors.message);
         SH_CORE::log(msg.c_str());
       }
-      blk->destroy(blk);
+      decRef(blk);
```
sinkingsugar, line 468: Not super happy about this but I guess...

rust/src/types.rs
```diff
@@ -3292,7 +3292,15 @@ impl ShardsVar {
...
-      shard.destroy();
+
+      let mut tmp = SHVar {
+        valueType: SHType_ShardRef,
+        payload: SHVarPayload {
+          __bindgen_anon_1: SHVarPayload__bindgen_ty_1 { shardValue: shard.0 },
+        },
+        ..Default::default()
+      };
+      destroyVar(&mut tmp);
```
sinkingsugar, line 3303: The more I look at this PR the more I'm scared haha.. really ShardsVar should only be used for introspection

src/core/runtime.hpp
```diff
@@ -747,12 +746,8 @@ struct Serialization {
         read((uint8_t *)&klen, sizeof(uint32_t));
         keyBuf.resize(klen);
         read((uint8_t *)keyBuf.c_str(), klen);
-        // TODO improve this, avoid allocations
-        SHVar tmp{};
-        deserialize(read, tmp);
         auto &dst = (*map)[keyBuf];
-        dst = tmp;
-        varFree(tmp);
+        deserialize(read, dst);
```
sinkingsugar, line 750: Leak is likely still there though

src/core/runtime.hpp
```diff
@@ -1259,6 +1257,14 @@ template <typename T> struct WireDoppelgangerPool {
     _wireStr = stream.str();
   }
 
+  ~WireDoppelgangerPool() {
+    SHLOG_TRACE("===== Destroying the pool");
+    SHLOG_TRACE("===== Destroying the pool");
+    SHLOG_TRACE("===== Destroying the pool");
+    SHLOG_TRACE("===== Destroying the pool");
+    SHLOG_TRACE("===== Destroying the pool");
```
sinkingsugar, line 1265: Remove

sinkingsugar (changes requested):

include/shardwrapper.hpp
guusw: Are you sure? shardwrapper doesn't seem to initialize the header struct?

rust/src/types.rs
guusw: Yeah it doesn't destroy anymore now (I also didn't expect to find destroy there) just decreases the reference count

include/utility.hpp
guusw: I wanted to have the same behavior everywhere to avoid issues in the feature.
So it's either all or no reference counting

src/core/runtime.hpp
guusw: 😅

include/shardwrapper.hpp
guusw: Removed it does

src/core/runtime.cpp
```diff
@@ -3139,4 +3072,21 @@ SHCore *__cdecl shardsInterface(uint32_t abi_version) {
...
+  if (shard->refCount == 0) {
+    SHLOG_TRACE("DecRef 0 shard {:x} {}", (size_t)shard, shard->name(shard));
+    shard->destroy(shard);
+  }
+}
+
+void incRef(ShardPtr shard) {
+  if (shard->refCount == 0)
+    SHLOG_TRACE("IncRef 0 shard {:x} {}", (size_t)shard, shard->name(shard));
+  ++shard->refCount;
```
Kryptos-FR, line 3090: Shouldn't this be atomic? Or do we guarantee that ShardRef are never shared between threads?

src/core/shards/wires.cpp
```diff
@@ -1549,6 +1549,8 @@ struct Spawn : public CapturingSpawners {
     capturing = true;
     passthrough = false;
   }
+  ~Spawn() {
+  }
```
Kryptos-FR, line 1553: Is it necessary?

I read on a thread that an explicit empty destructor can prevent inlining.

src/core/runtime.cpp
guusw: Hmm yeah good point, it should be

src/core/shards/wires.cpp
guusw: Left from debugging, will remove

sinkingsugar (approved):

PR #557, opened 2023-02-09
Update boost

guusw:
- Update boost


guusw:
I had to include a manual fix in [stacktrace](https://github.com/shards-lang/boost-stacktrace/tree/fix-uuids) anyway to ignore the duplicate symbols

PR #558, opened 2023-02-09
Working with data docs

ChelseaChX:
Added the segment on Scopes, Flow Methods, Defined Constants, Passthrough and Specific Data Types.


docs/docs/learn/shards/working-with-data.md
```diff
@@ -1,11 +1,236 @@
...
\ No newline at end of file
+### Local and Global
+
+Scope determines the visibility of data at different points in your program. When data is assigned to variables, these variables will either have a **local** or **global** scope.
+
+- Local: The variable is only known within the Wire it originated from.
+
+- Global: The variable is known throughout the entire Mesh.
+
+In the example below, the Wire `retrieve-x` attempts to retrieve the value of `.x` defined in the Wire `define-x`. Note how it is able to retrieve the value of 1 even though it was defined in a separate Wire. This is due to how `.x` has been defined as a global variable, making its value available to all Wires on the Mesh.
```
Kryptos-FR, line 15: Referencing `retrieve-x` but called `get-x` in the code sample.

```suggestion
In the example below, the Wire `get-x` attempts to retrieve the value of `.x` defined in the Wire `define-x`. Note how it is able to retrieve the value of 1 even though it was defined in a separate Wire. This is due to how `.x` has been defined as a global variable, making its value available to all Wires on the Mesh.
```

docs/docs/learn/shards/working-with-data.md
```diff
@@ -1,11 +1,236 @@
...
+### Flow Methods
+
+If Wire Y is run from a separate Wire X using methods such as [`Do`](../../../reference/shards/General/Do), the variables on Wire X will be copied such that Y has access to its value at the moment it was called.
+
+However, this does not mean that Wire Y is in the same scope as X. Wire Y holds only a copy of the value - it does not have access to the actual variable.
+
+If a method such as [`Step`](../../../reference/shards/General/Step) is used instead, Wire Y would be scheduled on Wire X itself, giving it the same scope and access to X's actual variables.
+
+![Methods to control the Flow of Shards can affect scope.](assets/wires-scope.png)
+
```
Kryptos-FR, line 72: Would be good to also include the code representing this situation (on top of the image, or below it).

docs/docs/learn/shards/working-with-data.md
```diff
@@ -1,11 +1,236 @@
...
+=== "Output"
+    ```{.clojure .annotate linenums="1"}
+    [get-x] x: 0
+    ```
+
+### Flow Methods
+
+If Wire Y is run from a separate Wire X using methods such as [`Do`](../../../reference/shards/General/Do), the variables on Wire X will be copied such that Y has access to its value at the moment it was called.
+
+However, this does not mean that Wire Y is in the same scope as X. Wire Y holds only a copy of the value - it does not have access to the actual variable.
```
Kryptos-FR, line 67: This doesn't seem to be the case. `Do` runs inline so I expect it to not only access the variable but modify it.
```clj
(defmesh main)

(defwire wire-y
  1 > .x)

(defwire wire-x
  0 >= .x
  (Log "x") ;; 0
  
  (Do wire-y)
  
  .x
  (Log "x") ;; 1
)

(schedule main wire-x)
(run main)
```

The previous page (https://docs.fragcolor.xyz/learn/shards/the-flow/#summary) also states that `Do` doesn't use the original variables. I guess we missed it during the review. Or is it a bug? cc @sinkingsugar

This would be different if `wire-y` was declared as pure:

```clj
(defmesh main)

(defpure wire-y
  1 >= .x)

(defwire wire-x
  0 >= .x
  (Log "x") ;; 0
  
  (Do wire-y)
  
  .x
  (Log "x")) ;; 0

(schedule main wire-x)
(run main)
```

docs/docs/learn/shards/working-with-data.md
```diff
@@ -1,11 +1,236 @@
...
+
+!!! note
+    Note that the constant `x` defined in the example above is named differently from variables, which would have been `.x` in this case.
+
+## Passthrough
+
+Passthrough determines if data can pass through a Wire unaltered. It allows you to better control the state of the data moving through your program.
+
+Wires will have passthrough disabled by default. In order to enable passthrough for Wires, we employ the shard [`Sub`](../../../reference/shards/General/Sub/). Any shards passed into the `Shards` parameter of `Sub` will run as per usual, except that the final output will be replaced with the initial input passed into `Sub`, thereby creating a passthrough effect.
+
```
Kryptos-FR, line 143: It is not really the wire that has "passthrough" disabled, but more that most shards consume the data and produce another value. So the only way to "reuse" the same value multiple times is to either save it to a variable or to use `Sub` (or `|`). I don't think we should talk about "wire" in this context. Letting @sinkingsugar and @guusw give their opinion.

On that "reuse" note, I would like to have another equivalent sample to show that `Sub` allows to "save" a temporary variable.

```
=== "Example with temporary variable"
    ```{.clojure .annotate linenums="1"}
    (defmesh main)

    (defwire temp-test
      1 (Log "Before")
      >= .temp
      (Math.Add 2) >= .x
      .x (Log "In")
      .temp
      (Log "After"))
    (schedule main  temp-test)
    (run main)
    ```
```

docs/docs/learn/shards/working-with-data.md
```diff
@@ -1,11 +1,236 @@
...
+=== "Output"
+    ```
+    [sub-test] Before Sub: 1
+    [sub-test] In Sub: 3
+    [sub-test] After Sub: 1
+    ```
+
+In the example below, John attempts to find out the price of an apple in different currencies. The base price of 1 USD is passed into the Wire, and goes through a series of shards that each performs mathematical operations on it to obtain its foreign value.
+
+In order to keep the initial value unchanged as it passes through Wires, passthrough has to be enabled for each Wire.
```
Kryptos-FR, line 185: Need to see the corresponding code.

ChelseaChX: I think I will leave the code out for this example. The actual code is a little too verbose and might not make much sense (especially since a string will be created for logging, so the actual output without placing the converted value at the end would be a string instead of a float.

Kryptos-FR (approved):

PR #559, opened 2023-02-10
Various fixes / improvements

guusw:
Various fixes and improvements I made while working on my graphics branch

### DIFF
--- a/include/shards.h
+++ b/include/shards.h
@@ -94,19 +94,22 @@ SH_ARRAY_DECL(SHPayloadArray, struct SHVarPayload);
 struct SHVar;
 SH_ARRAY_DECL(SHSeq, struct SHVar);
 
+struct SHTableImpl;
+struct SHSetImpl;
+
+struct SHTableInterface;
 // 64 bytes should be huge and well enough space for an iterator...
 typedef char SHTableIterator[64];
-struct SHTableInterface;
 struct SHTable {
-  void *opaque;
+  SHTableImpl *opaque;
   struct SHTableInterface *api;
 };
 
+struct SHSetInterface;
 // 64 bytes should be huge and well enough space for an iterator...
 typedef char SHSetIterator[64];
-struct SHSetInterface;
 struct SHSet {
-  void *opaque;
+  SHSetImpl *opaque;
   struct SHSetInterface *api;
 };
 
--- a/include/shards.h
+++ b/include/shards.h
@@ -314,6 +314,27 @@ struct SHSetInterface {
 #define SH_UNION_NAME(_name_)
 #endif
 
+typedef struct SHTableTypeInfo {
+  // If tableKeys is populated, it is expected that
+  // tableTypes will be populated as well and that at the same
+  // key index there is the key's type
+  SHStrings keys;
+  // If tableKeys is not populated, len == 0 and tableKeys is populated len
+  // > 0 it is assumed that tableTypes contains a sequence with the possible
+  // types in the table
+  SHTypesInfo types;
+} SHTableTypeInfo;
+
+typedef struct SHObjectTypeInfo {
+  int32_t vendorId;
+  int32_t typeId;
+} SHObjectTypeInfo;
+
+typedef struct SHEnumTypeInfo {
+  int32_t vendorId;
+  int32_t typeId;
+} SHEnumTypeInfo;
+
 struct SHTypeInfo {
 #if defined(__cplusplus) || defined(SH_USE_ENUMS)
   SH_ENUM_DECL SHType basicType;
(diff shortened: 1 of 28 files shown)


include/shards.hpp
```diff
@@ -682,6 +686,10 @@ struct Var : public SHVar {
     return res;
   }
 
+  template <typename T> static Var Enum(T valuePtr, const Type& type) {
+   return Object(valuePtr, SHTypeInfo(type).enumeration.vendorId, SHTypeInfo(type).enumeration.typeId);
```
Kryptos-FR, line 690: Should be Enum not Object

rust/Cargo.toml
```diff
@@ -7,7 +7,7 @@ authors = ["Giovanni Petrantoni <sinkingsugar@gmail.com>"]
 edition = "2021"
 
 [lib]
-crate-type = ["rlib", "staticlib"]
+crate-type = ["rlib", "staticlib"]
```
Kryptos-FR, line 10: Looks like a parasite change: not sure why there should be a whitespace at the end.

rust/Cargo.toml
```diff
@@ -20,7 +20,7 @@ libsecp256k1 = { version = "0.7.0", optional = true }
 hex = { version = "0.4.2", optional = true }
 crossbeam = { version = "0.8.1", optional = true }
 rapier3d = { version = "0.16.1", optional = true }
-bs58 = { version = "0.4.0", optional = true }
+bs58 = { version = "0.4.0", optional = true }
```
Kryptos-FR, line 23: Looks like a parasite change: not sure why there should be a whitespace at the end.

rust/build.rs
```diff
@@ -11,7 +11,7 @@ fn generate_shardsc() {
 
   let shards_dir = var("SHARDS_DIR").unwrap_or_else(|_| "..".to_string());
   let shards_include_dir = format!("{}/include", shards_dir);
-  let src_extra_dir = format!("{}/src/extra", shards_dir);
+  let src_extra_dir = format!("{}/src/extra", shards_dir);
```
Kryptos-FR, line 14: Looks like a parasite change: not sure why there should be a whitespace at the end.

src/core/shards/core.hpp
```diff
@@ -3211,7 +3211,7 @@ struct ForRangeShard {
 
     SHVar output{};
     SHVar item{};
-    if (from < to) {
+    if (from <= to) {
```
Kryptos-FR, line 3214: Good catch

Kryptos-FR (changes requested):
Overall LGTM.

There is a mistake around Object and Enum and a few unnecessary whitespace changes.
Also please make sure to run the format script once before merging.

sinkingsugar (approved):

include/shards.hpp
guusw: 😱

include/shards.h
```diff
@@ -94,19 +94,22 @@ SH_ARRAY_DECL(SHPayloadArray, struct SHVarPayload);
 struct SHVar;
 SH_ARRAY_DECL(SHSeq, struct SHVar);
 
+struct SHTableImpl;
+struct SHSetImpl;
```
Kryptos-FR, line 98: What's the difference between doing this and `typedef void *SHTableImpl`.
Does C++ offers more guarantees in the former case?

For instance we also have `typedef void *SHPointer` that as far as I can tell is only used for `SHObject`. So should we have `struct SHObjectImpl` instead?

src/core/foundation.hpp
```diff
@@ -1339,12 +1343,23 @@ struct VariableResolver {
 
 template <typename T> T *varAsObjectChecked(const SHVar &var, const shards::Type &type) {
   SHTypeInfo typeInfo(type);
-  if (var.valueType != SHType::Object)
+  if (var.valueType != SHType::Object) {
+    auto err = fmt::format("Invalid type, expected: {} got: {}", type, magic_enum::enum_name(var.valueType));
+    assert(false);
```
Kryptos-FR, line 1348: Any reason for this? Debuggers can also catch exception when they are thrown.

Kryptos-FR (approved):
LGTM. Just a few questions.

include/shards.hpp
Kryptos-FR: Still not fixed.

src/core/foundation.hpp
guusw: I wanted to catch it every time, since it shouldn't happen if the program is written correctly.

include/shards.h
guusw: Because I'm actually implementing these inside foundation.hpp they need to be forward declared as structs

guusw: And the debugger will show the final definitions which are just std's maps and sets

src/core/foundation.hpp
sinkingsugar: `SHLOG_FATAL` will crash the program via abort, should use that

PR #565, opened 2023-02-13
Graphics improvements

guusw:
# Graphics improvements

PR contains several changes I made while working on a simple PBR demo, porting over the shaders from the bgfx days. 

For now, the demo code can be found here: https://github.com/shards-lang/pbr-demo (run entry.edn)

## Generators

Adds the concept of generators:
```clojure
(GFX.Feature :ViewGenerators (->
  (GFX.Texture ...) >= .tex
  (GFX.RenderInto :Textures {:color .tex})
  {:myTexture .tex}))
```

Which when added, will run the generator for each unique `View` being rendered, and the resulting table will be injected as shader paramters.

To precompute once and reuse forever you can wrap your code inside `(Setup)` inside the generator

Note that this also replaces DrawData previously available on Feature, so something like the following can be used to replace the previous DrawData usages:

```clojure
(GFX.Feature :ViewGenerators (->
   {:time 3.0 :baseColor (Float4 1.0)}))
```

If you need to pass data to the generator, channels should be used.

## Render target creation

Renderable textures can now be created using the same Texture shard, but passing it a Format and Size.
```clojure
(GFX.Texture :Format TextureFormat.RGBA8UnormSrgb :Resolution (Int2 1024) :Dimension TextureDimension.D2)
(GFX.Texture :Format TextureFormat.RGBA8UnormSrgb :Resolution .size :MipLevels .num-mips :Dimension TextureDimension.Cube)
```

In addition parameters are supported for most fields and Cubemap and mip levels were added.

## RenderInto

A dedicated shard was added to push a collection of render attachments onto the view stack

```clojure
(GFX.Texture) >= .cube-tex
(GFX.Texture) >= .velocity-tex
(GFX.RenderInto
  :Textures {
         :color {:Texture cube-tex :Face n :Mip mip-index}
         :velocity .velocity-tex }
  :Contents (-> (GFX.Render :Steps steps :View .view))))
```

Attachments can be a whole texture, a specific mip or cubemap (array) layer. The attachments will bind to the named outputs specified on any of the render steps.

This also replaces the previous UI rendering `(GFX.Region)` shards, now you'd use the RenderInto shard for this as well, but pass the :WindowRegion argument as well


src/core/runtime.cpp
```diff
@@ -910,9 +910,10 @@ bool matchTypes(const SHTypeInfo &inputType, const SHTypeInfo &receiverType, boo
             auto bkey = receiverType.table.keys.elements[y];
             // if anyLast and the last perform a match anyway
             if (strcmp(akey, bkey) == 0 || (anyLast && y == (bkeys - 1))) {
-              if (matchTypes(atype, btype, isParameter, strict))
+              if (matchTypes(atype, btype, isParameter, strict)) {
                 missingMatches--;
-              else
+                y = bkeys; // break
```
sinkingsugar, line 915: Add more docs for this as from time to time we get confused

src/core/runtime.cpp
```diff
@@ -2673,6 +2674,8 @@ void SHWire::cleanup(bool force) {
       }
     }
 
+    dispatcher.trigger(SHWire::OnStopEvent{this});
```
sinkingsugar, line 2677: This is not correct, the event is on stop but happens on cleanup

src/core/runtime.hpp
```diff
@@ -237,8 +237,6 @@ inline bool stop(SHWire *wire, SHVar *result = nullptr) {
 
   SHLOG_TRACE("stopping wire: {}", wire->name);
 
-  wire->dispatcher.trigger(SHWire::OnStopEvent{wire});
```
sinkingsugar, line 240: Might need to have two of those, one in the coroutines path the other in the inline wire call path

src/extra/gfx/drawable_utils.hpp
```diff
@@ -2,56 +2,57 @@
...
-  };
+inline TexturePtr varToTexture(const SHVar &var) {
+  if (var.payload.objectTypeId == Types::TextureCubeTypeId) {
+    return *varAsObjectChecked<TexturePtr>(var, Types::TextureCube);
+  } else if (var.payload.objectTypeId == Types::TextureTypeId) {
+    return *varAsObjectChecked<TexturePtr>(var, Types::Texture);
+  } else {
+    SHInstanceData data{};
+    auto varType = shards::deriveTypeInfo(var, data);
+    DEFER({ shards::freeDerivedInfo(varType); });
```
sinkingsugar, line 29: Why defer?
Ah error message

sinkingsugar (changes requested):

src/core/runtime.cpp
guusw: Yeah I guess I can add a second one, I needed a correct place to clean up variables I manually inserted into the wire

sinkingsugar: Add an on cleanup event I guess

sinkingsugar: Also I know the issue, stop often doesn't get called at all!!

sinkingsugar: For now add on cleanup new event. I'll fix the stop thing.

sinkingsugar: Fixed in Tracy PR, can remove this. Hopefully works for you.

src/extra/gfx/drawable_utils.hpp
guusw: Yeah I want to display the type in the error

src/core/runtime.cpp
guusw: Will do

src/core/runtime.cpp
guusw: Added a dedicated OnCleanup event for this

src/core/runtime.hpp
guusw: see last comment

src/gfx/drawable_processors/mesh_drawable_processor.hpp
```diff
@@ -380,14 +299,20 @@ struct MeshDrawableProcessor final : public IDrawableProcessor {
...
         auto &drawableData = drawableDatas->emplace_back();
-        generateDrawableData(drawableData, context.context, cachedPipeline, drawable, context.viewData, context.frameCounter,
-                             needProjectedDepth);
+
+        ParameterStorage *baseDrawData{};
+        if (context.generatorData.drawParameters) {
+          baseDrawData = &(*context.generatorData.drawParameters)[index];
+        }
+
+        // TODO: Temporary to store per-view textures in per-object bind group
```
sinkingsugar, line 308: still todo?

src/gfx/drawable_processors/mesh_drawable_processor.hpp
```diff
@@ -438,21 +363,24 @@ struct MeshDrawableProcessor final : public IDrawableProcessor {
...
 
-      // Collect dynamic view parameters
-      collectGeneratedViewParameters(
-          FeatureCallbackContext{
-              .context = context.context,
-              .view = &context.viewData.view,
-              .cachedView = &context.viewData.cachedView,
-          },
-          cachedPipeline, viewParameters);
+      // Merge generator parameters
```
sinkingsugar, line 370: remove?

src/gfx/pmr/set.hpp
```diff
@@ -2,11 +2,12 @@
 #define F8A69182_378F_46E3_9942_2FEBE21C2FC4
 
 #include "wrapper.hpp"
+#include <functional>
 #include <set>
 
 #if !HAVE_CXX_17_MEMORY_RESOURCE
 namespace shards::pmr {
-template <typename K, typename Less> using set = std::set<K, Less, PolymorphicAllocator<K>>;
+template <typename K, typename Less = std::less<K>> using set = std::set<K, Less, PolymorphicAllocator<K>>;
```
sinkingsugar, line 10: can't use unordered set? this is slower

sinkingsugar (changes requested):

src/gfx/drawable_processors/mesh_drawable_processor.hpp
guusw: Yes, although it only has a slight performance/memory impact since per-view texture bindings are duplicated across object currently.
I plan to address it later. See => #580

src/extra/gfx/main_window.cpp
```diff
@@ -20,6 +22,10 @@ struct MainWindow final {
        SHCCSTR("If the device backing the window should be created with "
                "debug layers on."),
        {CoreInfo::BoolType}},
+      {"IgnoreCompilationErrors",
+       SHCCSTR("When enabled, shader or pipeline compilation errors will be ignored and either use fallback rendering or not "
```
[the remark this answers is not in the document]
guusw: It's harder to do since it happens deeper within `GFX.Render`

src/gfx/drawable_processors/mesh_drawable_processor.hpp
guusw: Nothing to remove

src/gfx/pmr/set.hpp
guusw: Changed to unordered (pmr::maps too)

sinkingsugar (approved):

PR #567, opened 2023-02-14
Add hack to treat Render outputs as cleared/written

guusw:
This allows having multiple GFX.Render calls overlay on the same color buffer for now without having to merge the render graphs

### DIFF
--- a/src/extra/gfx.cpp
+++ b/src/extra/gfx.cpp
@@ -30,6 +30,9 @@ struct RenderShard {
       {"Views",
        SHCCSTR("The views to render into. (Optional)"),
        {CoreInfo::NoneType, Type::VariableOf(Types::ViewSeq), Types::ViewSeq}},
+      {"ClearedHint",
+       SHCCSTR("Hints about already cleared views. Unsupported, do not use!"),
+       {CoreInfo::NoneType, Type::SeqOf(CoreInfo::StringType)}},
   };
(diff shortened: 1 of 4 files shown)


src/tests/gfx-cube.edn
```diff
@@ -50,7 +50,23 @@
     .time (Math.Multiply 0.9) (Math.AxisAngleZ) (Math.Rotation) >= .rotZ
     .rotX (Math.MatMul .rotY) (Math.MatMul .rotZ) > .transform
     .drawable (GFX.Draw)
-    (GFX.Render :Steps .render-steps :View .view))))
```
guusw, line 53: Have to revert this. It was just for testing

src/extra/gfx.cpp
```diff
@@ -30,13 +30,17 @@ struct RenderShard {
       {"Views",
        SHCCSTR("The views to render into. (Optional)"),
        {CoreInfo::NoneType, Type::VariableOf(Types::ViewSeq), Types::ViewSeq}},
+      {"ClearedHint",
+       SHCCSTR("Hints about already cleared views. Unsupported, do not use!"),
+       {CoreInfo::NoneType, Type::SeqOf(CoreInfo::StringType)}},
```
Kryptos-FR, line 35: Is this still accurate ("unsupported")?

PR #568, opened 2023-02-14
fix hashing and some other stuff

sinkingsugar:


src/core/runtime.cpp
```diff
@@ -625,16 +625,27 @@ SHWireState suspend(SHContext *context, double seconds) {
...
-  thread_local std::unordered_set<const SHWire *> wires;
-  return wires;
-#endif
-}
+#define SH_HASHER(prefix)                                                               \
+  std::deque<std::unordered_set<const SHWire *>> &prefix##WiresStack() {                \
+    thread_local std::deque<std::unordered_set<const SHWire *>> s;                      \
+    return s;                                                                           \
+  }                                                                                     \
+  std::optional<std::unordered_set<const SHWire *> *> &prefix##Wires() {                \
```
guusw, line 633: Might as well make this return a reference to the set directly (and assert internally that it's not null)
Especially since the usage below implies doing the same

src/core/runtime.cpp
```diff
@@ -625,16 +625,27 @@ SHWireState suspend(SHContext *context, double seconds) {
...
-#ifdef WIN32
-  // we have to leak.. or windows tls emulation will crash at process end
-  thread_local std::unordered_set<const SHWire *> *wires = new std::unordered_set<const SHWire *>();
-  return *wires;
-#else
-  thread_local std::unordered_set<const SHWire *> wires;
-  return wires;
-#endif
-}
+#define SH_HASHER(prefix)                                                               \
```
guusw, line 628: Doesn't seem that hash specific, maybe just call the macro `SH_WIRE_SET_STACK`

src/mal/SHCore.cpp
```diff
@@ -2488,6 +2488,33 @@ SHARDS_API __cdecl SHBool shLispEval(void *env, const char *str, SHVar *output)
     return false;
   }
 }
+
+SHARDS_API __cdecl SHBool shLispAddToEnv(void *env, const char *str, SHVar input, bool reference) {
```
guusw, line 2492: Can you add the declaration to `SHLisp.h` too

guusw (changes requested):

src/mal/SHCore.cpp
sinkingsugar: Didn't know we had that file :D

PR #575, opened 2023-02-18
add support for fibers in tracy

sinkingsugar:
Fibers tracing works fine unless mixed with threads. Might not be a big deal anyway as tracking wires is enough.

### DIFF
--- a/src/core/brancher.hpp
+++ b/src/core/brancher.hpp
@@ -156,8 +156,15 @@ public:
     }
   }
 
-  void activate() {
+  void activate(SHContext *context) {
+#ifdef TRACY_FIBERS
+    auto wire = context->wireStack.back();
+    TracyCoroExit(wire);
+#endif
     if (!mesh->tick()) {
+#ifdef TRACY_FIBERS
+      TracyCoroEnter(wire);
+#endif
       switch (failureBehavior) {
       case BranchFailureBehavior::Ignore:
         break;
@@ -172,6 +179,10 @@ public:
         }
         break;
       }
+    } else {
+#ifdef TRACY_FIBERS
+      TracyCoroEnter(wire);
+#endif
     }
   }
 
--- a/src/core/brancher.hpp
+++ b/src/core/brancher.hpp
@@ -157,14 +157,7 @@ public:
   }
 
   void activate(SHContext *context) {
-#ifdef TRACY_FIBERS
-    auto wire = context->wireStack.back();
-    TracyCoroExit(wire);
-#endif
     if (!mesh->tick()) {
-#ifdef TRACY_FIBERS
-      TracyCoroEnter(wire);
-#endif
       switch (failureBehavior) {
       case BranchFailureBehavior::Ignore:
         break;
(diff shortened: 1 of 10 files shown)


sinkingsugar:
@guusw this adds the correct on stop event. Merge if good 👍

src/core/runtime.hpp
```diff
@@ -41,6 +41,15 @@ using SHTimeDiff = decltype(SHClock::now() - SHDuration(0.0));
 #include <emscripten/val.h>
 #endif
 
+#ifdef TRACY_ENABLE
```
guusw, line 44: You don't need the #if since tracy already defines them to nothing if disabled

src/core/runtime.hpp
```diff
@@ -301,8 +298,9 @@ inline bool tick(SHWire *wire, SHDuration now) {
     wire->coro->resume();
 #endif
 
-#ifdef SH_USE_TSAN
-    __tsan_switch_to_fiber(curr, 0);
+#ifdef TRACY_FIBERS
+    TracyFiberLeave;
```
guusw, line 302: This isn't correct I think, if you're switching back to another wire you should call Enter but with it's name instead (Try Step/Do/etc., the profiler just says invalid zone, left twice, etc.)

Tracy doesn't keep a stack. Probably the easiest way to fix this is to pass an (optional) current context to switching functions (tick, prepare) so you can switch back to the calling 'fiber' after returning. Or keep a thread_local stack somewhere (just for the profiler)

guusw (changes requested):

src/core/runtime.hpp
sinkingsugar: kinda need in this file due to Valgrind

sinkingsugar: also they don't work well with threads so need to disable them by default (fibers)

src/core/runtime.hpp
sinkingsugar: hmm, not sure if that is an issue, should not be the case

sinkingsugar: But maybe this is exactly what's happening with threads related shards... 💡

sinkingsugar: Threadlocal wont help sadly. Flow might. I'll think about it.

sinkingsugar: Thing is profiler says there never was a thread.. not twice. But might be related indeed. But not sure about the wire thing though. Wire pointer won't ever change cos it's decided when call is called. No matter when we resume.

sinkingsugar:
@guusw Can you check again?
As discussed on discord indeed even if they state in docs it should be ok to nest coros it wasn't... 
I solved with thread local as you suggested but works only because we never move coroutines between threads yet (as they run).

src/core/brancher.hpp
```diff
@@ -156,7 +156,7 @@ struct Brancher {
     }
   }
 
-  void activate() {
+  void activate(SHContext *context) {
```
sinkingsugar, line 159: We could remove this as I don't use anymore but I think it's better to have for future usages. Even to just signal activate runs in context.

guusw: I didn't add it because I was using it in a place where I don't actually have a context in my gfx branch. but I can remove it there.

guusw:
Merged in bc5e1c3

PR #578, opened 2023-02-21
Remove Drawable input tables

guusw:
Changing the API of `GFX.Drawable` shard so they're attached to their state and they should be ran from a loop to update parameters/transform/etc.

- Make loopable, no longer return unique instances
- Remove table input, use parameters
- Merge Params and Textures (this means you can no longer have texture and basic parameters with the same name)
- Added helpers that will iterate over parameters defined with PARAM macros and insert references into requiredVariable()

### DIFF
--- a/src/tests/gfx-cube.edn
+++ b/src/tests/gfx-cube.edn
@@ -30,7 +30,6 @@
    0.0 >= .time
    cube (GFX.Mesh :Layout cube-layout :WindingOrder WindingOrder.CW) >= .mesh
    (Float3 0 0 0) (Math.Translation) >= .transform
-   {:Transform .transform :Mesh .mesh} (GFX.Drawable :Transform .transform) >= .drawable
 
     ; Create render steps
    (GFX.BuiltinFeature :Id BuiltinFeatureId.Transform) >> .features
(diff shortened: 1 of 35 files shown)


guusw:
I might actually undo that `Merge Params and Textures` change though

src/tests/gfx-cube.edn
```diff
@@ -49,7 +48,7 @@
     .time (Math.Multiply 0.7) (Math.AxisAngleY) (Math.Rotation) >= .rotY
     .time (Math.Multiply 0.9) (Math.AxisAngleZ) (Math.Rotation) >= .rotZ
     .rotX (Math.MatMul .rotY) (Math.MatMul .rotZ) > .transform
-    .drawable (GFX.Draw)
+    (GFX.Drawable :Transform .transform :Mesh .mesh) (GFX.Draw)
```
sinkingsugar, line 51: hmm to be honest I'd like to have transform as input here
simply cos you can do something very nice like
```clojure
.transform
(GFX.Drawable .home)
(GFX.Drawable .fence)
(GFX.Drawable .chair)
...
```
Very Shards, and very real if pivot is proper

sinkingsugar (changes requested):
`GFX.Drawable` should take a transform as input !

PR #581, opened 2023-02-22
Giovanni/fixes-22-2

sinkingsugar:


rust/src/shards/gui/widgets/combo.rs
```diff
@@ -253,8 +253,12 @@ impl Shard for Combo {
...
       }
 
-      // this is fine because we don't own input and seq is just a view of it in this case
-      Ok(seq[*index])
+      if seq.is_empty() {
+        Ok(Var::default())
+      } else {
+        // this is fine because we don't own input and seq is just a view of it in this case
+        Ok(seq[*index])
+      }
```
Kryptos-FR, line 261: 😱😵‍💫🙏

sinkingsugar: I missed this when I removed cloneVar as well (which also would have crashed) 🤣

PR #582, opened 2023-02-22
File dialog

Kryptos-FR:
Basic implementation.

For now, only works 100% on Windows and macOS.
It should work on Linux for Desktop environments but won't for pure headless.
For wasm, we need to use rust async which I'm not sure how to make it work with our own coroutine.

- [x] filters
- [x] multi-selection
- [ ] ~~fallback UI~~

### DIFF
--- a/docs/samples/shards/FS/FileDialog/1.edn
+++ b/docs/samples/shards/FS/FileDialog/1.edn
@@ -2,6 +2,10 @@
 (FS.FileDialog)
 (Log "file path")
 
+; pick only .edn and .log files
+(FS.FileDialog :Filters ["edn" "log"])
+(Log "file path")
+
 ; pick any folder
 (FS.FileDialog :Folder true)
 (Log "folder path")
--- a/docs/samples/shards/FS/FileDialog/1.edn
+++ b/docs/samples/shards/FS/FileDialog/1.edn
@@ -2,6 +2,10 @@
 (FS.FileDialog)
 (Log "file path")
 
+; pick any file from the parent directory
+(FS.FileDialog :CurrentDir "..")
+(Log "file path")
+
 ; pick only .edn and .log files
 (FS.FileDialog :Filters ["edn" "log"])
 (Log "file path")
--- a/docs/samples/shards/FS/FileDialog/1.edn
+++ b/docs/samples/shards/FS/FileDialog/1.edn
@@ -1,0 +1,7 @@
+; pick any file
+(FS.FileDialog)
+(Log "file path")
+
+; pick any folder
+(FS.FileDialog :Folder true)
+(Log "folder path")
--- a/docs/samples/shards/FS/FileDialog/1.edn
+++ b/docs/samples/shards/FS/FileDialog/1.edn
@@ -6,7 +6,7 @@
 (FS.FileDialog :CurrentDir "..")
 (Log "file path")
 
-; pick only .edn and .log files
+; pick only .edn and .log file
 (FS.FileDialog :Filters ["edn" "log"])
 (Log "file path")
 
--- a/rust/src/shards/fs.rs
+++ b/rust/src/shards/fs.rs
@@ -9,11 +9,14 @@ use crate::types::common_type;
 use crate::types::ClonedVar;
 use crate::types::Context;
 use crate::types::ExposedTypes;
+use crate::types::InstanceData;
 use crate::types::ParamVar;
 use crate::types::Parameters;
 use crate::types::Seq;
+use crate::types::Type;
 use crate::types::Types;
 use crate::types::Var;
+use crate::types::BOOL_TYPES_SLICE;
 use crate::types::BOOL_VAR_OR_NONE_SLICE;
 use crate::types::NONE_TYPES;
 use crate::types::STRING_TYPES;
@@ -25,6 +28,7 @@ lazy_static! {
     common_type::strings_var,
     common_type::none
   ];
+  pub static ref STRING_OR_STRINGS_TYPES: Types = vec![common_type::string, common_type::strings];
   static ref FILEDIALOG_PARAMETERS: Parameters = vec![
     (
       cstr!("Filters"),
@@ -39,6 +43,12 @@ lazy_static! {
     )
       .into(),
     (
+      cstr!("Multiple"),
+      shccstr!("To select multiple files instead of just one."),
+      BOOL_TYPES_SLICE,
+    )
+      .into(),
+    (
       cstr!("CurrentDir"),
       shccstr!("Set the current directory"),
       STRING_VAR_OR_NONE_SLICE,
@@ -50,6 +60,7 @@ lazy_static! {
 struct FileDialog {
   filters: ParamVar,
   folder: ParamVar,
+  multiple: bool,
   current_dir: ParamVar,
   output: ClonedVar,
 }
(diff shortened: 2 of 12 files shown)


sinkingsugar (changes requested):
I don't see it wrapped in awaits/blocking activation utilty.. how does this thing not block the thread/couroutine?

rust/src/shards/gui/widgets/file_dialog.rs
```diff
@@ -0,0 +1,147 @@
...
+    parents.set_name(PARENTS_UI_NAME);
+    Self {
+      parents,
+      requiring: Vec::new(),
+      folder: ParamVar::new(false.into()),
+    }
+  }
+}
+
+impl Shard for FileDialog {
```
guusw, line 42: Why does this require a UI context, should be possible to use without I think

rust/src/shards/gui/widgets/file_dialog.rs
```diff
@@ -0,0 +1,147 @@
...
+    // Note: technically we don't need egui when using rfd, but we still keep the dependency for consistency with the fallback
+    let path = if self.folder.get().try_into()? {
+      rfd::FileDialog::new().pick_folder()
+    } else {
+      rfd::FileDialog::new().pick_file()
+    };
+    let output = if let Some(path) = path {
+      let tmp = Var::ephemeral_string(path.display().to_string().as_str());
+      let mut output = Var::default();
+      cloneVar(&mut output, &tmp);
```
guusw, line 132: Avoid using cloneVar wherever possible, shards shouldn't output cloned vars as this will leak.
If you need storage it should be in one of the rust helper structs like ClonedVar and stored on the shards and return a reference.

guusw (changes requested):

rust/src/shards/gui/widgets/file_dialog.rs
```diff
@@ -0,0 +1,147 @@
...
+    } else {
+      Err("No UI parent")
+    }
+  }
+}
+
+impl FileDialog {
+  // FIXME rfd is supported on Linux but requires linking with GTK3
+  #[cfg(any(target_os = "windows", target_os = "macos"))]
+  fn activate_ui(&mut self, ui: &mut egui::Ui, _context: &Context, _input: &Var) -> Result<Var, &str> {
```
guusw, line 122: Should probably implement:
```rs
   impl BlockingShard for $shard_name {
      fn activate_blocking(&mut self, _context: &Context, input: &Var) -> Result<Var, &str> {
           ...}}
```
to not block the wire, check http.rs for reference

guusw (changes requested):

rust/src/shards/gui/widgets/file_dialog.rs
Kryptos-FR: Two reasons.

1. It is to make sure that it is only used in a context where we are able to draw UI (won't make sense for headless programs).
2. It isn't supported on all platforms (e.g. wasm) so we will need a egui replacement for those in the future.

rust/src/shards/gui/widgets/file_dialog.rs
Kryptos-FR: For string it is the only way as far as I know: the ephemeral string cannot be kept around.

rust/src/shards/gui/widgets/file_dialog.rs
Kryptos-FR: Good point. I will look into that.

rust/src/shards/gui/widgets/file_dialog.rs
sinkingsugar: this is a leak, you need to store output as `ClonedVar` in self and stop using `cloneVar`

rust/src/shards/gui/widgets/file_dialog.rs
Kryptos-FR: But I guess I can remove that constraint for now.

sinkingsugar: I think makes sense for a headless program tbh, a shell program could use a picker, if UI is not needed don't need it.

rust/src/shards/gui/widgets/file_dialog.rs
guusw: The thing is it also needs to be releaseVar'd, but ClonedVar takes care of that since it implements drop.
By convention shards return values should be owned by the shard returning it, so store the result in a ClonedVar member (might need option<> around it) and return that converted to SHVar

Kryptos-FR:
Looks like we can't test it on the macOS machine in CI.
> thread '<unnamed>' panicked at 'You are running RFD in NonWindowed environment, it is impossible to spawn dialog from thread different than main in this env.', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rfd-0.11.1/src/backend/macos/utils.rs:39:13
https://github.com/fragcolor-xyz/shards/actions/runs/4251745822/jobs/7394972737#step:6:5127

Kryptos-FR:
I guess I have to exclude `FileDialog` from the tests. Since it requires interaction, it can't be tested properly with our current setup.

rust/src/shards/fs.rs
```diff
@@ -0,0 +1,162 @@
...
+      dialog.pick_file()
+    };
+    if let Some(path) = path {
+      let path = path.display().to_string();
+      let output = Var::ephemeral_string(path.as_str());
+      self.output = output.as_ref().into();
+      Ok(self.output.0)
+    } else {
+      // FIXME should we return an empty string instead?
+      Ok(Var::default())
```
guusw, line 150: It should definitely be a string if that's specified as output type.
I'm thinking maybe even throw if the user cancels, that way we can use it with (Maybe) to handle that case

guusw: However not sure how to do that from rust, do we have anything for that @sinkingsugar ?

sinkingsugar: err... `Err("error")` 🤣

Kryptos-FR: But is it really an error? I was thinking that returning None or an empty string was less intrusive...

sinkingsugar: It is an error because the actual flow of the app won't be as expected. And should be able by another alternative path driven by Maybe.

guusw (approved):
Looks good

rust/src/shards/fs.rs
```diff
@@ -135,6 +148,13 @@ impl BlockingShard for FileDialog {
         dialog = dialog.add_filter(filter, &[filter]);
       }
     }
+    let cd = self.cd.get();
```
sinkingsugar, line 151: what's the default? process current?

docs/samples/shards/FS/FileDialog/1.edn
```diff
@@ -0,0 +1,15 @@
+; pick any file
+(FS.FileDialog)
+(Log "file path")
+
+; pick any file from the parent directory
+(FS.FileDialog :CD "..")
```
sinkingsugar, line 6: Can we be verbose here ? `CurrentDir` at least? this will be very visual soon. CD makes little sense to new users

sinkingsugar (changes requested):

sinkingsugar:
looks good, minor stuff

rust/src/shards/fs.rs
Kryptos-FR: It seems to depend on the OS.
On Windows it remembers latest location. On my Ubuntu it goes to "recent locations". It's likely driven by the OS itself.

sinkingsugar (approved):

PR #583, opened 2023-02-23
Guus/tracy fixes

guusw:


.github/workflows/build-windows.yml
```diff
@@ -177,7 +177,7 @@ jobs:
           ./bootstrap
           mkdir build
           cd build
-          cmake -G Ninja -DCMAKE_BUILD_TYPE=${{ steps.setup.outputs.build-type }} ..
+          cmake -G Ninja -DCMAKE_BUILD_TYPE=${{ steps.setup.outputs.build-type }} -DTRACY_ENABLE=ON -DTRACY_TIMER_FALLBACK=ON ..
```
Kryptos-FR, line 180: Could you include in this PR to not clear the ` target`  folder?
Also, I think we don't need to build libbacktrace anymore, do we?

guusw: nope, didn't seem used at least

PR #587, opened 2023-02-24
Custom listbox with user template

Kryptos-FR:
By default list box will display a sequence of strings.
If the sequence is not of strings, a template must be provided for displaying.

```clj
[1 2 3 4]
(UI.ListBox
 :Index .index
 :Template (-> (ToString) (UI.Label))
>= .selected-value
```


rust/src/shards/gui/widgets/listbox.rs
```diff
@@ -131,7 +144,26 @@ impl Shard for ListBox {
...
     }
 
-    Ok(common_type::any)
+    let input_type = data.inputType;
+    let slice = unsafe {
+      let ptr = input_type.details.seqTypes.elements;
+      std::slice::from_raw_parts(ptr, input_type.details.seqTypes.len as usize)
+    };
+    let element_type = if !slice.is_empty() {
+      slice[0]
```
sinkingsugar, line 153: this is not safe, you should check this way:
If len is 1 ok to use slice[0]
If len is >1 use any cos we cannot be sure

sinkingsugar (changes requested):

rust/src/shards/gui/widgets/listbox.rs
Kryptos-FR: I could also check all types if `len > 1` and return ` any` only if two types are different.

sinkingsugar: Type info for seq should not include duplicates, if it does we do something wrong 😅 you can assert that if you want

rust/src/shards/gui/widgets/listbox.rs
```diff
@@ -131,7 +144,33 @@ impl Shard for ListBox {
...
+    let slice = unsafe {
+      let ptr = input_type.details.seqTypes.elements;
+      std::slice::from_raw_parts(ptr, input_type.details.seqTypes.len as usize)
+    };
+
+    let element_type = match slice.len() {
+      0 => common_type::none,
+      1 => slice[0],
+      _ => {
+        if slice.iter().skip(1).all(|t| *t == slice[0]) {
```
sinkingsugar, line 157: As said on other comment while this is correct it would not be correct to have multiple seqTypes of same type. So in theory this should not be needed. We should verify correctness at some point (until I worked alone I tried to apply this). But anyway out of this PR scope.

sinkingsugar (approved):

PR #588, opened 2023-02-25
Giovanni/25 2 2023

sinkingsugar:


rust/src/shards/gui/layouts/table.rs
```diff
@@ -404,18 +428,17 @@ impl Shard for Table {
               }
             }
           })
-        } else {
-          // fallback to default columns with no header
-          for _ in 0..seq.len() {
-            builder = builder.column(Column::remainder());
-          }
-          builder.header(0.0, |_| {})
```
Kryptos-FR, line 412: It is still necessary to do something if `:Columns` isn't set. Otherwise it will have phantom headers that take some vertical space.

Could be simplified though with a condition around the header creation above:
```rust
let row_height = if columns.len() > 0 { 20.0 } else { 0.0 };
builder.header(row_height, |mut header_row| {
// ...
}
```

rust/src/shards/gui/layouts/table.rs
```diff
@@ -347,10 +365,11 @@ impl Shard for Table {
...
           builder = builder.resizable(resizable.try_into()?);
         }
+
         // configure columns
-        let columns = self.columns.get();
-        let table = if !columns.is_none() {
-          let columns: Seq = columns.try_into()?;
+        let table = {
+          let columns: Seq = self.columns.0.as_ref().try_into()?;
+
```
Kryptos-FR, line 372: `:Columns` can be unset, so this condition is necessary.

Kryptos-FR (changes requested):

Kryptos-FR (approved):

PR #592, opened 2023-02-27
Overview sections

ChelseaChX:
- Added an overview segment to reference for Shards.

- Removed focus on Shards and fixed the documentation to encompass other projects
Changelog:

> 1. Changed download/index.md to encompass other projects
> 
> 2. Changed learn/index.md and learn/.pages to encompass other projects
> 2a. Moved "shards" and "tutorials" folders into a "shards" folder under "learn"
> 2b. Renamed "learn/shards/shards" into "learn/shards/primer"
> 2c. Add a new .index and .pages for the new primer folder
> 
> 3. Changed reference/index.md and .pages to encompass other projects
> 3a. Added a new "shards" folder into "reference"


guusw:
Maybe we can add the functions in general: `AppendTo`,`PrependTo` to Strings as well, since they might be useful in that context

guusw:
What happened to the overview sections? I can't see them anymore and they 404 when I build the docs

ChelseaChX:
> What happened to the overview sections? I can't see them anymore and they 404 when I build the docs

It should be under "reference/shards/overview/"

ChelseaChX:
> What happened to the overview sections? I can't see them anymore and they 404 when I build the docs

Found the issue. The file wasn't added in after creating the new directories.

docs/docs/learn/shards/.pages
```diff
@@ -1,6 +1,4 @@
 nav:
   - index.md
-  - what-is-shards.md
-  - coding-with-shards.md
-  - the-flow.md
-  - working-with-data.md
+  - primer
+  - tutorials
```
Kryptos-FR, line 4: If we decide to move the tutorials there, we have to also modify the CI job to have them copied at the right place.

Kryptos-FR:
@ChelseaChX Can you describe all the changes (esp. pages moved) in the description of this PR?

There are bunch of pages missing that are I think still necessary. In addition, moving things around means we have to update the CI jobs otherwise it won't work.

sinkingsugar (commented):
That's a lot of content deleted for very little added.... are we sure?

Kryptos-FR:
> That's a lot of content deleted for very little added.... are we sure?

I think some removed files were supposed to be moved but weren't properly committed.

sinkingsugar:
> > That's a lot of content deleted for very little added.... are we sure?
> 
> 
> 
> I think some removed files were supposed to be moved but weren't properly committed.

Wow 😮 ok... we are moving backwards then 😅 @ChelseaChX can we pay more attention please?
Keep an eye or two on it please @Kryptos-FR

docs/docs/learn/shards/.pages
ChelseaChX: Yes we'll have to shift the tutorials there since our learn pages will be grouped up by projects. I've renamed the primer folder from "shards" to "primer" too since the parent folder would be "shards" too
.

ChelseaChX:
> > That's a lot of content deleted for very little added.... are we sure?
> 
> I think some removed files were supposed to be moved but weren't properly committed.

image

It's quite strange that it doesn't detect all the unversioned files. I'll have to manually add them with TortoiseGit from the File Explorer instead.

Kryptos-FR:
> It's quite strange that it doesn't detect all the unversioned files. I'll have to manually add them with TortoiseGit from the File Explorer instead.

That's because there is a filter to exclude some .md files (those generated by the scripts for the API). We also need to update that in `.gitignore`. You can always display unversioned files by selecting the right option in the drop down (and please don't use TortoiseGit :smile:):
image

Kryptos-FR:
@sinkingsugar Do we have to change all copyright notice in source files as well to Fragnova? I thought Fragcolor was ok there.

Kryptos-FR (approved):

PR #596, opened 2023-02-28
Guus' Daily Fixes

guusw:


src/core/runtime.cpp
```diff
@@ -864,7 +864,7 @@ bool matchTypes(const SHTypeInfo &inputType, const SHTypeInfo &receiverType, boo
         }
         return false;
       } else if (inputType.seqTypes.len == 0 || receiverType.seqTypes.len == 0) {
-        return false;
```
sinkingsugar, line 867: bad feeling about this :)

sinkingsugar: but hmm makes sense to be true I guess...

Kryptos-FR: I think it is consistent with algorithms where `seq.any(/* whatever condition */)` returns `true` for empty sequences.

sinkingsugar: @guusw this is wrong, RECEIVER SHOULD have a type. no types is an anomaly, should at least have ANY. Please restore `false`

sinkingsugar: you can workaround `expect` issues by forcing things to be any with generic `ExpectTable` and `ExpectSeq` but let's avoid breaking delicate behavior

guusw: Yeah didn't realize what this was meant for, will add a comment. undo and work around it for now

sinkingsugar (approved):

PR #599, opened 2023-03-01
DoMany

sinkingsugar:
Not sure why I considered this not possible before... But seems fine.

### DIFF
--- a/src/core/shards/wires.cpp
+++ b/src/core/shards/wires.cpp
@@ -1761,6 +1761,61 @@ struct StepMany : public TryMany {
   }
 };
 
+struct DoMany : public TryMany {
+  static inline Parameters _params{
+      {"Wire", SHCCSTR("The pure wire to run many times sequentially."), WireBase::WireVarTypes},
+  };
+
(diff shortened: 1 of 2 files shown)


src/core/shards/wires.cpp
```diff
@@ -1761,6 +1761,57 @@ struct StepMany : public TryMany {
...
+        return cref->wire->previousOutput; // doesn't matter actually
+      } else {
+        // we don't want to propagate a (Return)
+        if (unlikely(runRes.state == SHRunWireOutputState::Stopped)) {
+          context->continueFlow();
+        }
+      }
+
+      // this can be anything really...
+      cloneVar(_outputs[i], cref->wire->previousOutput);
```
Kryptos-FR, line 1808: ~In `StepMany`, this line is `_outputs[i] = cref->wire->previousOutput`.
Any reason why `cloneVar()` is necessary here but in `StepMany`?~

nvm: it was the fix in #594

Kryptos-FR (approved):
LGTM

PR #601, opened 2023-03-02
Overview sections

ChelseaChX:
- Added a segment on Conversion shards

### DIFF
--- a/docs/docs/reference/shards/overview/index.md
+++ b/docs/docs/reference/shards/overview/index.md
@@ -291,13 +291,16 @@ These shards are used for converting between data types.
 
 For type conversions:
 
-- [ToBytes](../../../reference/shards/General/ToBytes/)
-
 - [ToInt](../../../reference/shards/General/ToInt/)
 
 - [ToFloat](../../../reference/shards/General/ToFloat/)
 
 - [ToString](../../../reference/shards/General/ToString/)
+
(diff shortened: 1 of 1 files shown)


docs/docs/reference/shards/overview/index.md
```diff
@@ -285,4 +285,33 @@ Shards to aid Physics calculations:
 
 - [Physics.CastRay](../../../reference/shards/Physics/CastRay/)
 
+## Conversion
+
+These shards are used for converting between data types.
+
+For type conversions:
+
+- [ToBytes](../../../reference/shards/General/ToBytes/)
```
Kryptos-FR, line 294: So this one is a bit problematic because it is not what it seems. It does convert to bytes but using our serialization mechanism. In other words:
```clj
"Hello shards" (ToBytes) (FromBytes) ; ok
"Hello shards" (ToBytes) (BytesToString) ; ko
"Hello shards" (StringToBytes) (BytesToString) ; ok
```
@sinkingsugar should we rename ` ToBytes` and `FromBytes` to something like `Serde.ToBytes` `Serde.FromBytes` (or `Serialization` but that's a bit long).
Or at the very least we need to write good documentation for those. I think the pages of these shards are quite empty at the moment.

Kryptos-FR (approved):

PR #606, opened 2023-03-05
Guus/global render graph

guusw:
Adds a global queue that all Render commands go into, this way you can have them in different locations but will be combined together during rendering


guusw:
Note to self: Should make the FrameQueue part of the view stack, that way it also works when not rendering to the main output.

sinkingsugar (approved):

PR #609, opened 2023-03-06
CI: use clang to build Windows x64

Kryptos-FR:
**Summary**
Split Windows workflow:
- `build-windows.yml` will build 64-bit version using clang
- `build-windows-32bit.yml` will continue building using MSYS 32-bit

**Tasks**
- [x] update documentation for setting up the env (i.e. "Getting started")
- [x] update documentation for building on Windows (using clang)


docs/docs/contribute/getting-started.md
```diff
@@ -14,109 +14,193 @@ license: CC-BY-SA-4.0
...
 
-!!! note
-    You are free to use any code editor of your choice, although you will have to complete steps of this tutorial through alternate means.
+A compiler translates human-readable code into machine code. We will be setting up the C/C++ compiler so that our code written in the human-readable C++ language can be understood by the computers!
 
-## Setting up the GCC Compiler ##
+=== "Windows"
 
-A compiler translates human-readable code into machine code. We will be setting up the GCC compiler so that our code written in the human-readable C++ language can be understood by the computers!
+    First, [download Chocolatey](https://docs.chocolatey.org/en-us/choco/setup#installing-chocolatey) and install it using an [administrative shell](https://www.howtogeek.com/194041/how-to-open-the-command-prompt-as-administrator-in-windows-8.1/). Chocolatey is a package manager that simplifies the installation of the tools we will need.
```
ChelseaChX, line 24: It will be better to avoid nested instructions (which our previous documentation suffered from) as it redirects the tutorial flow and makes it more complicated than necessary. If possible, we should just state the instructions for using an administrative shell here.

docs/docs/contribute/getting-started.md
```diff
@@ -14,109 +14,193 @@ license: CC-BY-SA-4.0
...
 
-First, [download MSYS2](https://www.msys2.org/) and install it with the default settings. MSYS2 helps to set up MinGW-w64, which uses the GCC compiler to create Windows programs.
+    We will now install the compiler and other build tools using Chocolatey.
+    Open up the Command Prompt and enter the following commands:
 
-We will now install the Mingw-w64 toolset which includes GCC, and other build tools that will be employed later. When the MinGW terminal pops up, enter the following command:
+    ??? tip
```
ChelseaChX, line 29: This tip should appear after the the commands as it breaks the instructional flow from "commands:" to the commands.

docs/docs/contribute/getting-started.md
```diff
@@ -14,109 +14,193 @@ license: CC-BY-SA-4.0
...
-```
+    ```cmd
+    choco install -y cmake --installargs '"ADD_CMAKE_TO_PATH=System"'
+    choco install -y ninja
+    choco install -y llvm
+    refreshenv
+    ```
 
+    !!! note
+        By default, `System` will make `cmake` available to all users on the Windows machine. To only set it for the current user, replace `System` with `User`.
```
ChelseaChX, line 41: It's not really by default as system is specified in the code. 

How about this instead?

```The code above will make `cmake` available to all users on the same machine. To only set it for the current user, replace `System` with `User`.```

File: docs/docs/contribute/getting-started.md [hunk 4]
```diff
@@ -14,109 +14,193 @@ license: CC-BY-SA-4.0
...
 
-Git is a system used for managing and tracking changes in your code. It is also known as a Version Control System (VCS). It makes it easier for collaborators to work together, and allows you to access our projects too!
+    Alternatively, you can also use chocolatey:
+    ```cmd
+    choco install -y git
+    ```
 
-Install Git [here](https://git-scm.com/download). The installation settings can be left unchanged.
+=== "Linux"
+    ```bash
```
ChelseaChX, line 156: You might want an opening line to explain what the command does.

For example:
`Install Git with the following command:`

docs/docs/contribute/getting-started.md
```diff
@@ -14,109 +14,193 @@ license: CC-BY-SA-4.0
...
+    sudo apt install git
+    ```
 
 Although Git is powerful, it can be daunting to use on its own. Instead, we use GitHub, a service which employs Git’s version control with its own to make project collaboration much easier. 
 
 We recommend that you use a [Git GUI Client](https://git-scm.com/downloads/guis) to facilitate your work by pruning off the manual input of code and making the version control process more visual.
 
-We will be using [GitHub Desktop](https://desktop.github.com/) for the tutorials. You can install it with the default settings, and create a GitHub account when prompted if you do not have one.
+=== "Windows"
```
ChelseaChX, line 164: Is there supposed to be Linux tab here too?

docs/docs/contribute/getting-started.md
```diff
@@ -14,109 +14,193 @@ license: CC-BY-SA-4.0
...
+    You can download it [here](https://code.visualstudio.com/download), and install it with the default settings.
+
+    Alternatively, you can also use chocolatey:
+    ```cmd
+    choco install -y vscode
+    ```
+
+=== "Linux"
+
+    The easiest is to install the snap package.
```
ChelseaChX, line 192: Some mention of what the code is meant to do would be good. 

`You can install VS Code with the snap package. Use the command:`

docs/docs/contribute/getting-started.md
```diff
@@ -14,109 +14,193 @@ license: CC-BY-SA-4.0
...
+    ```bash
+    sudo snap install --classic code
+    ```
+
+    Follow the [installation](https://code.visualstudio.com/docs/setup/linux#_installation) page for more details, and other installation alternatives.
```
ChelseaChX, line 197: Might want to remove this link to reduce redirecting the tutorial to external sources. The tutorials should be short and simple. If you plan to add this link for alternate installation options, you might want to place it within a content tab to make it optional. (e.g., using `!!! note`)

docs/docs/contribute/getting-started.md
```diff
@@ -134,57 +218,97 @@ Search for and install the following extensions:
...
-```
-rustup update
-```
+    4. Install the Rust Web Assembly toolchain.
+    ```bash
+    rustup +nightly target add wasm32-unknown-unknown
+    ```
 
-9. Install [Git](https://git-scm.com/download).
+    5. Update Rust
```
ChelseaChX, line 297: Fullstop at the end.

docs/docs/contribute/getting-started.md
```diff
@@ -134,57 +218,97 @@ Search for and install the following extensions:
...
-```
-rustup toolchain install nightly
-```
-```
-rustup +nightly target add x86_64-pc-windows-gnu
-```
-```
-rustup default nightly-x86_64-pc-windows-gnu
-```
+    1. Install `cmake`, `ninja`, `clang` and other development dependencies
```
ChelseaChX, line 278: Full stop at the end.

docs/docs/contribute/getting-started.md
```diff
@@ -134,57 +218,97 @@ Search for and install the following extensions:
...
+    ```
 
-8. YAML
+    5. Install the Rust Web Assembly toolchain.
+    ```cmd
+    rustup +nightly target add wasm32-unknown-unknown
+    ```
 
-## Overview ##
+    6. Update Rust
```
ChelseaChX, line 254: Full stop at the end.

docs/docs/contribute/getting-started.md
```diff
@@ -134,57 +218,97 @@ Search for and install the following extensions:
...
+6. CMake Tools
+
+7. YAML
+
+
+## Overview
+
+=== "Windows"
+
+    1. Install [Chocolatey](https://docs.chocolatey.org/en-us/choco/setup#installing-chocolatey)
```
ChelseaChX, line 230: Full stop at the end.

docs/docs/contribute/getting-started.md
Kryptos-FR: I actually just copied the link that is on the chocolatey page. I could just remove that line and just keep "follow instructions to install Chocolatey" but since their page is not easy to read I wanted to help direct the user here.

I don't want to copy/paste the command line to install chocolatey from their page because it could change in the future and we shouldn't have to maintain that part.

docs/docs/contribute/getting-started.md
Kryptos-FR: I thought about that but given that we also have a note below the command it didn't feel right to me. I'm unsure for that one, I can't find a good place to have it.

docs/docs/contribute/getting-started.md
Kryptos-FR: GitHub Desktop is not available officially on Linux and usually Linux people have their favorite client and don't like to be told what to use :sweat_smile:

docs/docs/contribute/getting-started.md
Kryptos-FR: I'll keep it as optional. Linux as lots of distribution and people use different methods to get their package so, It's good to mention alternatives.

docs/docs/contribute/getting-started.md
ChelseaChX: I was thinking we could move the Tip on opening the command prompt at line 96 to this segment instead.

So the instructions could be something like:
    
```
    1. Open up the Command Prompt.

    ??? Tip
        You can use the Windows Search Bar to find the Command Prompt application.
        ![Search for the Command Prompt in the Window’s search bar.](assets/search-command-prompt.png)
     
     2. Use the Command Prompt to install Chocolatey by following the instructions [here](https://docs.chocolatey.org/en-us/choco/setup#installing-chocolatey).
```

docs/docs/contribute/getting-started.md
ChelseaChX: Noted 👍🏻

docs/docs/contribute/code/build-shards.md
```diff
@@ -35,29 +35,26 @@ Go to the [Shard’s repository on Github](https://github.com/fragcolor-xyz/shar
...
 ??? "cygpath"
-    The MinGW terminal requires you to convert the backslashes(\\) in your directory address to forward slashes(/). You will also have to remove the colon(:) after the drive alphabet. 
+    The Git Bash terminal requires you to convert the backslashes(\\) in your directory address to forward slashes(/). You will also have to remove the colon(:) after the drive alphabet. 
     
     This is the Unix form of an address, and can be done automatically by using the cygpath program.
 
     e.g. `cygpath -u 'C:\Fragcolor\Shards'` will output `/c/Fragcolor/Shards`
 
-If your Shards repository is located at `C:\Fragcolor\Shards`, the command used in the MinGW terminal would be `cd $(cygpath -u 'C:\Fragcolor\Shards')`.
+If your Shards repository is located at `C:\Fragcolor\Shards`, the command used in the terminal would be `cd $(cygpath -u 'C:\Fragcolor\Shards')`.
```
ChelseaChX, line 57: We'll have to remove the mention of Fragcolor here too.

Kryptos-FR: I'll simplify to just `C:\Shards` then

docs/docs/contribute/getting-started.md
Kryptos-FR: The thing is chocolatey must use an administrative command prompt (i.e. open a prompt with admin privileges) but the rest of the commands should not (I think).

PR #612, opened 2023-03-08
Refactor set/ref/update

sinkingsugar:
to solve multiple issues with tables


src/core/shards/core.hpp
```diff
@@ -777,56 +778,59 @@ struct SetBase : public VariableBase {
...
+    cloneVar(*_target, input);
+    return *_target;
   }
 };
 
-struct Set : public SetBase {
+struct Set : public SetUpdateBase {
   bool _exposed{false};
+  Shard *_self{};
+  entt::connection _onStartConnection{};
```
guusw, line 833: What is this event for?

sinkingsugar: Some lines under, but basically to avoid updating variables that rare sets when loading domain. (Will apply to exposed variables)

PR #614, opened 2023-03-09
fix inlining for good

sinkingsugar:


guusw:
Okay to switch that define from inline to always_inline.
However moving the functions to inlined.cpp doesn't actually do anything. When inlining is enabled the translation unit looks like this:
```cpp
#include "runtime.cpp"
#include "shards/inlined.cpp"
#include "shards/core.cpp"
#include "shards/math.cpp"
```
So it'll already contain all those functions in the same translation unit

rust/src/shards/gui/containers/window.rs
```diff
@@ -245,6 +270,15 @@ impl Shard for Window {
...
       let title: &str = self.title.get().try_into()?;
       let mut window = egui::Window::new(title);
+      if let Ok(id) = <&str>::try_from(self.id.get()) {
+        if let Some(id) = self.cached_id {
+          window = window.id(id);
+        } else {
+          let id = egui::Id::new(id);
+          window = window.id(id);
+          self.cached_id = Some(id);
+        }
```
Kryptos-FR, line 282: I think this can be written also like this:
```rust
let id = self.cached_id.get_or_insert_with(|| egui::Id::new(id));
window = window.id(id);
```

rust/src/shards/gui/mod.rs
```diff
@@ -31,6 +31,7 @@ static FLOAT_VAR_OR_NONE_SLICE: &[Type] = &[
   common_type::none,
 ];
 static FLOAT2_VAR_SLICE: &[Type] = &[common_type::float2, common_type::float2_var];
+static HASH_VAR_OR_NONE_SLICE: &[Type] = &[common_type::none, common_type::int2_var];
```
Kryptos-FR, line 34: Shouldn't it be named `INT2_VAR_OR_NONE_SLICE` or do you want to have the actual type an implementation details so that we use the same type everywhere we need a hash, but it could become something else than int2?

rust/src/types.rs
```diff
@@ -2212,13 +2212,9 @@ impl TryFrom<&Var> for &str {
...
       && var.valueType != SHType_Path
       && var.valueType != SHType_ContextVar
-      && var.valueType != SHType_None
     {
-      Err("Expected None, String, Path or ContextVar variable, but casting failed.")
+      Err("Expected String, Path or ContextVar variable, but casting failed.")
     } else {
-      if var.valueType == SHType_None {
-        return Ok("");
-      }
```
Kryptos-FR, line 2221: I understand the reason, but it might break somewhere... Let's keep an eye on it.

rust/src/shards/gui/mod.rs
sinkingsugar: Yeah that was the idea, if it was INT2 would better put to types.rs also, (in fact we should unify at some point)

rust/src/types.rs
sinkingsugar: yea tbh I think it's my fault, some of my old old old code was using this hack....

rust/src/shards/gui/containers/window.rs
sinkingsugar: I will have a look

sinkingsugar: hmm sort of, but I prefer the clarity actually, functional rust is weird

PR #617, opened 2023-03-10
Add various state save/restore shards

guusw:
- EGUI memory (window states)
- Window state


sinkingsugar (changes requested):
SaveMemory -> SaveState etc

PR #620, opened 2023-03-11
Use delta timers

sinkingsugar:
Improved timing as now it will try to recover overhead etc


sinkingsugar:
@guusw this seems to do better in tests (`timing.edn`) but doesn't help much in Rare

sinkingsugar:
This is wrong again, I feel like `devel` is the most solid in fact

PR #627, opened 2023-03-14
Make numeric DragValues logarithmic

guusw:


rust/src/shards/gui/widgets/drag_value.rs
```diff
@@ -0,0 +1,36 @@
...
+    let mut drag_value = DragValue::new(value);
+    drag_value = adjust_drag_value(&num_copy, drag_value);
+    Self { drag_value }
+  }
+
+  pub fn from_egui<Num: emath::Numeric>(value: &Num, drag_value: DragValue<'a>) -> Self {
+    Self {
+      drag_value: adjust_drag_value(value, drag_value),
+    }
+  }
```
Kryptos-FR, line 29: When is this called?

Kryptos-FR (changes requested):
I like the idea but it is extremely slow to change the value at the beginning.

Also, I think it should be optional. And the same could be applied to the slider: you only applied to the numeric input, which makes the title misleading 🤣.

rust/src/shards/gui/widgets/drag_value.rs
guusw: Yeah will remove this once

guusw:
Hmm I think there is a weird issue that it depends on screen size, so if you don't have a 4k screen it goes slower

guusw:
Actually it seems okay either way when I adjust the scaling (from 100% to 300%).
For sliders you usually have a fixed range, this was purely meant to make scrolling more intuitive for a large range of values (Unity/UE does something similar)

sinkingsugar:
Merging cos we need for GDC, we got a ton of things to fix in the UI after, start making a list.

PR #628, opened 2023-03-14
Guus/various

guusw:
Add shader ComposeWith to inject static variables into shader code


sinkingsugar:
@guusw coverage issue and conflicts

PR #641, opened 2023-03-23
experiment to fix SDL on android

sinkingsugar:
Simple fix to call into main/parent thread.


sinkingsugar:
@guusw Get inspired by this to fix Android SDL. Just wrap everything in a "run on main" which stops coro, runs on main F and resumes coro. Hyper fast, simple.
image

guusw:
Need to resume to the root in case we have nested Step for example

sinkingsugar:
Closed in #835 etc

PR #642, opened 2023-03-23
Explicitly link ntdll when building test-gfx on Windows

Kryptos-FR:
Following changes introduced by rust-lang/rust#108262


sinkingsugar:
Not happy, rust should fix it, `ntdll` is not meant to be used directly as it's a private library...

sinkingsugar:
wow they really did it, just saw the PR.. super weird... btw it's only for GNU toolchain. so we should at least respect that.

Kryptos-FR:
> wow they really did it, just saw the PR.. super weird... btw it's only for GNU toolchain. so we should at least respect that.

No it's actually only failing for MSVC. They made it available for gnu and as a side-effect made MSVC fail in our case. Not sure why it only occurs for test-gfx but it looks like it was already included in other places when we build shards libraries.

PR #647, opened 2023-03-29
Fix rust version to a specific nightly

guusw:
Try to keep a fixed working nightly version to avoid sudden breakages


Kryptos-FR:
Not sure I like this. Breakage are relatively rare and usually fixed during the next nightly (we witnessed this recently).

With this we would have to also maintain this file and it is likely that developers will have a more recent nightly on their machine that the one specified here.

In addition this doesn't even fix it since the CMakeFiles still forces the latest nightly.

PR #648, opened 2023-03-30
Various

sinkingsugar:


src/core/async.hpp
```diff
@@ -20,79 +112,100 @@ struct SharedThreadPoolConcurrency {
...
 #if defined(__EMSCRIPTEN__) && !defined(__EMSCRIPTEN_PTHREADS__)
   return func();
 #else
-  SHVar res{};
-  std::packaged_task<SHVar()> task(std::forward<FUNC>(func));
-  auto future = task.get_future();
+  struct BlockingCall : TidePool::Work {
+    BlockingCall(FUNC &&func) : func(std::move(func)), exp(), res(), complete(false) {}
+
+    virtual ~BlockingCall() {}
```
guusw, line 126: Doesn't need virtual destructor, only if you were to destroy it from it's base class, but then you'd need it on Work instead

guusw:
Just some style comments but looks good

src/core/async.hpp
```diff
@@ -1,11 +1,103 @@
...
+struct TidePool {
+  struct Work {
+    virtual void call() = 0;
+  };
+
+  TidePool() {
+    _running = true;
+
+    // spawn workers first
+    for (size_t i = 0; i < NumWorkers; ++i) {
```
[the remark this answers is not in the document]
sinkingsugar: not the _running though, as it can race if destroyed immediately

src/core/async.hpp
sinkingsugar: I had a warning, that's why I added it in test_runtime which is similar.
Doesn't hurt anyway does it?

sinkingsugar: no warning in this header though seems indeed

sinkingsugar: err at least on win, will remove for now we see

PR #675, opened 2023-05-14
Gio/networking server refactor

sinkingsugar:


src/core/shards/network.cpp
```diff
@@ -254,45 +274,106 @@ struct Server : public NetworkBase {
...
+
+            // keep receiving
+            return do_receive();
+          }
+        });
   }
 
   std::vector<uint8_t> _buffer;
 
+  IterableExposedInfo _sharedCopy;
```
guusw, line 373: Can you use `ExposedInfo` instead? I want to get rid of this weird IterableArray eventually.

This should be possible:
```cpp
_sharedCopy = ExposedInfo(data.shared);
_sharedCopy.push_back(...);
```

src/core/shards/network.cpp
```diff
@@ -335,7 +457,58 @@ struct Server : public NetworkBase {
...
+
+  static inline Type PeerType{{SHType::Object, {.object = {.vendorId = CoreCC, .typeId = PeerCC}}}};
+  static inline Type PeerObjectType = Type::VariableOf(PeerType);
+
+  static SHTypesInfo outputTypes() { return PeerType; }
+
   NetworkPeer _peer;
+  ShardsVar _blks{};
+
+  static inline ParamsInfo params = ParamsInfo(
```
guusw, line 477: These need to be in requiredVariables, you can use the `collectRequiredVariables` helper in compose:
```cpp
ExposedInfo _required;
SHTypeInfo compose(const SHInstanceData &data) {
   _required.clear();
   collectRequiredVariables(data.shared, _required, (SHVar&)_addr);
   collectRequiredVariables(data.shared, _required, (SHVar&)_port);
}
```

guusw: Check the other shards as well since they have similar parameters.

guusw (changes requested):

src/core/shards/network.cpp
sinkingsugar: But shardsvar is not an actual variable, just shards no?

src/core/shards/network.cpp
sinkingsugar: but why? what you suggested would not work, I need a clone of the whole array.
Also what's weird of IterableArray, just a wrapper on top of the simple C array we have? Lets KISS?

sinkingsugar: Btw we do this in many other places.

src/core/shards/network.cpp
sinkingsugar: `CoreInfo::ShardsOrNone` nothing to require

guusw: I meant for _addr and _port

sinkingsugar: Ah good point

PR #680, opened 2023-05-19
Add RotationGizmo

Yttruire:
This pull request adds additional functionality for ShapeRenderer to draw solid Quads and Discs (2D) and their respectively shards, as well as the RotationGizmo and its shard. Refactors Handles and its usage in TranslationGizmo.

Refer to the individual commit for comments and more details on each change to the repo in that commit.

Adjustments may need to be made in the future for the exact rotation per unit distance moved while a handle is selected.

### DIFF
--- a/src/extra/gizmos/gizmos.cpp
+++ b/src/extra/gizmos/gizmos.cpp
@@ -4,6 +4,7 @@
 #include <linalg_shim.hpp>
 #include <params.hpp>
 #include <gfx/gizmos/translation_gizmo.hpp>
+#include <gfx/gizmos/rotation_gizmo.hpp>
 #include <stdexcept>
 #include <type_traits>
 
@@ -83,6 +84,91 @@ struct TranslationGizmo : public Base {
     return outputTypes().elements[0];
   }
 };
-void registerGizmoShards() { REGISTER_SHARD("Gizmos.Translation", TranslationGizmo); }
+
+struct RotationGizmo : public Base {
+  static SHTypesInfo inputTypes() {
+    static Types inputTypes = {{CoreInfo::Float4x4Type, gfx::Types::Drawable}};
+    return inputTypes;
+  }
+  static SHTypesInfo outputTypes() { return CoreInfo::Float4x4Type; }
+  static SHOptionalString help() { return SHCCSTR("Shows a rotation gizmo "); }
+
+  // declare parameter named scale
+  PARAM_VAR(_scale, "Scale", "Gizmo scale", {CoreInfo::NoneType, CoreInfo::FloatType, CoreInfo::FloatVarType});
+  PARAM_IMPL(RotationGizmo, PARAM_IMPL_FOR(_scale));
+
+  // gizmo from translation_gizmo.hpp, not this file
+  gfx::gizmos::RotationGizmo _gizmo{};
+
+  SHVar activate(SHContext *shContext, const SHVar &input) {
+    float4x4 inputMat;
+    std::function<void(float4x4 & mat)> applyOutputMat;
+
+    switch (input.valueType) {
+    // if sequence given (transform matrix)
+    case SHType::Seq:
+      inputMat = (shards::Mat4)input;
+      break;
+    // if object given (drawable)
+    case SHType::Object: {
+      gfx::SHDrawable *drawable = reinterpret_cast<gfx::SHDrawable *>(input.payload.objectValue);
+      std::visit(
+          [&](auto &drawable) {
+            // let T be the type of the drawable
+            using T = std::decay_t<decltype(drawable)>;
+            // if same type as MeshTreeDrawable
+            if constexpr (std::is_same_v<T, gfx::MeshTreeDrawable::Ptr>) {
+              // get the transform matrix from drawable, and set applyOutputMat to a function that sets the transform
+              inputMat = drawable->trs.getMatrix();
+              applyOutputMat = [&](float4x4 &outMat) { drawable->trs = outMat; };
+            } else { // not sure what type this is expecting, just a type with transform member variable?
+              inputMat = drawable->transform;
+              applyOutputMat = [&](float4x4 &outMat) { drawable->transform = outMat; };
+            }
+          },
+          drawable->drawable);
+      break;
+    }
+    default:
+      // unexpected input type
+      throw std::invalid_argument("input type");
+      break;
+    }
+
+    _gizmo.transform = inputMat;
+
+    // Scale based on screen distance
+    float3 gizmoLocation = gfx::extractTranslation(_gizmo.transform);
+    _gizmo.scale = _gizmoContext->gfxGizmoContext.renderer.getSize(gizmoLocation) * 0.3f;
+
+    _gizmoContext->gfxGizmoContext.updateGizmo(_gizmo);
+
+    // if valid applyOutputMat function, apply the new transform calculated by _gizmo to the drawable
+    if (applyOutputMat)
+      applyOutputMat(_gizmo.transform);
+
+    return reinterpret_cast<Mat4 &>(_gizmo.transform);
+  }
+
+  void warmup(SHContext *context) {
+    PARAM_WARMUP(context);
+    baseWarmup(context);
+  }
+
+  void cleanup() {
+    PARAM_CLEANUP();
+    baseCleanup();
+  }
+
+  SHTypeInfo compose(const SHInstanceData &data) {
+    gfx::composeCheckGfxThread(data);
+    return outputTypes().elements[0];
+  }
+};
+
+void registerGizmoShards() { 
+  REGISTER_SHARD("Gizmos.Translation", TranslationGizmo); 
+  REGISTER_SHARD("Gizmos.Rotation", RotationGizmo); 
+}
 } // namespace Gizmos
 } // namespace shards
--- a/src/extra/gizmos/gizmos.cpp
+++ b/src/extra/gizmos/gizmos.cpp
@@ -166,9 +166,9 @@ struct RotationGizmo : public Base {
   }
 };
 
-void registerGizmoShards() { 
-  REGISTER_SHARD("Gizmos.Translation", TranslationGizmo); 
-  REGISTER_SHARD("Gizmos.Rotation", RotationGizmo); 
+void registerGizmoShards() {
+  REGISTER_SHARD("Gizmos.Translation", TranslationGizmo);
+  REGISTER_SHARD("Gizmos.Rotation", RotationGizmo);
 }
 } // namespace Gizmos
 } // namespace shards
--- a/src/gfx/gizmos/gizmo_input.cpp
+++ b/src/gfx/gizmos/gizmo_input.cpp
@@ -18,10 +18,7 @@ void InputContext::begin(const InputState &inputState, ViewPtr view) {
   heldHandleUpdated = false;
 }
 
-// TODO: move raycasting out of updateHandle, maybe even call it with a distance parameter
-// let the gizmo itself handle the raycasting however it needs to instead perhaps
-
-// call with hit distance from raycast
+// Call with hit distance from raycast
 void InputContext::updateHandle(Handle &handle, float hitDistance) {
   bool isHeld = &handle == held;
 
--- a/src/gfx/gizmos/gizmo_input.cpp
+++ b/src/gfx/gizmos/gizmo_input.cpp
@@ -92,5 +92,13 @@ void InputContext::updateView(ViewPtr view) {
   rayDirection = linalg::normalize(unprojected.xyz() - eyeLocation);
 }
 
+float3x2 InputContext::getScreenSpacePlaneAxes() {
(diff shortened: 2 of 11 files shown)


guusw:
image
You can mark PR's as drafts in github , that way we won't automatically merge the PR 😉 
Like above but the button says "Mark as draft" or similar.

(although it'll prevent CI from running on draft PRs, so if you want that, un-draft it)

src/gfx/gizmos/gizmo_input.cpp
```diff
@@ -17,20 +18,18 @@ void InputContext::begin(const InputState &inputState, ViewPtr view) {
   heldHandleUpdated = false;
 }
 
+// TODO: move raycasting out of updateHandle, maybe even call it with a distance parameter
+// let the gizmo itself handle the raycasting however it needs to instead perhaps
 void InputContext::updateHandle(Handle &handle) {
```
guusw, line 23: Yeah I was thinking, instead of using inheritance and then calling `handle.resolveHover`, the InputContext can instead have a function `addRaycastHit(Handle&, float distance)` since the raycast happens during this call right now, it wouldn't be that much different to do the raycast outside of InputContext and just give it the results.

So usage in the translation gizmo will be something like this:
```cpp
  auto &box = ...;
  float4x4 invHandleTransform = linalg::inverse(...);
  float3 rayLoc1 = linalg::mul(invHandleTransform, float4(eyeLocation, 1)).xyz();
  float3 rayDir1 = linalg::mul(invHandleTransform, float4(rayDirection, 0)).xyz();
  float2 hit = intersectAABB(rayLoc1, rayDir1, box.min, box.max);
  if (hit.x < hit.y) {
   inputContext.addRaycastHit(handle, hit.x); // <- call here
  }
```
Handle is only used to identify and persist the held state of some control then.
Of course the above raycast can then be refactored into a function as well so other gizmos can use the same logic.

sinkingsugar:
👏👏👏
so far looks good👍

src/gfx/gizmos/gizmo_input.cpp
Yttruire: Right, that makes sense, I'll work on refactoring to that instead, likely I'll shift the raycast logic into `gizmo_math.hpp` to be used across gizmos. Thanks for the pointers!

src/gfx/gizmos/rotation_gizmo.hpp
```diff
@@ -111,10 +115,17 @@ struct RotationGizmo : public IGizmo, public IGizmoCallbacks {
...
 
-      // solnB using the plane normal to the eye location as the plane of rotation, where tangent is projected onto this plane
+      // solnB using the plane that is normal to the eye location as the plane of rotation (from the hitPoint), 
+      // where tangent is projected onto this plane. this solution is not using true screen space, but the plane should be
+      // almost parallel to the viewport, such that we do not get the same spin-out issue
+      // not sure if possible to get true distane in screen space because information about the true normal vector to the
+      // viewport is not visible to the gizmo (need to be calculated through e.g. InputContext::updateView?)
+      // builds upon the results of solnA rather than using the 2 lines above as-is
       dragNormalDir = linalg::normalize(context.eyeLocation - hitPoint);
       dragTangentDir = linalg::normalize(dragTangentDir - linalg::dot(dragTangentDir, dragNormalDir) * dragNormalDir);
```
Yttruire: solnA is basically raycasting onto a plane that the disc lies on, which means that if the disc is at an angle almost perpendicular to the viewport, the distance along the tangent to the hitPoint will grow exponentially and cause the object to spin-out

solnB is finding the tangent to the hitPoint along the disc, then projecting that tangent onto a plane whose normal vector goes from the hitPoint to the eyeLocation, so that the plane will never be too perpendicular. 

The tangent is to adjust the direction that the user should drag the handle along, to the point of the disc clicked on.

Yttruire: @guusw Currently theres some limitations of the current solution (solnB) having not true screen space becomes more obvious when the gizmo has been translated to the corner of the screen/viewport, where the amount of rotation per unit mouse distance moved in either direction becomes skewed.

In the future, if possible, I think this would likely be better if it can be linked to the movement of cursorPosition to overcome this issue, and possibly have the same feature as Unity where the user can wrap the cursor around the screen when it goes out of the screen while holding the RotationGizmo handle, so that the user may keep dragging and rotating without running out of screen space to drag the cursor.

Yttruire:
Note: If both RotationGizmo and TranslationGizmo are running at the same time, handles from both gizmos can be selected at the same time, and possibly result in awkward behaviour, but should not be an issue if we ensure only one runs at any time

Also, due to the refactoring of handles and raycasting for box, merging this will likely break the ScalingGizmo

Yttruire:
Currently working on adding 4th handle to rotate the object on the plane parallel to the screen. Not committed yet because it's still WIP

src/extra/gizmos/gizmos.cpp
```diff
@@ -83,6 +84,91 @@ struct TranslationGizmo : public Base {
...
+  // declare parameter named scale
+  PARAM_VAR(_scale, "Scale", "Gizmo scale", {CoreInfo::NoneType, CoreInfo::FloatType, CoreInfo::FloatVarType});
+  PARAM_IMPL(RotationGizmo, PARAM_IMPL_FOR(_scale));
+
+  // gizmo from translation_gizmo.hpp, not this file
+  gfx::gizmos::RotationGizmo _gizmo{};
+
+  SHVar activate(SHContext *shContext, const SHVar &input) {
+    float4x4 inputMat;
+    std::function<void(float4x4 & mat)> applyOutputMat;
```
sinkingsugar, line 105: This might cause a very useless allocation depending on the standard.
I know it's part of the original implementation but even there should be fixed, pointers are not dangerous just use plain C constructs instead of if/else. KISS. cc @guusw

sinkingsugar (approved):
Looks good, just a minor issue with std::function not being suitable in general but can be fixed later if becomes a bigger issue.

src/gfx/gizmos/rotation_gizmo.hpp
```diff
@@ -0,0 +1,184 @@
...
+      normal[i] = 1.0f;
+      handleSelectionDiscs[i].normal = normal;
+    }
+  }
+
+  // update from IGizmo, seems to update gizmo based on inputcontext
+  // updates mainly the hitbox for the handle and calls updateHandle to check if the handle is selected (via raycasting)
+  virtual void update(InputContext &inputContext) {
+    float3x3 invTransform = linalg::inverse(extractRotationMatrix(transform));
+    float3 localRayDir = linalg::mul(invTransform, inputContext.rayDirection);
```
guusw, line 52: Unused

src/tests/gfx-gizmos-rotation.edn
```diff
@@ -0,0 +1,109 @@
...
+
+    ; Draw on top of everything (ignore depth)
+    ;; translation gizmo itself, attached to .translation-0
+    (Gizmos.Context :Queue .editor-queue-no-depth :View .view
+                    :Content (->
+                              .transform-0 (Gizmos.Translation) > .transform-0))
+
+    (Gizmos.Context :Queue .editor-queue-no-depth :View .view
+                    :Content (->
+                              .transform-0 (Gizmos.Rotation) > .transform-0))
```
guusw, line 104: If you put it in the same content as the context above the input will work correctly and only be able to interact with one gizmo at a time

src/gfx/gizmos/rotation_gizmo.hpp
```diff
@@ -0,0 +1,184 @@
...
+      float delta = linalg::dot(deltaVec, dragTangentDir);
+      // Use this to check direction vector of tangent and delta for the effective distance moved along the tangent
+      // SPDLOG_DEBUG("dragTangentDir {} {} {}", dragTangentDir.x, dragTangentDir.y, dragTangentDir.z);
+      // SPDLOG_DEBUG("Delta: {}", delta);
+      float sinTheta = std::sin(delta);
+      float cosTheta = std::cos(delta);
+
+      float4x4 rotationMat;
+
+      switch (handleIndex) {
```
guusw, line 125: You can use something like
```
float3 axis;
axis[handleIndex] = 1.0f;
linalg::rotation_matrix(linalg::rotation_quat(axis, delta));
```
to keep it a bit simpler/shorter (the extra calculations don't matter too much for gizmos here)

src/gfx/gizmos/translation_gizmo.hpp
```diff
@@ -6,10 +6,15 @@
 
 namespace gfx {
 namespace gizmos {
+// Note: While only a single handle of the TranslationGizmo may be selected at any point in time,
```
guusw, line 9: Shouldn't occur if they are used within the same gizmo context (see other comment in *.edn)

src/tests/gfx-gizmos-rotation.edn
```diff
@@ -0,0 +1,109 @@
...
+                              ;; draws 2 rotation circles around imaginary object
+                              (Gizmos.Circle :Center (Float3 1 1 1) :XBase .xbase :YBase .ybase)
+                              ;; (Gizmos.Circle :Center (Float3 1 1 1) :XBase .zbase :YBase (Float3 -1 1 1) :Color (Float4 0.2 1.0 0.2 1.0) :Thickness 4)
+                              (Gizmos.Circle :Center (Float3 1 1 1) :XBase .xbase :YBase .norm :Color (Float4 0.2 1.0 0.2 1.0) :Thickness 1)
+                              ;; rect on xy plane
+
+                              (Gizmos.Rect :Center (Float3 0 0 0) :XBase .xbase :YBase .norm :Size (Float2 1 1) :Color (Float4 0.2 1.0 0.2 1.0) :Thickness 4)
+
+                              ;; solid rect test on xy plane
+                              ;; (Gizmos.SolidRect :Center (Float3 1 1 0) :XBase .xbase :YBase .ybase :Size (Float2 2 2) :Color (Float4 0 0.5 0 1.0))
```
guusw, line 83: Can you leave one of these enabled somewhere just so that they're ran as part of the test. Either in here or some separate .edn file

src/tests/gfx-gizmos-rotation.edn
```diff
@@ -0,0 +1,109 @@
...
+                              .transform-0 (Gizmos.Translation) > .transform-0))
+
+    (Gizmos.Context :Queue .editor-queue-no-depth :View .view
+                    :Content (->
+                              .transform-0 (Gizmos.Rotation) > .transform-0))
+
+    (GFX.Render :Steps .render-steps :View .view))))
+
+(schedule root test-wire)
+(if (run root timestep) nil (throw "Root tick failed"))
```
guusw, line 109: Change it to `(if (run root timestep 100))` (fixed duration) so CI doesn't get stuck when running it. (It picks up any gfx-*.edn files automatically)

guusw: Also you can undraft it later so CI can run

guusw (changes requested):

src/tests/gfx-gizmos-rotation.edn
Yttruire: Fixing the requested changes as well as fixing the 4th handle axis for the RotationGizmo, will undraft and run CI when all the features for the 4th handle work

Yttruire:
Requested changes have been resolved in the commit [9ddd16b](https://github.com/fragcolor-xyz/shards/pull/680/commits/9ddd16b7a1196d1b80d43eeb55941a15815ee49e)
- Remove extraneous thickness parameter for solid shapes
- Remove trailing commas in macros
- Remove unused code/cleanup
- Updated comments at the top of RotationGizmo and TranslationGizmo about using them in the same Gizmos.Context lest unexpected behaviour

Requested changes to the rotation gizmo .edn test file has been resolved in commit [0852dbc](https://github.com/fragcolor-xyz/shards/pull/680/commits/0852dbc9e8a1ac830cbeb605b4a2f40a9277437f)
- Change RotationGizmo and TranslationGizmo to use the same Gizmos.Context so only one handle may be interacted at a time
- Cleanup gfx-gizmos-rotation.edn to include better test cases for discs, solid rects and the rotation gizmo.

src/extra/gizmos/gizmos.cpp
Yttruire: I think this might not have a simple solution to it, but I may have some misunderstanding around it. Can @guusw confirm?

From what I understand, `applyOutputMat` has to be a function pointer rather than a plain C pointer because it is assigned a closure with captured variables, and it cannot be assigned to C pointers. There is a possible heavy cost due to virtual functions with std::function, but i'm not sure if the type erasure aspect of std::function is also whats allowing us to call `drawable->trs` and `drawable->transform`. I can't seem to declare `drawable` as a `Variant` in the closure provided to std::visit and still access those two members.

src/gfx/gizmos/rotation_gizmo.hpp
Yttruire: Now linked to a plane parallel to the screen space, but not linked to movement of cursorPosition, which may be a bit more efficient

PR #682, opened 2023-05-25
Fix PNG images being loaded without premultiplying its alpha channels

### LINKED ISSUE #671: Problem with alpha channels when dealing with LoadImage and GFX.Texture
**Describe the bug**
--
Possible issue with alpha channels when using LoadImage with certain images and loading them into textures via GFX.Texture. The alpha channels might not be premultiplied correctly at the moment.

**To Reproduce**
--
Specifically, this issue can be reproduced by doing 
`  (LoadImage "GlodImages/Damage_Effect/Damaged_Effect_1.png")
  (GFX.Texture))`
with the damaged effect image file from the shards glod-game tutorial and then rendering this texture on screen. Instead of a fully transparent image (expected), it will show a red almost opaque image (reality).

Running the code under `Full Code So Far` at the end of Step 7.2, right before Step 7.3 of the glod-game tutorial will replicate this exact issue.

**Screenshots**
Issue:
image
Expected:
image

Yttruire:
PNG images are guaranteed by the file standards to not contain premultiplied alpha channels, and use straight alpha channels. As a result, PNGs will be rendered wrongly as the un-premultiplied alpha channels are used to render the image. 

This PR fixes `LoadImage`. It will now check if the loaded image is a PNG and if so, premultiplies the alpha channels, then flags the image as having a premultiplied alpha channel. If images that are not PNG, it will assume that they are already premultiplied and simply flag them as premultiplied. This PR also adds two shards PremultiplyAlpha and DemultiplyAlpha for manually premultiplying the alpha and demultiplying alpha of an image.

Some refactoring is also done for the premultiplying and demultiplying of the alpha channels. `LoadImage`'s code for loading byte inputs and files have also been streamlined in the process.

Note: Checking image file type for PNG can be done according to the specifications @ http://www.libpng.org/pub/png/spec/1.2/PNG-Structure.html

Additional list of file signatures: https://en.wikipedia.org/wiki/List_of_file_signatures

Fixes #671


src/core/shards/serialization.cpp
```diff
@@ -296,69 +297,109 @@ struct LoadImage : public FileBase {
...
+          Imaging::premultiplyAlpha<uint16_t>(_output, _output, _output.payload.imageValue.width, _output.payload.imageValue.height);
+          break;
+        case 4:
+          Imaging::premultiplyAlpha<float>(_output, _output, _output.payload.imageValue.width, _output.payload.imageValue.height);
+          break;
+      }
+    } else {
+      // if not a PNG, assume that it has already been premultiplied
+      _output.payload.imageValue.flags |= SHIMAGE_FLAGS_PREMULTIPLIED_ALPHA;
     }
```
Yttruire, line 401: This is kind of a blanket solution though, I haven't fully verified that all of the image types supported by STBI, other than PNG, are all using premultiplied alpha channels. By attaching this flag to every image loaded by `LoadImage`, it also does mean that the `DemultiplyAlpha` shard will work on all images loaded through `LoadImage`, even if they aren't actually using premultiplied alpha channels. (The two new shards both do nothing if the SHIMAGE_FLAGS_PREMULTIPLIED_ALPHA is present for the `PremultiplyAlpha` shard and if it is not present for `DemultiplyAlpha`).

I think verifying that an image has actually been premultiplied or not may be impossible in [some cases](https://stackoverflow.com/q/9147973), so we can only do this based on specifications of the supported image file types

sinkingsugar: I actually am thinking that we should remove this implicit premultiply. 
Always prefer explicit behaviours and it just takes an extra shard after LoadImage after all. It often will also be the case alpha is not even needed and such after all.
LoadImage should just load image basically simply.

Yttruire: Right, not all images will even have an alpha channel. That makes sense.

But also to clarify, you mean removing the premultiply for PNGs as well right? I was thinking that some users might not know much about graphics and that PNGs specifically do not use premultiplied alpha channels and thus need to be premultiplied before it will render properly, so it might be easier to use if `LoadImage` handled it automatically. 

If we remove the implicit premultiply entirely, then shall we just write down the necessary additional step into the `LoadImage` documentation regarding PNGs and non-premultiplied alpha channel images as well as the tutorials?

sinkingsugar: Yeah I think we should remove the implicit premultiply. It's easy to explain, and and it's easy to fix just need to add a shard after LoadImage after all.

sinkingsugar: that or add a parameter to LoadImage, `Premultiply` or something so that it can be controlled

Yttruire: `PremultiplyAlpha` parameter has been added in [5f6ec7d](https://github.com/fragcolor-xyz/shards/pull/682/commits/5f6ec7d04ef5ffe2924c72209e717ba3cd3f9d16) with default value of false

src/core/shards/imaging.cpp
```diff
@@ -165,6 +165,172 @@ struct StripAlpha {
...
+    }
+
+    return output;
+  }
+
+private:
+  std::vector<uint8_t> _bytes;
+};
+
+// Premultiplies the alpha channel in input image and writes the resulting image to a given SHVar output
```
guusw, line 251: Copy pasted comment.
Edit or remove

src/core/shards/imaging.cpp
```diff
@@ -165,6 +165,172 @@ struct StripAlpha {
...
+      pixsize = 2;
+    else if ((input.payload.imageValue.flags & SHIMAGE_FLAGS_32BITS_FLOAT) == SHIMAGE_FLAGS_32BITS_FLOAT)
+      pixsize = 4;
+
+    int32_t w = int32_t(input.payload.imageValue.width);
+    int32_t h = int32_t(input.payload.imageValue.height);
+    
+    _bytes.resize(w * h * 4 * pixsize);
+    auto output = Var(&_bytes.front(), uint16_t(w), uint16_t(h), 4, input.payload.imageValue.flags);
+    output.version = input.version;
```
guusw, line 317: version is used to detect changes to images so you'd want to store it in a variable inside this shard and increment it every time regardless of input version. (same for DemultiplyAlpha)

src/core/shards/serialization.cpp
```diff
@@ -296,69 +304,101 @@ struct LoadImage : public FileBase {
...
-    case BPP::u16:
-      _output.payload.imageValue.flags = SHIMAGE_FLAGS_16BITS_INT;
-      break;
-    case BPP::f32:
-      _output.payload.imageValue.flags = SHIMAGE_FLAGS_32BITS_FLOAT;
-      break;
-    default:
-      _output.payload.imageValue.flags = 0;
-      break;
+      case BPP::u16:
```
guusw, line 374: Can you put these in the same switch as above?

guusw (changes requested):
Minor comments, but looks good.

Bit annoying that only UI expects pre-multiplied alpha currently, but no easy way around it since it has to go through egui.

guusw:
Also, you need to move the premultiple/unpremultiply template functions to a header (can add imaging.hpp) if you want to use them in serialization.cpp

src/core/shards/imaging.cpp
Yttruire: @guusw to clarify on this, if version is stored as a variable inside the shard and incremented every time regardless of input version, do you mean that its a static variable specific to the shard? Or should each version number be unique across all images and be incremented by the use of any of the shards in `imaging.cpp`?

I think I'm a bit unsure about the use of version because at the moment, it seems that all of the shards simply assign the output version to be that of the input's version without incrementing or storing any version variable.

sinkingsugar: Version is useful for images and bytes arrays and strings cos we do pointer comparisons sometimes but that's not fool proof, as in the end you compare just an address and a size, version is a bit like a weak hash for the content. So to know the image has changed indeed. Please search all usages of version and it should be clearer.

Yttruire: If that's the case then I think the version doesn't need to be stored in a variable inside the shard, but just copied from input to output and incremented every time since version is only used as a tiebreaker. 

I have made the changes for version in commit [c02c1a6](https://github.com/fragcolor-xyz/shards/pull/682/commits/c02c1a6373528119ab82d55ccbee35232eb01bb8). The only existing shard that increments the version is `LoadImage`, but that seems to be when the values in the SHVar are first initialized, so `_output.version++` may not be initialized, so I have changed that to `_output.version = 0`, and to increment the version of the output in every shard in `Imaging.cpp`. Do let me know if my understanding conflicts with the intended behaviour for version.

Else, this should be all the requested changes addressed for this PR.

sinkingsugar (approved):

Yttruire:
CI should be fixed now, seems to have been an issue with visibility of the implemented functions within the header file

src/core/shards/imaging.cpp
```diff
@@ -313,10 +397,10 @@ struct Resize {
 void registerShards() {
   REGISTER_SHARD("Convolve", Convolve);
```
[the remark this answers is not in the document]
sinkingsugar: cc @guusw @ccfontes

sinkingsugar:
@Yttruire why this is still not merged? cc @guusw

guusw:
@Yttruire Can you add some add a test case for PremultiplyAlpha & DemultiplyAlpha in the image script (or a new script)

PR #686, opened 2023-05-30
Gfx parameter table cleanup

vantemoon:
Changed the DrawablePass shard from taking an input table to taking parameters.


src/extra/gfx/steps.cpp
```diff
@@ -343,42 +355,40 @@ struct EffectPassShard {
...
-      wrapperFeature = std::make_shared<Feature>();
-      applyComposeWith(context, composeWithVar);
-    }
+    // TODO: Check if there should be a ComposeWith parameter in the first place
+    // SHVar composeWithVar;
+    // if (getFromTable(context, inputTable, "ComposeWith", composeWithVar)) {
+    //   // Always create a new object to force shader recompile
+    //   wrapperFeature = std::make_shared<Feature>();
+    //   applyComposeWith(context, composeWithVar);
+    // }
```
Yttruire, line 379: For `EffectPassShard`, it appears to have expected a `ComposeWith` parameter from the input table in the past and has an applyComposeWith function, but this parameter is not mentioned in the original possible expected table format and .edn test files involving `EffectPassShard` have never called it with a `ComposeWith` key in its input table before. 

Thus need to clarify if this shard actually needs a ComposeWith parameter, or if the ComposeWith code was mistakenly leftover from modifying a copy of `DrawablePassShard`

src/extra/gfx/feature.cpp
```diff
@@ -293,9 +868,13 @@ struct FeatureShard {
 
     generatorInstanceData.inputType = GeneratedDrawInputTableType;
     _drawableGeneratorBranch.compose(generatorInstanceData);
+    // _drawableGeneratorsRaw = _drawableGenerators;
+    _drawableGeneratorBranch.setRunnables(_drawableGenerators);
 
     generatorInstanceData.inputType = GeneratedViewInputTableType;
     _viewGeneratorBranch.compose(generatorInstanceData);
+    // _viewGeneratorsRaw = _viewGenerators;
+    _viewGeneratorBranch.setRunnables(_viewGenerators);
```
vantemoon, line 877: The current implementation would result in a format exception as the shard cannot resolve the output type of the generators.

src/extra/gfx/steps.cpp
guusw: Yeah, it's not in any of the shards samples but an example can be found in the [fragcolor-xyz/rare-examples/gfx/gfx-master-pbr.edn](https://github.com/fragcolor-xyz/rare-examples/blob/6543ebf517a53507ab878f2ee0f91d282df5315d/gfx/gfx-master-pbr.edn#L35)
```edn
...
{:ComposeWith {:pbr-wgsl .file-pbr-wgsl.json}
   :EntryPoint (->
                (Shader.Literal :Type ShaderLiteralType.Header :Source .pbr-wgsl)
                (Shader.ReadInput "texCoord0") >= .uv
                (Shader.Literal :OutputDimension 2 :Source ["computeLUT(" .uv ", 1024)"])
                (ToFloat2) (Shader.WriteOutput "color") ; End shader
                .uv (Shader.WriteOutput "color"))}
  (GFX.EffectPass) >> .lut-steps-1
...
```

src/extra/gfx/feature.cpp
vantemoon: Call `setRunnables` in `composeGeneratorWire` results in the following error being thrown:
```
[trace] [2023-06-07 11:33:33.875] [T-142709] [foundation.hpp::434] Creating wire: <shader wire>
[error] [2023-06-07 11:33:33.876] [T-142709] [runtime.cpp::1091] Error composing shard: Failed to find shader parameter "time" in any buffer, wire: <shader wire>
```

PR #690, opened 2023-06-08
Track only a single exposed/required type per name

guusw:
- Fixup table types in Set.
- Return existing table type in Update.

Main thing this does is support the test cases in the added table-compose.edn:
```edn
{:a 0 :b 1 :c 2} >= .table
.table (Do wire-1)

; This should update b in the table
4 (Update .table :Key "b")
.table (Do wire-1)

5 (Set .table :Key "d")
.table (Do wire-2)
```

### DIFF
--- a/src/tests/failures.clj
+++ b/src/tests/failures.clj
@@ -88,19 +88,6 @@
 (schedule Root d)
 (if (run Root 0.1) (throw "Failure expected") nil)
 
-(defwire table-push-mix
-  {:k [1 2 3]} >= .table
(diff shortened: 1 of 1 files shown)


guusw:
TODO: Check if we still need isPushTable, since we have stable maps now.

sinkingsugar (commented):
Looks good I hope it works smoothly, for some reasons this is something I never wanted to touch!

sinkingsugar:
@guusw add coverage and possibly less coverage means some silent failure too! check logs

guusw:
Merged in devel ee544f4426b761c0782bafb84594f4cd29da8cd4

PR #694, opened 2023-06-11
Table key, String -> Var, C++ is done

sinkingsugar:


guusw:
We still have `VariableBase`:
```cpp
ParamsInfo::Param("Key",
                                   SHCCSTR("The key of the value to read/write from/in the table (parameter applicable only if "
                                           "the target variable is or will become a table)."),
                                   CoreInfo::StringStringVarOrNone)
```
Might have to update the type to Any, there might be some other cases.

guusw:
Merged in devel ee544f4426b761c0782bafb84594f4cd29da8cd4

PR #698, opened 2023-06-16
Add type variables & type builtin

guusw:
To replace provide a more straightforward alternative to `ExpectLike` & `ExpectX` by specifying types directly


guusw:
Merged in #699

PR #709, opened 2023-06-21
New Input System 

guusw:
This contains input changes and #656 

## New Input Backend Overview

There's 2 main things this system attempts to solve:

- Detaching the main input polling (& window) loop from places that actually check for input.
- Adding layers that can consume certain input events so they are ignore by the layers below them

### Window changes

Activations of `(GFX.MainWindow)` now run the main input loop, you can still use it as previously but to get the full separation you should run `MainWindow` without any Content parameter, this will cause it to no longer create a built-in renderer & graphics context. Alternatively you could have Contents handle input and have a separate renderer by passing the `:DetachRenderer true` parameter.

A separate renderer can be created and attached to the window like below:

```clojure
(GFX.MainWindow
  :Title "Rare"
  :Width 1280
  :Height 720) = .main-window

 (GFX.Renderer
   :Window .main-window
   :IgnoreCompilationErrors true ; This param was previously on MainWindow
   :Contents (-> ; GFX code here
   ))
```

### Window size changes

Specified initial window sizes are now display scaling aware, so a size of 1280x720 will approximately take up the same amount of screen space regardless of the amount of pixels

### Detached Input

To address the input layer issue, a new shard has been added: `Inputs.Detached`

It can be used anywhere & at any frequency independently from the input loop. The Name parameter is used for debug purposes and the Priority parameter controls in which order the input events are handled.

The consumption of input events are currently controlled internally by the `InputContext`'s (each detached input has it's own) `ConsumeFlags`:

```cpp
struct ConsumeFlags {
  bool wantsPointerInput{};
  bool wantsKeyboardInput{};
  bool requestFocus{};
};
```

Various shards such as the Camera/UI/Gizmos set these flags based on the last inputs it received. If various detached inputs run at different frequencies this might cause some delays but will generally not cause any issues and has been shown to work well currently for rare.

Example:

```clojure
; Layer 0, will handle any input last
(Inputs.Detached
   .main-window
   :Name "Game"
   :Priority 0
   :Contents (-> 
     (Inputs.IsKeyDown "w")
     ...))

; Layer1, UI layer, any input consumed by the UI will not be sent to the layers below it.
(Inputs.Detached
   .main-window
   :Name "UI"
   :Priority 1
   :Contents (-> 
     (UI (->... )))
```


guusw:
Feel free to leave any comments/feedback.

deps/CMakeLists.txt
```diff
@@ -49,8 +49,8 @@ if(NOT EMSCRIPTEN)
       message(STATUS "Fetching libressl")
       FetchContent_Declare(
         libressl
-        # URL      https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.4.1.tar.gz
-        URL      https://cloudflare.cdn.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.4.1.tar.gz
+        URL      https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.4.1.tar.gz
```
guusw, line 52: TODO: Undo this change

shards/gfx/texture.cpp
```diff
@@ -121,7 +121,7 @@ void Texture::initContextData(Context &context, TextureContextData &contextData)
     }
 
     if (textureFormatFlagsContains(desc.format.flags, TextureFormatFlags::RenderAttachment)) {
-      wgpuDesc.usage |= WGPUTextureUsage_RenderAttachment;
+      wgpuDesc.usage |= WGPUTextureUsage_RenderAttachment | WGPUTextureUsage_CopySrc;
```
sinkingsugar, line 124: Is this efficient? prob not

guusw, line 124: TODO: Check this, add a flag

guusw:
TODO: Test on iOS

shards/modules/imaging/imaging.cpp
```diff
@@ -308,11 +312,115 @@ struct Resize {
...
+  }
+};
+
 } // namespace Imaging
 } // namespace shards
 
 SHARDS_REGISTER_FN(imaging) {
   using namespace shards::Imaging;
+
+  REGISTER_SHARD("GetImagePixel", ImageGetPixel);
```
sinkingsugar, line 423: We never mention Image, lets call it `GetPixel`, or actually `TakePixel`

shards/modules/inputs/inputs.cpp
```diff
@@ -536,39 +513,45 @@ struct HandleURL : public Base {
   static SHTypeInfo outputType() { return CoreInfo::AnyType; }
 
   SHVar activate(SHContext *context, const SHVar &input) {
-    for (auto &ev : _inputContext->events) {
-      if (ev.type == SDL_DROPFILE || ev.type == SDL_DROPTEXT) {
-        _url.clear();
-        auto url = ev.drop.file;
-        _url.append(url);
-        SDL_free(url);
+    // TODO: Input
```
sinkingsugar, line 516: Hmm we need this on iOS cos it's the same thing to transfer things between apps

sinkingsugar (commented):
Looks good but a couple of things

shards/gfx/texture.cpp
guusw: #729

shards/modules/imaging/imaging.cpp
guusw: #730

shards/modules/inputs/inputs.cpp
guusw: #731

guusw:
Merged through guus/picking

PR #716, opened 2023-06-30
Fix panics in UI.Style

Yttruire:
Fix panics that occur when invaild input (invalid value/type or missing mandatory parameters) is given to UI.Style shard.

Panics that have been tested and fixed (No longer panics but gives an Error and ends gracefully):
- Invalid type supplied to `:override_text_style`
- Non-existent name supplied to `:override_text_style` (I.e. not an existent `TextStyle` and does not match any `text_style` added in `:text_styles`
- Invalid type supplied to `:text_styles`
- Invalid type or value supplied to `:name`, `:size`, `:family` in `:text_styles`, including negative size or non-existent family
- Invalid type supplied to `:visuals` and `:override_text_color`
- Better error messages using shlog! for all attributes including `:widgets` and `:selection`


sinkingsugar (approved):
Could you format with "cargo fmt" on the module folder please, for the rest lgtm

Yttruire:
cargo fmt has been run on the egui folder

PR #719, opened 2023-07-06
Fix strings highlighting

ccfontes:


shards/modules/egui/src/widgets/code_editor/sublime-syntax.yml
```diff
@@ -44,7 +44,10 @@ contexts:
     - match: '[a-z_][a-zA-Z0-9_.-]*'
       scope: variable.other.readwrite.shards
   strings:
-    - match: '".*"'
-      scope: string.quoted.double.shards
-    - match: '""".*"""'
```
sinkingsugar, line 49: Don't we need those?

shards/modules/egui/src/widgets/code_editor/sublime-syntax.yml
```diff
@@ -44,7 +44,10 @@ contexts:
...
-      scope: string.quoted.triple.shards
+    - match: '"|"""'
+      scope: string.quoted.shards
+      push:
+        - meta_content_scope: string.quoted.shards
+        - match: '"|"""'
```
sinkingsugar, line 51: why the `|`?

shards/modules/egui/src/widgets/code_editor/sublime-syntax.yml
ccfontes: No, that's what's causing the issue.

shards/modules/egui/src/widgets/code_editor/sublime-syntax.yml
ccfontes: To match a double quote or a triple double quote.
I'm going to make it a bit more correct.

sinkingsugar: Oh so unrelated to our `|`?

ccfontes: Yup, unrelated, it's just an "or" for a regex

ccfontes:
Fixed by #725
CC #Doc-105

PR #722, opened 2023-07-10
Translate test scripts to the new syntax

vantemoon:


guusw (changes requested):
Could you remove the original clj/edn files once they're at least translated fully (doesn't have to be fully working, just translated)
and also change the command lines in the .github/workflows/*.yml files so they run the new tests?

sinkingsugar:
I'm merging this @vantemoon , you can continue on another branch/PR

PR #724, opened 2023-07-11
Add source code editor for Shards

ccfontes:


utils/run-editor.shs
```diff
@@ -0,0 +1,17 @@
+@wire(editor-wire {
```
guusw, line 1: We could probably just have this as a UI.CodeEditor sample once we support showing the new syntax in the docs pages.

ccfontes: Do you mean moving this runner to the samples?
If so, this is not just a small sample buried down in the docs sub folders or copy pasteable from docs.fragnova.com. It's also a utility for development purposes, that one could easily find execute in the `utils` directory, with instructions like purpose in a companion readme.

ccfontes: I will merge in an hour or so since the comment regards future plans anyways.

PR #725, opened 2023-07-11
Syntax highlight: fix multiple strings in same line and escaping

ccfontes:


ccfontes:
Fixes cases in these examples:
```
foo("bar") | spam("eggs")
foo("""\\\.bar""") | spam("""eggs""")
""eggs"" ;; error: bad double quote matching
"""\\.eggs"""
""""eggs""" ;; error: bad double quote matching
```

May require removing `build` and/or `target` directories beforehand.

sinkingsugar (approved):

ccfontes:
Fixes #Doc-105

PR #752, opened 2023-08-07
Naga C++ bindings

guusw:
Interop layer for naga which allows shader modules to be constructed directly from C++.
Includes a test that generates a simple test shader


sinkingsugar (commented):
It's a lot of code, seems to include PBR too right?

guusw:
Yeah wanted to merge that first, just check the top few commits (anything after "Fix compileShaderFromWgsl on wasm")

PR #753, opened 2023-08-08
Side menu mock implementation + UI Classes

Yttruire:
Both the changes on the shards side in order to support UI elements that are necessary for the side menu in Rare, as well as a revamp to how UI currently can be done are included in this branch.

The changes for the side menu mock implementation include exposing certain types and the addition of UI.Layout shard which exposes the full functionality of egui::Layout rather than bit and pieces via shards like UI.Vertical and UI.Horizontal.

UI classes serve as an approach that is more similar to how web development frontend does styling. UI.Layout is improved with more functionality and options, and these options can be set and saved in objects called LayoutClasses. These LayoutClasses can be reused in multiple UI.Layout shards in order to setup the layout similarly (E.g. Create multiple textboxes with scrolling and framing).


shards/egui/src/layouts/layout.rs
```diff
@@ -0,0 +1,875 @@
...
+      fill_width: ParamVar::default(),
+      fill_height: ParamVar::default(),
+      disabled: ParamVar::default(),
+      frame: ParamVar::default(),
+      scroll_area: ParamVar::default(),
+    }
+  }
+}
+
+impl Shard for LayoutConstructor {
```
Yttruire, line 170: As per previously discussed, might need your guidance on the use of compose to verify types in a table parameter @guusw for this LayoutConstructor shard that takes tables of various types as parameters

PR #761, opened 2023-08-22
Mesh skinning

guusw:
Depends on #752 & #732


sinkingsugar:
Looks good but I see a lot of old syntax. We should do something about it.

PR #771, opened 2023-08-29
Window close, Popups, and Autogrid

Yttruire:
This branch adds the following:
- Layout and LayoutClass shards
- Makes certain variables in Shards public for use in Rare 
  - For UI elements such as the CircularImageButton in Rare (For the purpose of side menu in Rare)
- Updates SVG crate and SVG.ToImage, also adding an offset parameter


Yttruire:
@guusw Need your advice on adding type safety in the LayoutClass Shard's compose function I think. Regarding the type safety of variables in a table parameter passed to the shard that was mentioned a while ago

shards/rust/src/types.rs
```diff
@@ -4327,14 +4327,14 @@ macro_rules! shenum {
         const $EnumValue:ident = $value:expr;
       )+
     }
-    struct $SHEnumInfo:ident { }
+    pub struct $SHEnumInfo:ident { }
```
guusw, line 4330: Better to just add a  `$vis_2:vis` and then reference it below as
`$vis_2 struct $SHEnumInfo {` since this only changes the matched pattern, this would change it so the defined  SHEnumInfo uses the same visibility that was put here.

guusw (changes requested):

guusw:
Looks good, but I think the main thing I'd change would be to change the tables to params where possible
```clj
(UI.LayoutClass
   :Layout {:main_direction LayoutDirection.LeftToRight}
   :ScrollArea {:horizontal_scroll_enabled true})

=>

(UI.LayoutClass
   :MainDirection LayoutDirection.LeftToRight
   :HorizontalScrollEnabled true)
```

shards/rust/src/types.rs
Yttruire: Should be resolved in [9fff79b](https://github.com/fragcolor-xyz/shards/pull/771/commits/9fff79be08e8761316c8b9ee887e10812da626a4)

shards/egui/src/layouts/layout.rs
```diff
@@ -0,0 +1,1045 @@
...
+            retrieve_layout_class_attribute!(parent_layout_class, layout, main_dir).unwrap()
+          } else {
+            return Err("Invalid main direction provided. Main direction is a required parameter");
+          }
+        };
+
+        let mut layout = egui::Layout::from_main_dir_and_cross_align(main_direction, cross_align);
+
+        let main_align = if let Some(main_align) =
+          retrieve_enum_parameter!(layout_table, "main_align", LayoutAlign, LayoutAlignCC)
```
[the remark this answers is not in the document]
Yttruire: All table params changed to direct params on the main shard [28dbda6](https://github.com/fragcolor-xyz/shards/pull/771/commits/28dbda6978bc03765dbec62a589af304bf6a1bd2)

docs/samples/shards/UI/Layout/1.edn
```diff
@@ -0,0 +1,31 @@
+(GFX.MainWindow
+ :Contents
+ (->
+  (Setup
+   (GFX.DrawQueue) >= .ui-draw-queue
+   (GFX.UIPass .ui-draw-queue) >> .render-steps
+   (UI.LayoutClass
+    :Layout {:main_direction LayoutDirection.LeftToRight})
```
guusw, line 8: These seem to be broken currently (still using old Layout param), can you either remove them (leave just the shs tests or fix them)

guusw (changes requested):

docs/samples/shards/UI/Layout/1.edn
Yttruire: The old .edn files have been removed.

Separately, I've updated the `run-sample.shs` to use FS.Read and it seems to be reading the sample file now, and doesn't complain that if the file is missing either, but doesn't seem to be executing the file's shards code itself, do you know how to run the code in a file @guusw?

docs/samples/shards/UI/Area/2.shs
```diff
@@ -0,0 +1,71 @@
+@mesh(main)
```
guusw, line 1: This one still contains the manual run boilerplate (probably doesn't work with run-sample)

guusw (changes requested):

docs/samples/shards/UI/Area/2.shs
Yttruire: Thanks for catching, forgot to remove after testing. The manual run boilerplate has been removed in [06e0b24](https://github.com/fragcolor-xyz/shards/pull/771/commits/06e0b24f886f2a7cda3d248a783784f61818c0b3)

guusw:
I'd like to avoid having similar but slightly different shards like the popup shards. I think we can implement a generic popup by having most of our interactive shards store the last response they generate on the UI context. The popup shard can then be a parent of a button for example and use the response to render the popup.

If you check https://github.com/fragcolor-xyz/shards/blob/65b677e96012050a593c6094b888b1062fbb55ed/shards/egui/src/lib.rs#L128 you can see I change the context into a struct already so you could add an `Option<Response>` or something to keep the last response, then clear it before the next frame.

Yttruire:
Since the last merge, PopupWrapper, window closable & open fields, AutoGrid shard and some other minor changes have been made. drag-and-drop branch has also been rebased. For your reference and review @guusw 

Not sure why certain test files like gfx-cube.shs are failing though?

guusw:
Moved to #815 with some minor changes

PR #780, opened 2023-09-06
Fix double free & fix kdtree.shs

guusw:
Fixes a double free when Recur cleanup is ran multiple times.

Add the kdtree.shs test back that triggered this


guusw:
@vantemoon This is now merged, alongside your converted kdtree.shs test

PR #801, opened 2023-09-27
Daily

sinkingsugar:


shards/modules/network/network.cpp
```diff
@@ -546,16 +564,58 @@ struct Server : public NetworkBase {
...
+            assert(receivedSize == nextChunkSize);
+
+            // If this is the first chunk, read the expected total size from its prefix.
+            if (offset == 0) {
+              peer->expectedSize = *(uint32_t *)peer->recvBuffer.data();
+              // if (peer->expectedSize > 25000) {
+              //   SHLOG_TRACE("Receiving big message: {}", peer->expectedSize);
+              // }
+            }
+
```
guusw, line 596: Can you put this in a function and use it in the code below, looks like it does the same?

guusw (approved):

PR #820, opened 2023-10-16
Gio/16-10-2023

sinkingsugar:
**Description**

**Test plan**


shards/core/runtime.cpp
```diff
@@ -3032,7 +3033,6 @@ void shInit() {
   shInterface.set("createMesh", emscripten::val(reinterpret_cast<uintptr_t>(iface->createMesh)));
   shInterface.set("destroyMesh", emscripten::val(reinterpret_cast<uintptr_t>(iface->destroyMesh)));
   shInterface.set("schedule", emscripten::val(reinterpret_cast<uintptr_t>(iface->schedule)));
-  shInterface.set("unschedule", emscripten::val(reinterpret_cast<uintptr_t>(iface->unschedule)));
```
guusw, line 3035: What's the new way to unscheduled now? Or just not possible in the new C API?

guusw (approved):
Looks good, just a question

shards/core/runtime.cpp
sinkingsugar: it's better to stop the wire, always was

sinkingsugar: but I could revert that bit

PR #824, opened 2023-10-17
Abstract coroutine backend & debug helpers

guusw:
- Supports running coroutines as thread (with a CMake toggle, SHARDS_THREAD_FIBER=ON)
- Add thread names
image


sinkingsugar:
Wasm failing is a red flag, showing probably something changed too much.

shards/core/coro.cpp
```diff
@@ -1,34 +1,132 @@
...
 
-void SHCoro::init(const std::function<void()> &func) {
+void Fiber::init(const std::function<void()> &func) {
   SHLOG_TRACE("EM FIBER INIT");
   this->func = func;
   c_stack = new (std::align_val_t{16}) uint8_t[stack_size];
   emscripten_fiber_init(&em_fiber, action, this, c_stack, stack_size, asyncify_stack, as_stack_size);
+
+  // Initial resume to match functionality of other Fiber implementations
+  resume();
```
sinkingsugar, line 121: This might be breaking wasm, behavior changes from past no?

shards/core/runtime.cpp
```diff
@@ -571,13 +572,9 @@ SHWireState suspend(SHContext *context, double seconds) {
...
-#ifndef __EMSCRIPTEN__
-  context->continuation = context->continuation.resume();
-#else
-  context->continuation->yield();
-#endif
-
-  // RESUMING
+  SH_CORO_SUSPENDED(context->currentWire());
+  coroutineSuspend(*context->continuation);
+  SH_CORO_RESUMED(context->currentWire());
```
sinkingsugar, line 577: This is also weird

shards/core/runtime.cpp
```diff
@@ -2037,14 +2021,9 @@ void run(SHWire *wire, SHFlow *flow, SHCoro *coro)
...
-  return std::move(context.continuation);
-#else
-  context.continuation->yield();
-#endif
-
-  // we should never resume here!
-  SHLOG_FATAL("Wire {} resumed after ending", wire->name);
+  wire->context = nullptr;
+  
+  SH_CORO_SUSPENDED(wire)
```
sinkingsugar, line 2026: This should never happen

shards/core/utils.hpp
```diff
@@ -0,0 +1,64 @@
...
+#if SH_WINDOWS
+inline std::wstring toWindowsWString(std::string_view utf8) {
+  std::wstring result;
+  result.resize(MultiByteToWideChar(CP_UTF8, 0, utf8.data(), utf8.size(), nullptr, 0));
+  result.resize(MultiByteToWideChar(CP_UTF8, 0, utf8.data(), utf8.size(), result.data(), result.size()));
+  return result;
+}
+#endif
+
+inline void setThreadName(std::string_view name_sv) {
```
sinkingsugar, line 35: Nice

sinkingsugar (commented):
Tried to find red flags as wasm failing is suspicious

shards/core/coro.cpp
guusw: Only for emscripten to match the behavior of boost callcc (which resumes initially, emscripten doesn't)
I moved it here, the resume was initially in the prepare call itself

shards/core/runtime.cpp
guusw: Just for debug thread names, the suspend is the same as previous

shards/core/runtime.cpp
guusw: Fixed this, CI is green as wel now

PR #855, opened 2023-11-03
Fixes taken from input branch

guusw:


shards/core/ref_output_pool.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+template <typename T> struct RefOutputPoolItemTraits {
+  T newItem() { assert("not implemented"); }
+  void release(T &) { assert("not implemented"); }
+  size_t getRefCount(T &v) { assert("not implemented"); }
+};
+
+// Keeps a pool of output objects
+// the objects are assumed to be self ref counted, so destruction is not handled by the pool
+template <typename T> struct RefOutputPool {
```
sinkingsugar, line 17: We probably wanna add a dedicated test in test-runtime? Also maybe some `shassert` for runtime to avoid leaking somehow?

shards/core/runtime.hpp
```diff
@@ -798,6 +798,8 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
   }
 
   void addRef(const SHStringWithLen name, SHVar *var) {
+    assert((var->flags & SHVAR_FLAGS_REF_COUNTED) == SHVAR_FLAGS_REF_COUNTED);
```
sinkingsugar, line 801: use `shassert`

shards/modules/audio/audio.cpp
```diff
@@ -379,7 +379,7 @@ struct Channel {
     std::array<SHVar, 2> stereo;
     stereo[0] = Var(0);
     stereo[1] = Var(1);
-    _inChannels = Var(stereo);
```
sinkingsugar, line 382: revert, this is already fixed no?

shards/modules/core/core.hpp
```diff
@@ -1221,11 +1221,7 @@ struct Ref : public SetBase {
 
   ALWAYS_INLINE SHVar activateRegular(SHContext *context, const SHVar &input) {
     // must keep flags!
-    const auto rc = _target->refcount;
-    const auto flags = _target->flags;
-    memcpy(_target, &input, sizeof(SHVar));
-    _target->refcount = rc;
-    _target->flags = flags;
+    assignVariableValue(*_target, input);
```
sinkingsugar, line 1224: This is not gonna work also I don't want a redirection here, it's supposed to be as bare as possible on purpose.

shards/modules/debug/debug.cpp
```diff
@@ -0,0 +1,43 @@
+#include <shards/core/shared.hpp>
+#include <shards/core/params.hpp>
+#include <shards/common_types.hpp>
+
+namespace shards::Debug {
+struct DebugNoop {
```
sinkingsugar, line 6: there is already `Assert.Break`

sinkingsugar: Although this works differently I see! You just wanna put a breakpoint manually I guess here.

shards/modules/audio/audio.cpp
guusw: Still good to change the default I think. up to you

shards/modules/debug/debug.cpp
guusw: Yeah just kind of something I can put a breakpoint on without actually breaking always.

shards/core/ref_output_pool.hpp
guusw: Added

shards/modules/audio/audio.cpp
guusw: Undid the change

PR #858, opened 2023-11-04
Flow Refactor

sinkingsugar:
Refactor call on main thread to mesh scope (which is more correct), lowered tick pressure


sinkingsugar:
Untested btw

sinkingsugar:
Seems to work fine on fubble

guusw (approved):

shards/core/runtime.hpp
```diff
@@ -297,43 +299,13 @@ SHContext *&getCurrentContextPtr();
   }
 #else
 #define SH_CORO_RESUMED(_wire) \
-#define SH_CORO_SUSPENDED(_) \
```
guusw, line 300: I think this messed up a bit by formatting

PR #874, opened 2023-11-24
Guus/compose var type check

guusw:


shards/core/runtime.cpp
```diff
@@ -1677,8 +1530,8 @@ bool validateSetParam(Shard *shard, int index, const SHVar &value, SHValidationC
   for (uint32_t i = 0; param.valueTypes.len > i; i++) {
     // This only does a quick check to see if the type is roughly correct
     // ContextVariable types will be checked in validateConnection based on requiredVariables
-    if (matchTypes(varType, param.valueTypes.elements[i], true, true, false)) {
-      return true; // we are good just exit
+    if (matchTypes(varType, param.valueTypes.elements[i], true, true, true)) {
```
sinkingsugar, line 1533: We sure about this then?

sinkingsugar (approved):

sinkingsugar:
Can we wrap this @guusw ?

PR #877, opened 2023-12-01
Fix whendone logic

sinkingsugar:
**Description**

**Test plan**


shards/core/runtime.hpp
```diff
@@ -1752,8 +1752,11 @@ template <typename T> struct WireDoppelgangerPool {
 
       lock.lock();
       auto &fresh = _pool.emplace_back(std::make_shared<T>());
+      auto index = _pool.size();
+      lock.unlock();
```
guusw, line 1756: 😱 how did we miss this

guusw (approved):

PR #887, opened 2023-12-07
Add debug feature to CMakeLists.txt and Cargo.toml

sinkingsugar:
**Description**

**Test plan**

### DIFF
--- a/shards/rust/CMakeLists.txt
+++ b/shards/rust/CMakeLists.txt
@@ -1,7 +1,14 @@
 set(EXTRA_SOURCES ../../include/shards/shards.h)
 
+if((CMAKE_BUILD_TYPE STREQUAL "Release") OR (CMAKE_BUILD_TYPE STREQUAL "MinSizeRel"))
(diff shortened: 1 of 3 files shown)


guusw:
Might be easier to just disable debug in Rust.cmake since where you change it, it wouldn't get used for all the other crates.
Check cmake/Rust.cmake@57 where it adds the --release flag:
```cmake
if(CMAKE_BUILD_TYPE STREQUAL "Debug")
  set(RUST_BUILD_SUBDIR_CONFIGURATION debug)
else()
  set(RUST_CARGO_FLAGS_INT --release)
  set(RUST_BUILD_SUBDIR_CONFIGURATION release)
endif()
```

guusw:
Add a config file in project root(s)
.cargo/config.toml:
```toml
[profile.release-with-deb-info]
inherits = "release"
debug-assertions = true
debug = true
```


Change in Rust.Cmake:
```sh
cargo build --profile release-with-deb-info
```

PR #892, opened 2023-12-13
Render graph v2 (WIP)

guusw:
- Remove fixed sizes from render graph, which was not correct. Sizes are now checked during evaluation.
- Building is more flexible and allows automatic conversion of texture in different formats (srgb, float, wrong output size)


shards/modules/gfx/shards_types.hpp
```diff
@@ -340,7 +340,7 @@ struct Container {
 
   static inline ParameterInfo OutputScaleParameterInfo{"OutputScale",
                                                        SHCCSTR("The scale that the output should be rendered as"),
-                                                       {CoreInfo::NoneType, CoreInfo::Float2Type, CoreInfo::Float2VarType}};
+                                                       {CoreInfo::NoneType, CoreInfo::AnyType}};
```
sinkingsugar, line 343: seems weird?

shards/gfx/renderer_storage.hpp
```diff
@@ -43,7 +43,7 @@ struct RendererStorage {
     if (textureData.externalView)
       return textureData.externalView;
 
-    assert(textureData.texture);
+    assert(textureData.texture && "Invalid texture");
```
sinkingsugar, line 46: `shassert` preferred so will assert in RelWithDebInfo

shards/modules/gfx/shards_types.hpp
guusw: Didn't write tests for it yet but the new :OutputScale parameter is either:
`{main: none}` => Scale to main output
`none` =>      Same as above
`{main: @f2(0.25)}` => Scale to 25% of main output
`{main: 0.25}` =>  Same as above
`@f2(0.25)` => Same as above
`{input: "color" scale: @f2(0.25)}` => Scale to 25% of the color input

Had too many combinations and optionals to provide the type

shards/gfx/renderer_storage.hpp
guusw: Changed them all to shassert and moved shassert to shared header

PR #894, opened 2023-12-15
Guus/Fast String

guusw:
Implements string interning for usage inside graphics/animation where fast comparison & storage is preferred over lookups of the actual string


shards/fast_string/fmt.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef C15A8B87_4A27_4D56_AD9E_4DA2AACA21FE
+#define C15A8B87_4A27_4D56_AD9E_4DA2AACA21FE
+
+#include "fast_string.hpp"
+#include <spdlog/fmt/fmt.h>
+
+// template<> struct fmt::formatter<shards::fast_string::FastString> {
```
guusw, line 7: Apparently fmt somhow does this automatically already, probably using the implicit conversion, remove this.

sinkingsugar (approved):
Can you test with TSAN before merge? Do we run any test with TSAN in ci btw?

PR #898, opened 2023-12-20
Add Zip shard for combining sequences

sinkingsugar:
**Description**

**Test plan**

### DIFF
--- a/shards/tests/hello.shs
+++ b/shards/tests/hello.shs
@@ -205,4 +205,4 @@ WhenNot(@foo-predicate(-1) {Msg("More than -2")})
 ["4" "5" "6"] = zs2
 ["7" "8" "9"] = zs3
 ["10" "20" "30" "40"] = zs4
-Zip([zs1 zs2 zs3 zs4]) | Assert.Is([["1" "4" "7" "10"] ["2" "5" "8" "20"] ["3" "6" "9" "30"]]) | Log("Zipped")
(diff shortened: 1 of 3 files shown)


shards/tests/hello.shs
```diff
@@ -200,3 +200,9 @@ WhenNot(@foo-predicate(-1) {Msg("More than -2")})
 "Hello World!" | String.Starts(With: "Hello") | Assert.Is(true) | Log
 ; also test that doesn't start with
 "Hello World!" | LogType("Type logging test") | String.Starts(With: "World") | Assert.Is(false) | Log
+
+["1" "2" "3"] = zs1
+["4" "5" "6"] = zs2
+["7" "8" "9"] = zs3
+["10" "20" "30" "40"] = zs4
+[zs1 zs2 zs3 zs4] | Zip | Assert.Is([["1" "4" "7" "10"] ["2" "5" "8" "20"] ["3" "6" "9" "30"]]) | Log("Zipped")
```
guusw, line 208: Maybe better as Zip(a b c d ...), can pass by reference in that case too

guusw (changes requested):

shards/tests/hello.shs
guusw: Have to do some manual setParam magic

sinkingsugar: But we don't have N params hmm

guusw: Never mind, we better just use a single seq param and allow vars & const sequences inside into a single parameter

guusw (approved):

PR #906, opened 2023-12-29
eagerly transform constants to var

sinkingsugar:
**Description**

**Test plan**


shards/modules/clipboard/clipboard.cpp
```diff
@@ -27,7 +27,9 @@ struct GetClipboard {
 
   SHVar activate(SHContext *context, const SHVar &input) {
     _output.clear();
-    _output = SDL_GetClipboardText();
+    auto clipboardPtr = SDL_GetClipboardText();
```
guusw, line 30: Good catch

guusw (approved):
Looks good

PR #922, opened 2024-01-22
Guus/pick multi pre

guusw:
Preliminary GFX changes for multi-select picking


sinkingsugar (approved):
It's ok, Big, next time maybe renaming should go in another PR :D

guusw:
Yeah tried to at least do this before changing the rest, but renaming could've been split more.
It's a separate commit at least

PR #924, opened 2024-01-24
Guus/pick multi

guusw:


include/shards/shards.h
```diff
@@ -590,6 +590,8 @@ struct SHVarPayload {
 // used internally to cancel and abort a wire flow, must be String type and receiver must abort and destroy the var
 // used only in async.hpp and rust side async activation so far
 #define SHVAR_FLAGS_ABORT (1 << 5)
+// can be used alongside foreign to mark a foreign variable as writable
+#define SHVAR_FLAGS_FOREIGN_WRITABLE (1 << 6)
```
sinkingsugar: But foreign can be written by default, should be fine.

shards/modules/inputs/inputs.cpp
```diff
@@ -620,4 +668,6 @@ SHARDS_REGISTER_FN(inputs) {
   using KeyUp = KeyUpDown<false>;
   REGISTER_SHARD("Inputs.KeyDown", KeyDown);
   REGISTER_SHARD("Inputs.KeyUp", KeyUp);
+
+  REGISTER_SHARD("Inputs.MatchModifierKeys", MatchModifierKeys);
```
sinkingsugar: A bit long, `MatchModifier` should be enough , we already have `Input` as context

shards/rust/src/core.rs
```diff
@@ -485,7 +485,11 @@ impl WireRef {
       len: name.len() as u64,
     };
     unsafe {
-      (*Core).setExternalVariable.unwrap()(self.0, name, &var.0 as *const _ as *mut _);
+      let ev = SHExternalVariable {
+        var: &var.0 as *const _ as *mut _,
+        type_: std::ptr::null(),
```
sinkingsugar, line 490: is that ok?

sinkingsugar (changes requested):
LGTM just some minor requests

include/shards/shards.h
guusw: No there's an assert in update I believe

shards/rust/src/core.rs
guusw: It's optional, null will still derive the type as before

shards/core/runtime.cpp
```diff
@@ -1936,7 +1943,9 @@ NO_INLINE void _destroyVarSlow(SHVar &var) {
 }
 
 NO_INLINE void _cloneVarSlow(SHVar &dst, const SHVar &src) {
-  shassert((dst.flags & SHVAR_FLAGS_FOREIGN) != SHVAR_FLAGS_FOREIGN && "cannot clone into a foreign var");
```
guusw: @sinkingsugar this one

sinkingsugar: Ah true, but then it's external?

shards/modules/inputs/inputs.cpp
guusw: Ok

shards/core/runtime.cpp
guusw: Changed it

guusw:
Merged manually

PR #929, opened 2024-02-02
Disable update_while_editing on variable widgets

guusw:


guusw:
Maybe current behavior is best for now

PR #943, opened 2024-02-26
Guus/asset things

guusw:
- Add DoubleClick for UI.Selectable
- Add Debug.DumpEnv
- Add DoOnce (C++)
- Add tracy markers

### DIFF
--- a/shards/modules/core/time.hpp
+++ b/shards/modules/core/time.hpp
@@ -16,7 +16,6 @@ struct ProcessClock {
 
 using namespace std::chrono_literals;
 struct DeltaTimer {
-
   // Limit delta time to avoid jumps after unpausing wires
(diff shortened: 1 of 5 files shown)


shards/modules/core/time.hpp
```diff
@@ -31,6 +30,26 @@ struct DeltaTimer {
     return delta.count();
   }
 };
+
+struct DoOnce {
```
sinkingsugar, line 34: Where do we use this?

sinkingsugar (approved):

shards/modules/core/time.hpp
guusw: I just used it in a place to log a warning message but only every 2 seconds at most
(similar to the old timer warning, but this time in fbl)

PR #953, opened 2024-03-25
Guus/ambal

guusw:


sinkingsugar (approved):

shards/modules/core/core.hpp
```diff
@@ -3576,7 +3576,7 @@ struct Repeat {
   }
 
   FLATTEN ALWAYS_INLINE SHVar activate(SHContext *context, const SHVar &input) {
-    auto repeats = _forever ? 1 : *_repeats;
+    auto repeats = (_forever || _pred) ? 1 : *_repeats;
```
sinkingsugar: You don't  need this logic during activate I think, Can be done directly in setParam no?

PR #959, opened 2024-04-15
Forbid usage of exposed variable in various mutating shards

sinkingsugar:
**Description**

**Test plan**

### DIFF
--- a/shards/modules/core/core.hpp
+++ b/shards/modules/core/core.hpp
@@ -2261,6 +2261,11 @@ struct Count : SeqUser {
     return SHCCSTR("Count of characters, elements, or key-value pairs contained in the `:Name` parameter variable.");
   }
 
+  SHTypeInfo compose(const SHInstanceData &data) {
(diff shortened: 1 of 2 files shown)


guusw (approved):

shards/modules/core/core.hpp
```diff
@@ -2199,6 +2212,18 @@ struct SeqUser : VariableBase {
...
+  SHTypeInfo compose(const SHInstanceData &data) {
+    for (auto &shared : data.shared) {
+      std::string_view vName(shared.name);
+      if (vName == _name && shared.exposed) {
+        SHLOG_ERROR("Exposed variable {} can only be updated using Update.", _name);
+        throw ComposeError("Exposed variables can only be updated using Update.");
+      }
+    }
+
+    return data.inputType;
```
guusw, line 2224: Wrong return type, since not all shards that derive from SeqUser return the input type (like count)

guusw (changes requested):

PR #975, opened 2024-05-06
add FS.Rename

sinkingsugar:
**Description**

**Test plan**

### DIFF
--- a/shards/tests/general.shs
+++ b/shards/tests/general.shs
@@ -821,7 +821,9 @@
   "text-copy.txt"
   FS.Remove
   Assert.Is(true true)
-  "text-copy2.txt" | FS.Absolute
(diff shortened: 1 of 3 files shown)


shards/tests/general.shs
```diff
@@ -821,7 +821,9 @@
   "text-copy.txt"
   FS.Remove
   Assert.Is(true true)
-  "text-copy2.txt" | FS.Absolute
+  "text-copy2.txt"
+  FS.Rename("text-copy3.txt")
+  "text-copy3.txt" | FS.Absolute
   FS.Remove
```
guusw, line 827: Can you add a test for the failure case

guusw (approved):
One thing, otherwise okay to merge.

PR #977, opened 2024-05-07
Remove useless code and fix define

sinkingsugar:
**Description**

**Test plan**


sinkingsugar:
Code was not useless but just not complete. We need to add a better way to check if identifiers exists globally

PR #989, opened 2024-06-03
add ast replacement utilities

sinkingsugar:


shards/lang/src/eval.rs
```diff
@@ -2096,6 +2115,19 @@ fn create_shard(
     return Err((format!("Forbidden shard {}", shard.name.name), line_info).into());
   }
 
+  let mut replacement_storage = None;
```
guusw, line 2118: Could you put this in a function, you wouldn't need this option in that case, just return

shards/lang/src/lib.rs
```diff
@@ -222,323 +222,328 @@ pub trait ShardsExtension {
...
 }
 
-#[repr(C)]
-pub struct SHLError {
-  message: *mut c_char,
-  line: u32,
-  column: u32,
-}
+#[cfg(feature = "exported")]
+mod exported {
```
guusw, line 226: Did you just shuffle stuff around here? Diff is a bit messy

guusw (approved):

shards/lang/src/lib.rs
sinkingsugar: added the mod tab

sinkingsugar: btw this trick does not work in release... I get duped symbols

PR #997, opened 2024-06-13
Add more documenting Shards (enum, types)

sinkingsugar:

### DIFF
--- a/shards/core/foundation.hpp
+++ b/shards/core/foundation.hpp
@@ -690,10 +690,10 @@ template <typename T> inline void arrayResize(T &arr, uint32_t size) {
   arr.len = size;
 }
 
-template <typename T>
-inline void arrayShuffle(T &arr) {
+template <typename T> inline void arrayShuffle(T &arr) {
   // Check if the array is empty
-  if (arr.len == 0) return;
+  if (arr.len == 0)
+    return;
 
   // Random number generator
   static thread_local std::random_device rd;
@@ -950,6 +950,7 @@ ALWAYS_INLINE inline void destroyVar(SHVar &var) {
   switch (var.valueType) {
   case SHType::None:
     return;
+  case SHType::Trait:
   case SHType::Type:
   case SHType::Table:
   case SHType::Set:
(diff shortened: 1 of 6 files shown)


shards/core/foundation.hpp
```diff
@@ -950,6 +950,7 @@ ALWAYS_INLINE inline void destroyVar(SHVar &var) {
   switch (var.valueType) {
   case SHType::None:
     return;
+  case SHType::Trait:
   case SHType::Type:
   case SHType::Table:
   case SHType::Set:
```
guusw, line 956: You should remove the `default: break;` so it will warn about missing cases

shards/modules/core/core.cpp
```diff
@@ -1482,6 +1482,34 @@ struct GetShards {
...
 };
 
+struct GetEnumTypes {
+  static SHTypesInfo inputTypes() { return CoreInfo::NoneType; }
+  static SHTypesInfo outputTypes() { return CoreInfo::IntSeqType; }
+
+  SeqVar _output{};
+
+  SHVar activate(SHContext *context, const SHVar &input) {
+    for (auto [id, _] : shards::GetGlobals().EnumTypesRegister) {
```
guusw, line 1494: Don't forget to clear or once internally (`if !empty()`)

shards/modules/core/core.cpp
```diff
@@ -1482,6 +1482,34 @@ struct GetShards {
...
+};
+
+struct GetObjectTypes {
+  static SHTypesInfo inputTypes() { return CoreInfo::NoneType; }
+  static SHTypesInfo outputTypes() { return CoreInfo::IntSeqType; }
+
+  SeqVar _output{};
+
+  SHVar activate(SHContext *context, const SHVar &input) {
+    for (auto [id, _] : shards::GetGlobals().ObjectTypesRegister) {
```
guusw, line 1510: ditto

guusw (changes requested):

shards/core/foundation.hpp
sinkingsugar: cannot really cos many blittable types pass from here, I will just add an assert

guusw (approved):

PR #1003, opened 2024-06-25
Visual Shards

sinkingsugar:


shards/tests/ui-style.shs
```diff
@@ -6,49 +6,30 @@
   Once({
```
sinkingsugar: Revert this and add new sample

include/shards/shards.swift
```diff
@@ -11,8 +11,8 @@ public struct Globals {
 
     init() {
         Core = shardsInterface(UInt32(SHARDS_CURRENT_ABI))
-        print("Shards Swift runtime initialized!")
-        Core.pointee.registerShard(BaseShard.name.withUnsafeBufferPointer{ $0.baseAddress }, { createSwiftShard(BaseShard.self) })
+        // print("Shards Swift runtime initialized!")
```
guusw: Clean up comment

sinkingsugar: Oops this should be in other pr as well 😅. But it's simple I'll wrap it here

shards/egui/src/visual/mod.rs
```diff
@@ -0,0 +1,2521 @@
...
+    ast.accept_mut(self)
+  }
+
+  fn mutate_shard(&mut self, x: &mut Function) -> Option<Response> {
+    let state = x.get_or_insert_custom_state(|| FunctionState {
+      params_sorted: false,
+    });
+    let params_sorted = state.params_sorted;
+
+    // // check if we have a result from a pending operation
```
guusw, line 404: Remove comment

shards/egui/src/visual/mod.rs
```diff
@@ -0,0 +1,2521 @@
...
+                      //   .unwrap();
+                      // x.get_custom_state::<FunctionState>().unwrap().receiver =
+                      //   Some(UniqueReceiver::new(receiver));
+
+                      // open a dialog to change the value
+
+                      let mouse_pos = ui
+                        .ctx()
+                        .input(|i| i.pointer.hover_pos().unwrap_or_default());
+                      self.context.swap_state = Some(SwapState::Param(ParamSwapState {
```
guusw, line 547: Is this for the context menu? I think egui already has that with open state tracking and all.
You can just have a button and then:
```rs
let response = ui.button( .... );
response.context_menu(|ui| {
  ...
});
```

shards/egui/src/visual/mod.rs
```diff
@@ -0,0 +1,2521 @@
...
+          Some(ui.allocate_response(egui::Vec2::ZERO, egui::Sense::hover()))
+        }
+      })
+      .inner
+  }
+
+  fn visit_assignment(&mut self, assignment: &mut Assignment) -> Option<Response> {
+    let mut combined_response = None;
+    let (r_a, r_b) = match assignment {
+      Assignment::AssignRef(pipeline, identifier) => (pipeline.accept_mut(self), {
```
guusw, line 1370: I'd still go with the same symbols as text, but maybe we can make some nice icons representing `>/=/>>/>=`

shards/egui/src/visual/mod.rs
```diff
@@ -0,0 +1,2521 @@
...
+              id: Id::new(nanoid!(16)),
+            })
+            .selected = selected;
+          });
+          i += 1;
+          r
+        }
+      };
+
+      // if let Some(response) = &response {
```
guusw, line 1460: Cleanup

shards/egui/src/visual/mod.rs
sinkingsugar: Not sure what you mean because the end of the operation will happen on different frames anyways.
This is literally when we are selecting a new shard  or replace

shards/egui/src/visual/mod.rs
sinkingsugar: Was keeping those as memo on how to use it once I had to actually

shards/egui/src/visual/mod.rs
sinkingsugar: Was keeping those as memo on how to use it once I had to actually

shards/egui/src/visual/mod.rs
sinkingsugar: We could, not sure how much people will associate and use both modes though tbh, but ok let's start with same

guusw (approved):

PR #1013, opened 2024-07-09
Gio/visual ast 3

sinkingsugar:


shards/rust/src/types.rs
```diff
@@ -319,6 +320,19 @@ impl Drop for Wire {
   }
 }
 
+// typedef void(__cdecl *SHSetWireError)(const SHWire *, void *errorData, struct SHStringWithLen msg);
+type SHSetWireError = unsafe extern "C" fn(*const SHWire, *mut c_void, SHStringWithLen);
+
+// void shards_set_wire_error_callback(SHWire *wire, SHSetWireError setWireError, void *errorData)
+// the above is defined c side, lets declare it here
+extern "C" {
```
guusw, line 328: This prevents using the rust library when building dll shards, not sure if you care about that still?

guusw: Would have to place it in SHCore to be correct

shards/tests/ui-style.shs
```diff
@@ -6,49 +6,30 @@
...
-          Open: {
-            BGFill: @color(255 255 255 255)
-            Stroke: @no-stroke
-            BGStroke: @stroke-2
-          }
-        )
-        "Label" | UI.Label
-        UI.Button("Button")
-        UI.TextField(text)
+      ; UI.Window(Title: "1" Contents: {
```
guusw: Doesn't work? can you revert?

guusw: Or make this a separate test, since seems like you hijacked this one for the UI.Shards

shards/egui/src/widgets/text_field.rs
```diff
@@ -192,11 +206,19 @@ impl Shard for TextField {
       text_edit = text_edit.hint_text(RichText::from(hint).color(egui::Color32::GRAY).italics());
     }
 
-    text_edit = if !self.justify_width.get().is_none() && self.justify_width.get().try_into()? {
-      text_edit.desired_width(f32::INFINITY)
+    if let Ok(desired_width) = TryInto::<f64>::try_into(&self.desired_width.0) {
```
guusw: Can you swap around, so if you set justify_width, it ignores desired width. It think that makes more sense, for example when you want to toggle that on and off while testing.

guusw (approved):

shards/rust/src/types.rs
sinkingsugar: I think I fixed those by using shards-langffi

sinkingsugar: see I don't want it in SHCore cos it's rather internal.

sinkingsugar: I will test it though

shards/tests/ui-style.shs
sinkingsugar: eh I thought I fixed this... again git town messing things up

PR #1014, opened 2024-07-12
Ast improvements

sinkingsugar:


shards/egui/src/widgets/image.rs
```diff
@@ -145,9 +154,17 @@ impl Image {
...
+    // support rendering multiple images from same shard
+    let current_step = shards::core::step_count(context);
+    if self.step_counter != current_step {
+      self.step_counter = current_step;
+      self.step_textures.clear();
+    }
+
     if let Some(ui) = util::get_current_parent_opt(self.parents.get())? {
-      let (texture_id, texture_size) = image_util::get_egui_texture_from_gfx(input)?;
+      let (texture_id, texture_size) =
```
guusw: Isn't it easier to just store it on the UI context? 
we have a custom rust struct for that already

shards/input/state.hpp
```diff
@@ -139,9 +144,12 @@ struct InputState {
     pointers.pointers.clear();
     int numDevices{};
     auto touchDevices = SDL_GetTouchDevices(&numDevices);
+    currentTouchDevice.reset(touchDevices, SDL_free);
```
guusw: Don't need to store here, can just `DEFER({SDL_free(...)})`

guusw (changes requested):

shards/input/state.hpp
sinkingsugar: That did not work, cos outFiner.touchId is actually copying

shards/egui/src/widgets/image.rs
sinkingsugar: Yeah, I suppose, I will check

shards/input/state.hpp
sinkingsugar: hmm but it's just a uint64.... somehow all my inputs were broken though If I didn't do this....
Will try to understand. I had defer before.

shards/egui/src/widgets/image.rs
sinkingsugar: But for example in this Shard it's not available, is that normal btw? I guess `parents` include it?

shards/input/state.hpp
sinkingsugar: Worked, I think I had different slow down issues when I tested.

shards/core/type_info.cpp
```diff
@@ -119,6 +119,8 @@ SHTypeInfo deriveTypeInfo(const SHVar &value, const SHInstanceData &data, std::v
       if (!types.count(derived)) {
         shards::arrayPush(varType.seqTypes, derived);
         types.insert(derived);
+      } else {
```
sinkingsugar, line 122: 🤦‍♂️

Issue #1021, opened 2024-07-22
Improved comments in `exposed_type_utils.hpp`


eltociear:
**Describe the bug**
refence -> reference

PR #1022, opened 2024-07-22
chore: update exposed_type_utils.hpp

eltociear:
refence -> reference

#1021


guusw:
Thanks, since it's a small change I included it in another branch 7eb70ad43

PR #1023, opened 2024-07-25
Optimize compose & DoMany

guusw:


sinkingsugar:
Still very surprised that boost flat_map was so much slower

sinkingsugar:
Btw the best optimization would be to recycle this https://github.com/fragcolor-xyz/shards/blob/1fa4e158c499f422ede2d46ec92720a16d936a4d/shards/core/runtime.cpp#L850 - should be easy to keep it in context

guusw:
> Btw the best optimization would be to recycle this
> 
> https://github.com/fragcolor-xyz/shards/blob/1fa4e158c499f422ede2d46ec92720a16d936a4d/shards/core/runtime.cpp#L850
> - should be easy to keep it in context

Didn't even see that one, can fix that yes

guusw:
Oh that's devel, nvm

shards/core/serialization.hpp
```diff
@@ -151,6 +156,11 @@ struct Serialization {
 
     output.valueType = nextType;
 
+    // Uncomment for more detailed breakdown
```
sinkingsugar: Maybe `#if 0` it instead?

shards/core/wire_doppelganger_pool.hpp
```diff
@@ -41,63 +59,131 @@ template <typename T> struct WireDoppelgangerPool {
...
+    requires WireData<T>
+  {
+    ZoneScoped;
 
-      return value;
-    }
+    auto batch = acquireBatch(1);
+    return acquireFromBatch(batch, 0, composer, anything);
+
+    // pmr::SharedTempAllocator tempAlloc;
```
sinkingsugar: Delete?

sinkingsugar (changes requested):
Minor stuff

sinkingsugar (approved):

PR #1024, opened 2024-07-29
add a list of peers to exclude from Network.Broadcast

sinkingsugar:


shards/core/serialization.hpp
```diff
@@ -493,6 +493,9 @@ struct Serialization {
         } else {
```
guusw, line 496: Shouldn't it check above if info.deserialize is not null

guusw (approved):

PR #1028, opened 2024-08-01
Shah docs

ShahFragcolor:

### DIFF
--- a/shards/modules/core/wires.cpp
+++ b/shards/modules/core/wires.cpp
@@ -244,9 +244,18 @@ struct Wait : public WireBase {
     passthrough = false; // also need this to have proper compose output type
   }
 
+  static SHOptionalString inputHelp() {
+    return SHCCSTR("Any input type is accepted. The input value will either pass through unchanged or be ignored.");
+  }
+
+  static SHOptionalString outputHelp() {
+    return SHCCSTR("If Passthrough is true, this shard outputs the input value, passed through unchanged. Otherwise, it "
+                   "returns the output of the Wire it waited for.");
+  }
+
   SHOptionalString help() {
-    return SHCCSTR("Waits for another wire to complete before resuming "
-                   "execution of the current wire.");
+    return SHCCSTR("Waits for the specified Wire to complete before resuming "
+                   "execution of the current Wire.");
   }
 
   SHExposedTypeInfo _requiredWire{};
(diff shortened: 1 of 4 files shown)


shards/modules/core/wires.cpp
```diff
@@ -244,9 +244,18 @@ struct Wait : public WireBase {
     passthrough = false; // also need this to have proper compose output type
   }
 
+  static SHOptionalString inputHelp() {
```
sinkingsugar, line 247: In the case of passthru we should not repeat the string all the time, we should have a string declared in a common place and always reuse that for consistance and less memory usage

sinkingsugar:
```
struct CoreInfo {
  static inline SHOptionalString PTInputHelp = SHCCSTR("Passthru input");
  static inline SHOptionalString PTOutputHelp = SHCCSTR("Passthru input")
  ```

File: shards/modules/core/wires.cpp [hunk 2]
```diff
@@ -357,11 +366,22 @@ struct IsRunning : public WireBase {
     passthrough = false; // also need this to have proper compose output type
   }
 
-  SHOptionalString help() { return SHCCSTR("Checks if a wire is running and outputs true if that is the case, false if not."); }
+  static SHOptionalString inputHelp() { return SHCCSTR("The input of this shard is ignored."); }
```
sinkingsugar: Standardize this too

shards/modules/core/wires.cpp
```diff
@@ -1399,9 +1477,9 @@ struct ParallelBase : public CapturingSpawners {
   DECL_ENUM_INFO(WaitUntil, WaitUntil, 'tryM');
 
   static inline Parameters _params{
-      {"Wire", SHCCSTR("The wire to spawn and try to run many times concurrently."), IntoWire::RunnableTypes},
-      {"Policy", SHCCSTR("The execution policy in terms of wires success."), {WaitUntilEnumInfo::Type}},
-      {"Threads", SHCCSTR("The number of cpu threads to use."), {CoreInfo::IntType}}};
+      {"Wire", SHCCSTR("The Wire to copy and schedule."), IntoWire::RunnableTypes},
```
sinkingsugar, line 1480: I noticed above too, not sure we need to capitalize those but if helps clarity why not

sinkingsugar (changes requested):

PR #1035, opened 2024-08-15
Finished documenting the shards in linalg.cpp

ShahFragcolor:
- Added/refined all the shards in linalg.cpp
- Removed the None type for the Flip Y parameter for Project and Unproject shards, gave it a default true instead if not set/
- Made Cross into a struct so that it can have specific help text.


sinkingsugar:
You some compilation errors, redeclaring methods or so

sinkingsugar (approved):
Need to fix tests, looks good, every statement verified thru ai?

PR #1041, opened 2024-08-22
Documented Http shards plus others

ShahFragcolor:
- Documented all Http shards
- documented ForEach and Map

### DIFF
--- a/shards/modules/crypto/src/hash.rs
+++ b/shards/modules/crypto/src/hash.rs
@@ -154,15 +154,15 @@ macro_rules! add_hasher {
 
       fn help(&mut self) -> OptionalString {
         if $name_str == "Hash.Keccak-256" {
-          OptionalString(shccstr!("This shard converts the input (string, bytes or a sequence of strings/bytes) into a 256-bit hash(of 32 bytes) using the Keccak-256 hashing algorithm."))
+          OptionalString(shccstr!("This shard outputs the 256-bit hash (32-bytes) of the input (string, bytes or a sequence of strings/bytes) using the Keccak-256 hashing algorithm."))
         } else if $name_str == "Hash.Keccak-512" {
-          OptionalString(shccstr!("This shard converts the input (string, bytes or a sequence of strings/bytes) into a 512-bit hash(of 64 bytes) using the Keccak-512 hashing algorithm."))
+          OptionalString(shccstr!("This shard outputs the 512-bit hash (64-bytes) of the input (string, bytes or a sequence of strings/bytes) using the Keccak-512 hashing algorithm."))
         } else if $name_str == "Hash.Sha3-256" {
-          OptionalString(shccstr!("This shard converts the input (string, bytes or a sequence of strings/bytes) into a 256-bit hash(of 32 bytes) using the SHA3-256 hashing algorithm."))
+          OptionalString(shccstr!("This shard outputs the 256-bit hash (32-bytes) of the input (string, bytes or a sequence of strings/bytes) using the SHA3-256 hashing algorithm."))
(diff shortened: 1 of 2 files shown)


shards/modules/core/core.cpp
```diff
@@ -889,6 +896,18 @@ struct ForEachShard {
 };
 
 struct Map {
+  static SHOptionalString help() {
```
sinkingsugar: I think we wanna do a quick mention of $0 and $1 here as well, short.

shards/modules/core/core.cpp
```diff
@@ -2687,6 +2729,20 @@ struct PassShard : public LambdaShard<unreachableActivation, CoreInfo::AnyType,
   static SHOptionalString outputHelp() { return DefaultHelpText::OutputHelpPass; }
 };
 
+struct HasherShard : public LambdaShard<hashActivation, CoreInfo::AnyType, CoreInfo::Int2Type> {
+  static SHOptionalString help() {
+    return SHCCSTR("This shard takes any input type, hashes them using the XXH128 hashing algorithm and returns their 128-bit hash value as a a sequence with 2 64-bit integer as elements.");
```
sinkingsugar: it's not a sequence, but just an Int2 to be correct

sinkingsugar (approved):
Looks good, just one minor change to do

shards/modules/crypto/src/hash.rs
```diff
@@ -149,6 +151,39 @@ macro_rules! add_hasher {
       fn name(&mut self) -> &str {
         $name_str
       }
+
+      fn help(&mut self) -> OptionalString {
+        if $name_str == "Hash.Keccak-256" {
+          OptionalString(shccstr!("This shard converts the input (string, bytes or a sequence of strings/bytes) into a 256-bit hash(of 32 bytes) using the Keccak-256 hashing algorithm."))
```
sinkingsugar: hmm , it does not really "convert", just outputs the hash of it

shards/modules/crypto/src/hash.rs
```diff
@@ -149,6 +151,39 @@ macro_rules! add_hasher {
...
+        }
+      }
+
+      fn inputHelp(&mut self) -> OptionalString {
+        OptionalString(shccstr!("The string, bytes or a sequence of strings/bytes to hash."))
+      }
+
+      fn outputHelp(&mut self) -> OptionalString {
+        if $name_str == "Hash.Keccak-256" {
+          OptionalString(shccstr!("Returns the 256-bit hash as a byte sequence (returns a sequence of bytes instead if the input was a sequence)."))
```
sinkingsugar: Something I forgot to mention that might be everywhere in our codebase sadly.. In Shards there is no real return, but rather output :)

shards/modules/json/json.cpp
```diff
@@ -617,6 +617,14 @@ struct ToJson {
     else
       return Var(_indent);
   }
+  static SHOptionalString help() { return SHCCSTR("This shard takes its input and converts it into a JSON string."); }
+
+  static SHOptionalString inputHelp() {
+    return SHCCSTR("If Pure is set to true, this shard acceps an input of anytype. If Pure is set to false, this shard only "
```
sinkingsugar: This is wrong, pure stands for `pure` json, sort of compatible with JS, you should check the output, non pure is instead shards flavored json, (with shards types info)

sinkingsugar (changes requested):

sinkingsugar:
@ShahFragcolor this is still pending for fixes

sinkingsugar (approved):

PR #1047, opened 2024-08-28
Candle

sinkingsugar:


shards/modules/candle/src/tokenizer.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+#[shards::shard_impl]
+impl Shard for MLTokenizer {
+  fn input_types(&mut self) -> &Types {
+    &STRING_TYPES
+  }
+
+  fn output_types(&mut self) -> &Types {
+    &NONE_TYPES
```
guusw, line 52: None types is wrong I think

shards/modules/candle/src/model.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+#[shards::shard_impl]
+impl Shard for ModelShard {
+  fn input_types(&mut self) -> &Types {
+    &BYTES_TYPES
+  }
+
+  fn output_types(&mut self) -> &Types {
+    &NONE_TYPES
```
guusw, line 96: Wrong output type?

guusw (changes requested):

PR #1048, opened 2024-08-29
Documented Shader Shards

ShahFragcolor:
Shader.LinearizedDepth
Shader.Literal
Shader.ReadBuffer
Shader.ReadGlobal
Shader.ReadInput
Shader.WithInput
Shader.WithTexture
Shader.WriteOutput
Shader.WriteGlobal

### DIFF
--- a/shards/modules/gfx/shader/core_shards.hpp
+++ b/shards/modules/gfx/shader/core_shards.hpp
@@ -422,14 +422,14 @@ template <typename TShard> struct Read : public IOBase {
 
 struct ShaderReadInput : public Read<blocks::ReadInput> {
   static SHOptionalString help() {
-    return SHCCSTR("This shard either reads the value of the specified vertex attribute supplied to the vertex or reads the interpolated value of that attribute, output from the vertex stage and provided to the pixel.");
+    return SHCCSTR("This shard reads the value of the shader input specified in the Name parameter.");
   }
   static SHOptionalString inputHelp() { return ShaderDefaultHelpText::InputHelpIgnored; }
-  static SHOptionalString outputHelp() { return SHCCSTR("The value of the  specified attribute."); }
+  static SHOptionalString outputHelp() { return SHCCSTR("The value of the  specified shader input."); }
 
   SHParametersInfo parameters() {
     static shards::Parameters params = {
-        {"Name", SHCCSTR("The name of the vertex attribute to read."), {shards::CoreInfo::StringType}},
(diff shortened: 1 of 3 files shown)


shards/modules/gfx/shader/core_shards.hpp
```diff
@@ -409,6 +420,34 @@ template <typename TShard> struct Read final : public IOBase {
   }
 };
 
+struct ShaderReadInput : public Read<blocks::ReadInput> {
+  static SHOptionalString help() {
+    return SHCCSTR("This shard either reads the value of the specified vertex attribute supplied to the vertex or reads the interpolated value of that attribute, output from the vertex stage and provided to the pixel.");
```
guusw: Might be useful to clearly state in what scenario what happens, e.g.
"In the vertex stage, this will read mesh vertex attributes"
"In the fragment stage this will read the interpolated output from the vertex stage"

shards/modules/gfx/shader/core_shards.hpp
```diff
@@ -565,6 +608,34 @@ template <typename TShard> struct Write : public IOBase {
   }
 };
 
+struct ShaderWriteOutput: public Write<blocks::WriteOutput> {
+  static SHOptionalString help() {
+    return SHCCSTR("This shard sets the value passed as input to the vertex attribute specified in the Name parameter and outputs it to the next stage or render target.");
```
guusw: In the fragment stage will write to one of the outputs attached to the render pass.
In the vertex stage will write or create a new output that can later be read from the fragment stage.

Above is more accurate. The shard doesn't set any vertex attributes

shards/modules/gfx/shader/core_shards.hpp
```diff
@@ -812,6 +887,20 @@ struct WithInput {
   static SHTypesInfo inputTypes() { return CoreInfo::NoneType; }
   static SHTypesInfo outputTypes() { return CoreInfo::NoneType; }
 
+  static SHOptionalString help() {
+    return SHCCSTR(
+        "This shard creates a conditional statement within a shader code. If the vertex attribute (or the interpolated value of the vertex attribute supplied to the pixel) specified in the Name parameter is being received at the stage that calls this shard, "
```
guusw: I would avoid terms like "interpolated value of the vertex attribute", vertex outputs can be anything they don't necessarily have to be based on vertex attributes, although often are

guusw: Just calling them vertex stage outputs seems enough to me, as long as it's clarified in detail elsewhere

guusw (changes requested):

PR #1062, opened 2024-09-15
feat(core): add support for C++ shared_ptr<void> objects

sinkingsugar:
Introduce SHVAR_FLAGS_CPP_SHARED_VOID_OBJECT flag for shared_ptr<void> objects. Update varAsObjectChecked, _destroyVarSlow, and _cloneVarSlow to handle the new flag. Refactor texture handling in gfx module.


sinkingsugar:
~~Does't work well, should just fix the core issue of using TObjectVar instead~~

sinkingsugar:
@guusw looking at it, I think this is the fastest way to fix the issue for now. You should check it. I think it just needs some fixes rust bindings side.
Solution is also not too bad actually conceptually as uses c++ standard features.

shards/core/foundation.hpp
```diff
@@ -1536,7 +1534,12 @@ template <typename T> T &varAsObjectChecked(const SHVar &var, const shards::Type
     throw std::runtime_error(fmt::format("Invalid object type id, expected: {} got: {}", type,
                                          Type::Object(var.payload.objectVendorId, var.payload.objectTypeId)));
   }
-  return *reinterpret_cast<T *>(var.payload.objectValue);
+  if ((var.flags & SHVAR_FLAGS_CPP_SHARED_VOID_OBJECT) == SHVAR_FLAGS_CPP_SHARED_VOID_OBJECT) {
+    auto void_sp = static_cast<std::shared_ptr<void> *>(var.payload.objectValue);
+    return *std::static_pointer_cast<T>(*void_sp);
```
sinkingsugar: This is wrong, we should detect that T could also be a shared ptr and extract only the inner type.

sinkingsugar:
it should work because:

Yes, in the specific scenario you've described, deleting the `std::shared_ptr<void>*` will correctly invoke the destructor of `SomeType`. However, it's essential to understand **why** this works and the underlying mechanics to ensure that it behaves as expected, especially since you're aiming for a low-level, optimized solution.

### Scenario Breakdown

1. **Original Pointer:**
   ```cpp
   std::shared_ptr<SomeType>* original_ptr = /* ... */;
   ```

2. **Casting to `std::shared_ptr<void>*`:**
   ```cpp
   auto void_sp = static_cast<std::shared_ptr<void>*>(original_ptr);
   ```

3. **Creating a New `std::shared_ptr<void>`:**
   ```cpp
   auto new_ptr = new std::shared_ptr<void>(*void_sp);
   ```

4. **Deleting the New Pointer:**
   ```cpp
   delete new_ptr;
   ```

### How It Works

#### 1. **Shared Control Block**

- **Shared Ownership:** Both `std::shared_ptr<SomeType>` and `std::shared_ptr<void>` share the same **control block**. This control block manages:
  - **Reference Count:** Keeps track of how many `shared_ptr` instances point to the managed object.
  - **Deleter:** Holds the mechanism to correctly destroy the managed object, irrespective of the `shared_ptr`'s type.

#### 2. **Type Erasure and Deleters**

- **Type-Erased Deleter:** The control block uses type erasure to store the deleter. This means that even if you have a `std::shared_ptr<void>`, the deleter still knows the actual type (`SomeType`) of the object it needs to destroy.
  
- **Correct Destructor Invocation:** When the reference count drops to zero, the deleter invokes the destructor of `SomeType`, ensuring proper resource cleanup.

#### 3. **Casting Implications**

- **Pointer Casting:** By casting `std::shared_ptr<SomeType>*` to `std::shared_ptr<void>*`, you're essentially treating the `shared_ptr<SomeType>` as a `shared_ptr<void>`. Since both types manage the same control block, this cast doesn't alter the underlying ownership semantics.

- **Copy Construction:** When you create a new `std::shared_ptr<void>` by copying `*void_sp`, you're increasing the reference count in the shared control block. This ensures that the managed object remains alive until all `shared_ptr` instances are destroyed.

#### 4. **Deletion Mechanics**

- **Deleting `new_ptr`:** When you delete `new_ptr` (which is a `std::shared_ptr<void>*`), the destructor of `std::shared_ptr<void>` is invoked. This destructor:
  - Decrements the reference count in the control block.
  - If the reference count reaches zero, the deleter is called, which correctly invokes `SomeType`'s destructor.

### Important Considerations

1. **Undefined Behavior Risks:**
   - **Strict Aliasing Rule:** Casting between unrelated pointer types (like `std::shared_ptr<SomeType>*` to `std::shared_ptr<void>*`) can lead to undefined behavior due to the C++ strict aliasing rules. This means that, from a standard compliance perspective, the behavior isn't guaranteed.
   
   - **Implementation-Dependent:** While many standard library implementations (like libstdc++ and libc++) have `std::shared_ptr` with similar internal layouts, this isn't mandated by the C++ standard. Relying on specific memory layouts can make your code non-portable and fragile.

2. **Safety vs. Optimization:**
   - **User's Stance:** You've indicated that safety isn't a primary concern and that you're optimizing for low-level performance. In such cases, if you **fully understand** the internal workings of `std::shared_ptr` in your target environment, and you're confident that the casting won't break assumptions, this approach can work.
   
   - **Potential Alternatives:** If possible, consider alternative designs that don't require such casting. For example, using `std::shared_ptr<void>` from the outset or leveraging polymorphism to manage different types without casting.

3. **Maintenance and Readability:**
   - **Code Clarity:** Casting between different `shared_ptr` types can make the code harder to understand and maintain. Future maintainers (or even yourself at a later time) might find it challenging to reason about ownership semantics.
   
   - **Documentation:** If you proceed with this approach, ensure thorough documentation explaining why the cast is safe and how ownership is managed.

### Conclusion

**In your specific case**, where you:

1. Cast a `std::shared_ptr<SomeType>*` to `std::shared_ptr<void>*`.
2. Create a new `std::shared_ptr<void>` by copying the casted pointer.
3. Delete the new `std::shared_ptr<void>*`.

**The `SomeType` destructor will indeed be called correctly** upon deletion, provided that:

- The `shared_ptr` implementations are compatible and share the same control block.
- The casting doesn't violate any underlying assumptions of the `shared_ptr`'s memory layout.

**However**, it's crucial to be aware of the potential undefined behavior and the reliance on specific implementation details. If you decide to proceed with this approach, ensure rigorous testing in your target environment to confirm that destructors are invoked as expected.

### Final Note

While this technique can be effective for low-level optimizations, always weigh the benefits against the potential risks and maintainability challenges. Whenever possible, prefer standard and well-defined mechanisms to manage resources in C++ to ensure portability and safety.

---

**References:**

- [C++ Standard on `std::shared_ptr`](https://en.cppreference.com/w/cpp/memory/shared_ptr)
- [C++ Strict Aliasing Rules](https://en.cppreference.com/w/cpp/language/reinterpret_cast#Strict_aliasing)

Yes—you can cast a `std::shared_ptr<SomeType>*` to a `std::shared_ptr<void>*`, create a new `std::shared_ptr<void>` from it, and then delete that new pointer. Due to `shared_ptr`'s type-erased control block, the `SomeType` destructor will be correctly called when the `std::shared_ptr<void>` is destroyed. However, this relies on implementation details and invokes undefined behavior according to the C++ standard, so while it may work in practice for low-level optimizations, it isn't guaranteed to be portable or safe in general.

PR #1070, opened 2024-10-14
refactor(channels): remove NoCopy flag and optimize channel ops

sinkingsugar:


shards/modules/channels/channels.hpp
```diff
@@ -25,55 +24,61 @@ struct DummyChannel : public ChannelShared {
...
+  }
+
+  bool try_unrecycle(OwnedVar &value) { return _recycle.try_pop(value); }
+
+  template <bool Recycle = true, typename Func> bool try_pop(Func &&func) {
+    OwnedVar value{};
+    if (_data.try_pop(value)) {
+      func(std::forward<OwnedVar>(value));
+      if constexpr (Recycle) {
+        _recycle.push(std::move(value));
```
guusw, line 58: Maybe add a limit , cause if you only use the moving push it might grow the recycle list endlessly

guusw (changes requested):

PR #1079, opened 2024-11-09
Add priority to SHFlow, replace _flowPool with flat_set

sinkingsugar:

### DIFF
--- a/shards/tests/network-ws.shs
+++ b/shards/tests/network-ws.shs
@@ -1,16 +1,11 @@
 ; Run the same network test except use Network.WS client & server
 @mesh(root)
 
-@wire(main-test {
-  "network.shs" | FS.Read = base-script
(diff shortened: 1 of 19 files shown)


shards/tests/network-ws.shs
```diff
@@ -1,12 +1,17 @@
 ; Run the same network test except use Network.WS client & server
-"network.shs" | FS.Read = base-script
-Regex.Replace("Network.Client\\(\".*\" [0-9]+" "Network.WS.Client(\"ws://127.0.0.1:9191\"")
-Regex.Replace("Network.Server\\(" "Network.WS.Server(") = new-script
-Shards.Read | Shards.Distill("runner") = runner
-WireRunner(runner)
-
 @mesh(root)
 
+@wire(main-test {
```
guusw: Why can't this run on the script root anymore, WireRunner doing something?

sinkingsugar: should be ok actually, will try revert

PR #1084, opened 2024-11-19
Refactor runtime composition context and type matching logic

sinkingsugar:
- Replace `inherited` map in `InternalCompositionContext` with a `LayeredMap` for better layer management.
- Adjust `validateConnection` and `internalComposeWire` to use `LayeredMap`.
- Add nested iterator to `LayeredMap` for traversing layers efficiently.
- Ensure proper handling of key insertions and deletions across layers.
- Refactor `matchTypes` call for improved readability.
- Remove unused constructor in `CompositionContext`.


shards/core/runtime.hpp
```diff
@@ -246,6 +246,222 @@ struct DefaultHelpText {
   static inline const SHOptionalString OutputHelpPass = SHCCSTR("Outputs the input value, passed through unchanged.");
 };
 
+// LayeredMap Class Definition
+template <typename K, typename V> class LayeredMap {
+  std::vector<std::unordered_map<K, V>> layers;
+  std::vector<bool> blocker_tags;
+  size_t first_blocker_index = std::numeric_limits<size_t>::max(); // Index of the first blocker
```
guusw: What does this do? shouldn't searching from back to front and stop on blocker tag work for all cases?

sinkingsugar: Ah eh, didn't check, blame AI 🤖🤣

sinkingsugar: I think its supposed to cache the index of the first layer but probably redundant, I'll double check

shards/core/runtime.cpp
```diff
@@ -1122,19 +1132,16 @@ SHComposeResult internalComposeWire(const std::vector<Shard *> &wire, SHInstance
...
           std::string_view sName(v.first.payload.stringValue, v.first.payload.stringLen);
-          ctx.inherited[sName] = *metadata;
+          ctx.sharedContext->inherited.insert(sName, *metadata);
         }
       }
     }
   }
 
   if (data.shared.elements) {
-#if SH_EXTENDED_COMPOSE_PROFILING
```
guusw, line 1132: Don't want to keep these for profiling?

shards/core/runtime.cpp
```diff
@@ -1025,17 +1034,18 @@ void validateConnection(InternalCompositionContext &ctx) {
     }
 
     if (!matching) {
-      std::string error;
```
guusw, line 1028: Can you change back to the shorter format (the fmt one)

guusw: It basically just doesn't log all the exposed variables and makes the error message a bit more concise

shards/core/type_info.cpp
```diff
@@ -131,20 +133,38 @@ SHTypeInfo deriveTypeInfo(const SHVar &value, const SHInstanceData &data, std::v
     if (expInfo) {
       auto sv = SHSTRVIEW(value);
       const auto varName = sv;
-      for (auto info : data.shared) {
-        std::string_view name(info.name);
-        if (name == sv) {
-          expInfo->push_back(SHExposedTypeInfo{.name = info.name, .exposedType = info.exposedType});
+      if (data.composeVersion == 2) {
```
guusw: Shouldn't this also be allowed on compose v1, since it should still have the privateContext, just filled shared, etc.

shards/modules/imaging/imaging.cpp
```diff
@@ -745,7 +747,7 @@ struct LoadImage {
       std::swap(_output, premultipliedOutput);
     }
 
-    _output.version = 0;
+    _output.version = input.version + 1;
```
guusw: Input is bytes here though

guusw: Should be 0 I think

shards/core/type_info.cpp
sinkingsugar: true

PR #1086, opened 2024-11-25
Add logging and ensure child wires are stopped properly

sinkingsugar:
- Add trace logs in SHMesh to log the stopping of scheduled and pending wires for better debugging.
- Ensure child wires in BaseRunner are stopped and set to nullptr to avoid potential issues with dangling references.

### DIFF
--- a/shards/modules/core/core.hpp
+++ b/shards/modules/core/core.hpp
@@ -2095,17 +2095,21 @@ struct Push : public SeqBase {
     shassert(data.privateContext && "Private context should be valid");
     auto inherited = reinterpret_cast<CompositionContext *>(data.privateContext);
 
-    const auto updateSeqInfo = [this, &data](const SHTypeInfo *existingSeqType = nullptr) {
+    // check if this type is already exposed
+    auto type = findExposedVariablePtr(inherited->inherited, _name);
+    auto global = _global || (type && type->global);
+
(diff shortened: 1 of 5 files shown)


shards/modules/core/core.hpp
```diff
@@ -1220,13 +1220,18 @@ struct Set : public SetUpdateBase {
...
+    }
+
     // bake exposed types
     if (_isTable) {
       // we are a table!
       _tableTypeInfo = updateTableType(_tableType, !_key.isVariable() ? &(SHVar &)_key : &Var::Empty, data.inputType,
                                        existingExposedType ? &existingExposedType->exposedType : nullptr);
 
-      if (_global) {
+      if (global) {
```
guusw, line 1234: Need to change Push as well, it just exposes based on _global (param)

guusw (changes requested):

shards/modules/core/wires.hpp
```diff
@@ -191,6 +191,11 @@ struct BaseRunner : public WireBase {
         }
       } else if (!wire->detached) {
         // but avoid stopping if detached and scheduled
+        if (wire->childWire) {
+          // stop the child wire if any
+          shards::stop(wire->childWire);
```
guusw: Maybe better inside stop, because this is "flow" state kind of.
Check other stop usages if it makes sense to have there

guusw (changes requested):

PR #1112, opened 2025-02-20
Table Take optimization

sinkingsugar:


guusw:
I think we had this in get already so good that you added it for take as well 😁

guusw (approved):

guusw (approved):

PR #1119, opened 2025-02-28
Guus/distill next

guusw:
Add tools for creating easy to reload bits of shards code, mainly inside Formabble's UI

### DIFF
--- a/lib/hot-reload.shs
+++ b/lib/hot-reload.shs
@@ -1,0 +1,147 @@
+@define(script-hot-reload false IgnoreRedefined: true)
+
+; Builds wrapper wires around imported hot-reloadable wires
+; usage:
+;   @hot-reload(@script-dir @namespace "my-script.shs" [wire-0 wire-1 ...])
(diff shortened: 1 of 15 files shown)


lib/hot-reload.shs
```diff
@@ -0,0 +1,147 @@
...
+      Maybe({
+          WireRunner((Get(_wt wire-name)))
+        } {
+          Msg((["Failed to run wire " wire-name] | String.Format) Level: LogLevel::Error Name: "reload")
+          true > _we
+        }
+      )
+    })
+  })
+  Regex.Replace("""\{"id":\{"name":"wrapper-id"\}\}""" wrapper-name-id-json) = body-shards-1
```
sinkingsugar, line 88: eh can't just use FromJson/ToJson here?

sinkingsugar (approved):
need to check the cr-sqlite module, not sure why is here

guusw:
Merged in 171186e256605ee671302179326e3d9fc3420f63

PR #1135, opened 2025-04-25
Guus/http-fixes

guusw:

### DIFF
--- a/shards/rust/src/core.rs
+++ b/shards/rust/src/core.rs
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /* Copyright © 2020 Fragcolor Pte. Ltd. */
+use crate::error::FastError;
 use crate::shard::legacy_shard_construct;
 use crate::shard::shard_construct;
 use crate::shard::LegacyShard;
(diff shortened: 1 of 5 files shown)


shards/rust/src/core.rs
```diff
@@ -555,15 +557,15 @@ unsafe extern "C" fn cancel_future_c_call<
...
+  F: Future<Output = Result<R, FastError>> + Send + 'static,
   R: Into<ClonedVar>,
   C: Fn(),
 >(
   _context: *mut SHContext,
   arg2: *mut c_void,
 ) -> SHVar {
   let data = arg2 as *mut FutureCallData<F, C>;
-  let f = std::ptr::read(&(*data).future); // Move out the future
+  let f = (*data).future.take().unwrap_unchecked(); // Move out the future
```
sinkingsugar, line 568: is this ok? cos before it would not `Drop`

sinkingsugar (approved):
Nice, no changes even needed cpp side!
Just check memory leaks/overflows

shards/rust/src/core.rs
guusw: asyncActivate blocks so I made the data owned by the caller on the stackm easier to manage.
Then pass it here by ref through arg2 void* and take future

PR #1136, opened 2025-04-29
Enhance SHMesh variable management

sinkingsugar:
Enhance SHMesh variable management by adding cleanup for dangling references and improving metadata handling. Introduce friend relationship with ExposedTypeInfo for better access control.

### DIFF
--- a/shards/core/foundation.hpp
+++ b/shards/core/foundation.hpp
@@ -1415,6 +1415,8 @@ struct ExposedTypeInfo {
   const SHExposedTypeInfo &operator*() const { return _innerInfo; }
 
 private:
+  friend struct ::SHMesh;
+
   void initFrom(const SHExposedTypeInfo &other) {
     _innerInfo.exposedType = cloneTypeInfo(other.exposedType);
     if (other.name) {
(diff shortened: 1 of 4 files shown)


shards/core/foundation.hpp
```diff
@@ -1415,6 +1411,8 @@ struct ExposedTypeInfo {
   const SHExposedTypeInfo &operator*() const { return _innerInfo; }
 
 private:
+  friend struct ::SHMesh;
```
guusw: Is this needed?

shards/core/runtime.hpp
```diff
@@ -733,16 +738,26 @@ struct SHMesh : public std::enable_shared_from_this<SHMesh> {
...
+        it->second.clean();
+      }
+      variablesMetadata.erase(it);
+    }
+    return vPtr;
   }
 
   constexpr auto &getVariables() { return variables; }
 
   void setMetadata(SHVar *var, SHExposedTypeInfo info) {
```
guusw, line 756: Needs a comment that it should be called AFTER getVariable/referenceVariable because otherwise it would cleanup the metadata

PR #1146, opened 2025-05-27
Guus/param var conv

guusw:
Cleanup and fix various warnings (update spdlog due to some)


guusw:
Merged through guus/test

PR #1148, opened 2025-06-05
Add Disconnect shard and implement peer disconnection logic

sinkingsugar:
- Introduced a new Disconnect shard to handle peer disconnection, supporting both KCP and WebSocket peers.
- Implemented disconnect methods in KCPPeer and WSPeer classes to update their state upon disconnection.
- Updated network tests to validate the disconnection functionality, ensuring proper logging and state management.


guusw (approved):
Looks good

PR #1150, opened 2025-06-13
added default tab for docking area

ShahFragcolor:
- added a default tab parameter for docking area to dictate the tab open
- added help text for docking area

### DIFF
--- a/shards/egui/src/containers/docking.rs
+++ b/shards/egui/src/containers/docking.rs
@@ -267,7 +267,7 @@ impl LegacyShard for DockArea {
   }
 
   fn outputHelp(&mut self) -> OptionalString {
-    OptionalString(shccstr!("This shard always outputs None."))
+    OptionalString(shccstr!("This shard outputs the input unmodified."))
   }
 
   fn parameters(&mut self) -> Option<&Parameters> {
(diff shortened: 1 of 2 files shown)


shards/egui/src/containers/docking.rs
```diff
@@ -241,23 +251,23 @@ impl LegacyShard for DockArea {
...
+    OptionalString(shccstr!("The input of this shard is ignored."))
   }
 
   fn outputTypes(&mut self) -> &Types {
     &ANY_TYPES
   }
 
   fn outputHelp(&mut self) -> OptionalString {
-    OptionalString(shccstr!("TODO."))
+    OptionalString(shccstr!("This shard always outputs None."))
```
guusw: Passthrough I think

guusw: We have a helper
```rs
fn outputHelp(&mut self) -> OptionalString {
    *HELP_OUTPUT_EQUAL_INPUT
  }
```

guusw (changes requested):
Other than that one comment: good!

shards/egui/src/containers/docking.rs
```diff
@@ -51,7 +54,13 @@ lazy_static! {
     cstr!("The UI contents containing tabs."),
     &SHARDS_OR_NONE_TYPES[..],
   )
-    .into(),];
+    .into(),
+  (
+    cstr!("DefaultTab"),
+    cstr!("The index of the tab to open by default. 0 being the right most tab."),
+    &INT_OR_NONE_TYPES_SLICE[..],
```
guusw: You can just put INT_TYPES (without none), since you're setting the default to 0

guusw (changes requested):

PR #1152, opened 2025-06-30
Tables refactor

sinkingsugar:
- Introduced a new member `fixedStructTable` in `SHTypeInfo` to optimize table handling.
- Updated `TableDecl` to mark tables as fixed struct tables when all keys are fixed values.
- Adjusted Rust type definitions to initialize `fixedStructTable` to false for consistency.

### DIFF
--- a/shards/modules/core/core.hpp
+++ b/shards/modules/core/core.hpp
@@ -21,6 +21,7 @@
 #include <cmath>
 #include <optional>
 #include <sstream>
+#include <numeric>
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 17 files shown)


shards/modules/core/core.hpp
```diff
@@ -2614,6 +2658,16 @@ struct TableDecl : public VariableBase {
     }
 
     if (!_isTable) {
```
guusw, line 2660: You should still check if type is `@type({none: something})`

guusw: or has none key

guusw (changes requested):

guusw (approved):

PR #1155, opened 2025-07-17
Add @include("path.shs" Many: true)

guusw:
Add a include many that allows inclusion of many files with the same name, to be used for formabble extensions


sinkingsugar:
Maybe should be `@include-many("...")` instead? or instead of `Many` `Repeating`? Not sure how this works btw? a bit confused haha

sinkingsugar:
Did we merge this with some other branch? @guusw

guusw:
> Did we merge this with some other branch? @guusw

No not yet I'm sure

PR #1163, opened 2025-08-10
Gio/tui

sinkingsugar:
This commit reformats the warmup method in the Separator structure to enhance code readability by adding line breaks. Additionally, it removes redundant element initialization in the activate method, streamlining the element management process within the TUI module.


shards/modules/tui/tui.cpp
```diff
@@ -0,0 +1,783 @@
...
+        // Update value variable with cleaned buffer content
+        auto tmp = Var(std::string_view(_buffer.data(), _buffer.size()));
+        cloneVar(_value.get(), tmp);
+
+        // Execute the OnEnter action
+        SHVar output{};
+        _onEnter.activate(context, input, output);
+
+        // Clear buffer for next input
+        _buffer.clear();
```
guusw: Can you just clear the value on the shards side. what if we don't want to clear on enter?

guusw (changes requested):

PR #1165, opened 2025-08-13
Guus/fbl api

guusw:
Changes for formabble C API stuff

### DIFF
--- a/include/shards/shards.h
+++ b/include/shards/shards.h
@@ -900,6 +900,8 @@ typedef void(__cdecl *SHUnregisterExitCallback)(SHString eventName);
 typedef struct SHVar *(__cdecl *SHReferenceVariable)(struct SHContext *context, struct SHStringWithLen name);
 typedef struct SHVar *(__cdecl *SHReferenceWireVariable)(SHWireRef wire, struct SHStringWithLen name);
 
+typedef bool(__cdecl *SHInit)();
+
 typedef struct SHExternalVariable {
   struct SHVar *var;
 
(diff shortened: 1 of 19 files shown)


shards/core/runtime.hpp
```diff
@@ -437,7 +466,7 @@ inline void sleep(std::chrono::nanoseconds duration) {
     while (nanosleep(&delay, &delay))
       (void)0;
 #endif
-  } else if (0.0) {
+  } else {
```
sinkingsugar, line 476: the idea here is that if you pass negative you SKIP sleeping totally tho

sinkingsugar (changes requested):
Some comment btw

include/shards/shards.h
```diff
@@ -1323,12 +1347,20 @@ typedef struct _SHCore {
...
+
+  SHSetupLogger setupLogger;
 
   //! ADD NEW FUNCTIONS AT BOTTOM OF THIS STRUCT
 } SHCore;
 
 typedef SHCore *(__cdecl *SHShardsInterface)(uint32_t abi_version);
 
+// Version of the shards interface that does not initialize itself automatically
+typedef SHCore *(__cdecl *SHShardsInterface1)(uint32_t abi_version);
```
sinkingsugar: why we need this anyway?

guusw: I added this so we can setup the log file before initialization. Since normally ShardsInterface does everything when you call it

shards/core/runtime.hpp
guusw: I added back that logic for the sleep(double) version to keep it compatible

include/shards/shards.h
sinkingsugar: call it `shardsInterfaceNoInit` maybe?

sinkingsugar: Maybe just breaking change is just fine tbh

guusw: Yeah lemme fix that

include/shards/shards.h
```diff
@@ -1126,6 +1128,28 @@ typedef struct SHStringWithLen(__cdecl *SHFastStringLoad)(uint64_t id);
...
+#define SHLOG_LEVEL_TRACE 0
+#define SHLOG_LEVEL_DEBUG 1
+#define SHLOG_LEVEL_INFO 2
+#define SHLOG_LEVEL_WARN 3
+#define SHLOG_LEVEL_ERR 4
+#define SHLOG_LEVEL_CRITICAL 5
+typedef void(__cdecl *SHLogCallback)(SHStringWithLen category, SHStringWithLen message, int level, void *userData);
+
+// set logToFile to log to the default log file
+#define SHLOG_LOG_TO_DEFAULT_FILE ((const char *)size_t(~0))
```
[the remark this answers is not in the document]
sinkingsugar: I agree with the AI

PR #1166, opened 2025-08-15
Remove CoreSwap references from CMakeLists.txt and inlined.cpp, and adjust related inline shard handling in core.hpp for improved performance and clarity.

sinkingsugar:


guusw (approved):
Looks good!

PR #1168, opened 2025-08-21
Enhance parameter composition by adding PARAM_COMPOSE_MERGE_REQUIRED in multiple shards

sinkingsugar:

### DIFF
--- a/shards/modules/gfx/renderer.cpp
+++ b/shards/modules/gfx/renderer.cpp
@@ -15,7 +15,8 @@ struct RendererShard {
   static const inline Type WindowType = WindowContext::Type;
 
   PARAM_PARAMVAR(_window, "Window", "The window to run the renderer on.", {CoreInfo::NoneType, Type::VariableOf(WindowType)});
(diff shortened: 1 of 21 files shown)


shards/modules/gfx/window.cpp
```diff
@@ -147,6 +147,7 @@ struct MainWindow final {
       SHInstanceData innerData = data;
       innerData.shared = SHExposedTypesInfo(_innerExposedVariables);
       _contents.compose(innerData);
+      PARAM_COMPOSE_MERGE_REQUIRED(_contents);
```
guusw: Window already has special handling with filtering below, so omit this

guusw (changes requested):

shards/modules/gfx/renderer.cpp
```diff
@@ -64,6 +64,8 @@ struct RendererShard {
     innerData.shared = SHExposedTypesInfo(_innerExposedVariables);
     _contents.compose(innerData);
 
+    PARAM_COMPOSE_MERGE_REQUIRED(_contents);
```
guusw: Same as window

guusw (changes requested):

PR #1170, opened 2025-08-23
Enhance mathematical operations with SIMD support

sinkingsugar:
- Updated CMakeLists.txt to include new SIMD-based math operations.
- Refactored inlined.cpp to handle new inline shards for various integer and float operations (addition, subtraction, multiplication, division, bitwise operations).
- Modified math.hpp to support SIMD operations in the composeV2 method and added activation methods for different data types.


guusw (approved):

shards/modules/core/CMakeLists.txt
```diff
@@ -68,14 +68,24 @@ add_shards_module(core SOURCES ${SOURCES}
...
   CoreIsLessEqual
-  MathAdd
-  MathSubtract
-  MathMultiply
-  MathDivide
-  MathXor
-  MathAnd
-  MathOr
-  MathMod
+  MathAddInt64x2
```
guusw, line 73: What about single x1 operations, should those be inlined for perf too?

guusw (changes requested):

PR #1207, opened 2025-11-10
Implement trampoline-based execution and migrate Sub shard

sinkingsugar:
## Summary

Implements trampoline-based execution optimization to eliminate recursion in control flow shards, addressing stack overflow and I-cache thrashing issues.

### Core Infrastructure
- Added `nestedShards` field to Shard struct for control flow communication
- Created `FlattenedShards` helper class for NULL-terminated shard arrays  
- Implemented stack-based execution loop in `shardsActivation` replacing recursive calls
- Added `preserveOutput` flag mechanism for shards that need to preserve output (like Sub)

### Sub Shard Migration
- Migrated Sub from recursive `_shards.activate()` to trampoline pattern
- Sub stores output in `_preservedOutput` field and sets `nestedShards`
- Runtime uses pointer stack (8 bytes/entry) + depth tracking for correct restoration
- Handles nested Subs and interleaving with other control flow correctly

### Performance Characteristics
- Eliminates ~150 bytes/recursion level stack overhead
- Avoids I-cache thrashing from ALWAYS_INLINE duplication
- Pointer stack for preserveOutput: 16 bytes/level (vs 32 for value copy)
- Most control flow shards (IfBlock, When, etc.) will have zero overhead when migrated

### Testing
- Basic Sub execution ✓
- Nested Sub execution (3+ levels) ✓  
- Interleaving with legacy recursive shards (When) ✓
- Backward compatibility with unmigrated shards ✓

### Next Steps
- Migrate IfBlock, When, Match, Repeat, ForRange to trampoline
- Performance benchmarks on deeply nested flows
- Measure actual stack usage reduction


sinkingsugar:
Turns out the simpler approach we had on devel was faster in proper benchmarks especially with not many nesting!

Issue #778, opened 2023-09-05
Semicolon ":" Take alias bug in template, formabble


ShahFragcolor:
**Describe the bug**
When a semicolon is used in a @template as an alias for Take, it will give an error

"Could not infer an output type and no Default value provided." where it does not recognise the variable that it is taking from

**To Reproduce**

**Expected behavior**
<img width="761" alt="image" src="https://github.com/fragcolor-xyz/shards/assets/99865253/2e08738a-eb6a-4239-8b09-996d7a168621">

With this example, the first take when explicit works fine, but the second take using the semicolon throws the error

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

sinkingsugar:
Actually didn't you fix those @guusw ? Or not for template?

sinkingsugar:
Also, another issue with `:` is that it assumes it's a string, so won't work with variables

sinkingsugar:
@claude can you check if this is still an issue a fix if so? - might have to mess with pest file.. make sure no performance regression pls... otherwise we can ignore this.. as it's minor probably?

PR #1219, opened 2025-11-20
fix: handle TakeTable/TakeSeq replacements in template parameters

sinkingsugar:
Fixes issue #778 where the colon ":" alias for Take didn't work correctly in template contexts when the parameter was another TakeTable expression.

## Changes
- Modified `create_take_table_chain` and `create_take_seq_chain` in `eval.rs` to handle TakeTable/TakeSeq replacements
- Added comprehensive test file `template-take-bug.shs`

Generated with [Claude Code](https://claude.ai/code)

---

> [!NOTE]
> Ensure `:` take syntax works in templates by resolving `TakeTable`/`TakeSeq` replacements and add tests covering nested/chained cases.
> 
> - **Lang/eval**:
>   - Update `create_take_table_chain` and `create_take_seq_chain` in `shards/lang/src/eval.rs` to resolve template parameter replacements of type `Value::TakeTable`/`Value::TakeSeq`.
>     - Apply base `Get` + `Take` chain from the replacement, then append additional keys/indices; insert `ExpectTable` where needed.
> - **Tests**:
>   - Add `shards/tests/template-take-bug.shs` with scenarios verifying `:` syntax inside templates, including nested, chained table/seq accesses.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5065e99d0c51934483e3cbbb6a80f971b962fea9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/shards/lang/src/eval.rs
+++ b/shards/lang/src/eval.rs
@@ -1551,6 +1551,48 @@ fn create_take_table_chain(
   line: LineInfo,
   e: &mut EvalEnv,
 ) -> Result<(), ShardsError> {
+  // Check if var_name has a replacement that is a TakeTable or TakeSeq
+  // This allows templates to use x:key syntax where x is a template parameter
+  // that could be another TakeTable like outer:inner
+  if let Some(replacement) = find_replacement(var_name, e) {
+    match replacement {
+      Value::TakeTable(base_name, base_path) => {
+        // First, get the base variable and apply the base path
+        add_get_shard(base_name, line, e)?;
+        add_expect_table_shard(line, e)?;
+        for base_part in base_path {
+          let s = Var::ephemeral_string(base_part.as_str());
+          add_take_shard(base_name, &s, line, e)?;
+        }
+        // Then apply the additional path from this TakeTable
+        for path_part in path {
+          let s = Var::ephemeral_string(path_part.as_str());
+          add_take_shard(var_name, &s, line, e)?;
+        }
+        return Ok(());
+      }
+      Value::TakeSeq(base_name, base_path) => {
+        // First, get the base variable and apply the base indices
+        add_get_shard(base_name, line, e)?;
+        for idx in base_path {
+          let idx_var: Var = (*idx).try_into().unwrap();
+          add_take_shard(base_name, &idx_var, line, e)?;
+        }
+        // Then apply the additional path from this TakeTable (treating as table keys)
+        add_expect_table_shard(line, e)?;
+        for path_part in path {
+          let s = Var::ephemeral_string(path_part.as_str());
+          add_take_shard(var_name, &s, line, e)?;
+        }
+        return Ok(());
+      }
+      _ => {
+        // For other replacements (including Identifier), fall through to default handling
+      }
+    }
+  }
+
+  // Default handling - var_name is either not a replacement or is an Identifier replacement
   add_get_shard(var_name, line, e)?;
   add_expect_table_shard(line, e)?;
   for path_part in path {
@@ -1566,6 +1608,46 @@ fn create_take_seq_chain(
   line: LineInfo,
   e: &mut EvalEnv,
 ) -> Result<(), ShardsError> {
+  // Check if var_name has a replacement that is a TakeTable or TakeSeq
+  // This allows templates to use x:0 syntax where x is a template parameter
+  if let Some(replacement) = find_replacement(var_name, e) {
+    match replacement {
+      Value::TakeTable(base_name, base_path) => {
+        // First, get the base variable and apply the base path
+        add_get_shard(base_name, line, e)?;
+        add_expect_table_shard(line, e)?;
+        for base_part in base_path {
+          let s = Var::ephemeral_string(base_part.as_str());
+          add_take_shard(base_name, &s, line, e)?;
+        }
+        // Then apply the additional indices from this TakeSeq
+        for path_part in path {
+          let idx: Var = (*path_part).try_into().unwrap();
+          add_take_shard(var_name, &idx, line, e)?;
+        }
+        return Ok(());
+      }
+      Value::TakeSeq(base_name, base_path) => {
+        // First, get the base variable and apply the base indices
+        add_get_shard(base_name, line, e)?;
+        for idx in base_path {
+          let idx_var: Var = (*idx).try_into().unwrap();
+          add_take_shard(base_name, &idx_var, line, e)?;
+        }
+        // Then apply the additional indices from this TakeSeq
+        for path_part in path {
+          let idx: Var = (*path_part).try_into().unwrap();
+          add_take_shard(var_name, &idx, line, e)?;
+        }
+        return Ok(());
+      }
+      _ => {
+        // For other replacements (including Identifier), fall through to default handling
+      }
+    }
+  }
+
+  // Default handling
   add_get_shard(var_name, line, e)?;
   for path_part in path {
     let idx = (*path_part).try_into().unwrap(); // read should have caught this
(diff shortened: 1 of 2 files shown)


sinkingsugar:
@claude build actually failing it seems, can you test it properly

sinkingsugar:
> ## Code Review - PR #1219
> ### Summary
> This PR fixes issue #778 where the colon `:` alias for `Take` didn't work correctly in template contexts when the parameter was another `TakeTable` or `TakeSeq` expression. The fix modifies `create_take_table_chain` and `create_take_seq_chain` in `shards/lang/src/eval.rs` to handle nested template parameter replacements.
> 
> ### ✅ Code Quality & Best Practices
> **Positive aspects:**
> 
> * **Clean separation of concerns**: The fix is well-isolated to the two functions that handle take chains
> * **Consistent pattern**: Both `create_take_table_chain` and `create_take_seq_chain` follow the same implementation pattern
> * **Proper borrow handling**: The code correctly clones data to drop immutable borrows before making mutable calls, resolving Rust E0502 errors
> * **Follows existing conventions**: Uses existing helper functions (`add_get_shard`, `add_take_shard`, `add_expect_table_shard`) consistently
> 
> **Suggestions:**
> 
> 1. **Potential logic error in mixed TakeTable/TakeSeq handling** (`eval.rs:1577-1593`):
>    When a `TakeSeq` replacement is followed by `TakeTable` path (string keys), the code inserts `add_expect_table_shard` _after_ processing all base indices. This assumes the final index access returns a table. However, this might fail if intermediate indices don't return tables. Consider whether `ExpectTable` should be inserted earlier or if type validation is needed.
>    ```rust
>    // Current code at eval.rs:1577-1593
>    Value::TakeSeq(base_name, base_path) => {
>      // ... process base indices ...
>      add_expect_table_shard(line, e)?;  // Only at the end
>      for path_part in path {
>        let s = Var::ephemeral_string(path_part.as_str());
>        add_take_shard(var_name, &s, line, e)?;
>      }
>    }
>    ```
> 2. **Similar issue in `create_take_seq_chain`** (`eval.rs:1621-1637`):
>    When a `TakeTable` replacement is followed by `TakeSeq` indices, there's no type checking to ensure the intermediate result is actually a sequence. This could lead to runtime errors with unclear error messages.
> 3. **Missing edge case: empty paths**:
>    While unlikely, if `base_path` is empty in a `TakeTable`/`TakeSeq` replacement, the code would just perform a `Get` followed by `ExpectTable`. Consider documenting this behavior or adding an assertion.
> 
> ### 🐛 Potential Bugs & Edge Cases
> **Issues identified:**
> 
> 1. **Identifier naming inconsistency** (`eval.rs:1568, 1573, 1591`):
>    When processing chained takes, the code uses `&base_name` for the base path takes but then switches to `var_name` for the additional path. This inconsistency could lead to incorrect error messages or debugging information. The `add_take_shard` function accepts a `name` parameter that's used for error reporting/debugging.
>    ```rust
>    // Line 1568: Uses base_name (correct for base path)
>    add_take_shard(&base_name, &s, line, e)?;
>    
>    // Line 1573: Uses var_name (should this also use base_name for the full chain?)
>    add_take_shard(var_name, &s, line, e)?;
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    **Impact**: Error messages might reference the template parameter name instead of the actual base variable name, making debugging harder.
> 2. **Type mismatch handling**:
>    There's no explicit error handling for type mismatches when chaining different access patterns (e.g., table followed by sequence index). The code assumes runtime type checking will catch this, but explicit validation would provide better error messages.
> 3. **Deep nesting concern**:
>    The code handles one level of replacement (`TakeTable` containing another `TakeTable`), but what about multiple levels? For example:
>    ```
>    @template(level3 [z] { z:key })
>    @template(level2 [y] { @level3(y:inner) })
>    @template(level1 [x] { @level2(x:middle) })
>    @level1(outer:first)
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    The current implementation only resolves one level. Is this intentional? If so, it should be documented.
> 
> ### 🧪 Test Coverage
> **Strengths:**
> 
> * Comprehensive test file `template-take-bug.shs` with 9 distinct test scenarios
> * Tests cover the main bug scenario (Test 6) and various edge cases
> * Good coverage of mixed table/sequence access patterns
> * Tests nested templates (Test 8)
> 
> **Missing test cases:**
> 
> 1. **Error cases**: No tests for expected failures (e.g., accessing non-existent keys, type mismatches)
> 2. **Empty collections**: What happens with `outer-table:nonexistent:nested`?
> 3. **Multiple chaining**: Template with multiple chained parameters: `@template(t [x y] { x:key1 y:key2 })`
> 4. **Deep nesting**: Three or more levels of template nesting with chained takes
> 5. **Mixed sequence/table in different orders**: Test both `table:key:0` and `seq:0:key` patterns more thoroughly
> 6. **Edge case with empty base_path**: Though unlikely to occur naturally, test the behavior when base_path is empty
> 
> **Test 7 has a misleading comment** (`template-take-bug.shs:43-49`):
> 
> ```
> ; Test 7: Template parameter is a TakeSeq, then use table key inside template
> @template(take-seq-then-table [t] {
>   ; After taking index 0, we get [1 2 3], which is a sequence
>   ; This test verifies TakeSeq replacement followed by index access
>   t:1 | Log("Seq then index") | Assert.Is(5)
> })
> @take-seq-then-table(outer-seq:1)
> ```
> 
> The comment says "After taking index 0" but the code uses `outer-seq:1` (index 1). The assertion expects 5, which is correct for `outer-seq[1][1] = [4,5,6][1] = 5`. The comment should be corrected.
> 
> ### ⚡ Performance Considerations
> **Analysis:**
> 
> * The fix adds a `find_replacement()` call at the start of each function, which is a hash table lookup (O(1))
> * The existing code already performed similar lookups in `add_get_shard` → `get_full_name` → `find_replacement_identifier`
> * **Net impact**: Negligible to slightly positive (we check earlier and potentially avoid unnecessary work in some cases)
> 
> **Observations:**
> 
> * Cloning `base_name` and `base_path` (strings and vectors) adds minor overhead, but this is necessary for Rust's borrow checker and only happens when replacements exist
> * The clones are small (identifier strings and path vectors are typically short)
> * No performance regressions expected
> 
> ### 🔒 Security Concerns
> **No security issues identified.**
> 
> This is a language runtime fix dealing with AST manipulation and code generation. The changes:
> 
> * Don't introduce new attack surfaces
> * Don't handle external input directly
> * Use safe Rust practices (no unsafe blocks)
> * Maintain existing type safety guarantees
> 
> ### 📝 Recommendations
> **Before merging:**
> 
> 1. **Fix the identifier naming issue**: Ensure consistent use of variable names in `add_take_shard` calls for better error messages
> 2. **Correct Test 7 comment**: Fix the misleading comment about "index 0"
> 3. **Add error test cases**: Include at least one test that verifies proper error messages for invalid access patterns
> 4. **Document multi-level nesting limitation**: If deep nesting isn't supported, document this limitation in comments
> 
> **Consider for follow-up:**
> 
> 1. Add type validation between different access patterns for better error messages
> 2. Consider whether `ExpectTable` should be inserted more granularly in mixed access scenarios
> 3. Evaluate whether deep template nesting should recursively resolve replacements
> 
> ### Overall Assessment
> This is a **solid fix** that addresses the reported issue correctly. The implementation follows existing patterns, handles the Rust borrow checker properly, and includes good test coverage. The main concerns are:
> 
> * Identifier naming consistency for error reporting
> * Potential type mismatch scenarios that could have unclear error messages
> * Test comment accuracy
> 
> With the suggested fixes, this PR should be ready to merge.

@claude I think the review is right, `add_expect_table_shard` can be used only within a TakeTable chain, but this PR adds it in many other places that could be dangerous? 
Check the review in general pls.

PR #1222, opened 2025-11-22
Allow pipes in function parameters without extra parentheses

sinkingsugar:
Adds PipeValue grammar rule to allow writing cleaner code like:
  Add(3 | Mul(4))
instead of requiring:
  Add((3 | Mul(4)))

Changes:
- Add PipeValue rule to shards.pest grammar
- Add process_pipe_value() function in read.rs
- Add test stubs for cargo test to work independently
- Add tests for pipe-in-params syntax with structural verification


sinkingsugar:
@claude can you check why CI is failing or run anyway linalg.shs to figure why this PR makes it fail?

sinkingsugar:
> **Claude finished @sinkingsugar's task** —— [View job](https://github.com/fragcolor-xyz/shards/actions/runs/19588470695)
> 
> ---
> ### CI Failure Analysis Complete
> 
> The CI failure has been identified and analyzed. Here's the summary:
> 
> ---
> 
> ### Root Cause
> 
> The PR introduces a **breaking change** in the grammar that changes how existing code is parsed.
> 
> **Failing Test:** `"Exp 1"` with code `"1 | Log | (2 | Log (3 | Log))"`
> 
> **The Error:**
> ```
> Parameter Prefix not accepting this kind of variable: Var: tQi6LhcBC6-QiHz7c (type: Var(Any), valid types: String)
> ShardsError { message: "Failed to set parameter 'Prefix', error: Set parameter validation failed", loc: <unknown>:1:16 }
> ```
> 
> ---
> 
> ### What Changed
> 
> The grammar change modifies `Param` from using `Value` to `PipeValue`:
> 
> | Old Grammar | New Grammar |
> |-------------|-------------|
> | `Param = { ParamName? ~ Value }` | `Param = { ParamName? ~ PipeValue }` |
> | Pipes not allowed in params | `PipeValue = { Block ~ ("\|" ~ Block)* }` |
> 
> This changes how `Log (3 | Log)` is parsed:
> 
> | Aspect | Old Parsing | New Parsing |
> |--------|-------------|-------------|
> | Structure | Two separate items: `Log` then `(3 \| Log)` | Single shard: `Log(3 \| Log)` |
> | Meaning | `Log` runs, then `(3 \| Log)` runs separately | `Log` runs with parameter `3 \| Log` |
> 
> The `Log` shard expects its `Prefix` parameter to be a `String`, but receives `Any` (result of evaluating the pipe), causing a type validation failure.
> 
> ---
> 
> ### Recommended Fix
> 
> **Update the test to use explicit pipe syntax:**
> 
> ```cpp
> // Old (breaks with new grammar - ambiguous):
> TEST_SUCCESS_CASE("Exp 1", "1 | Log | (2 | Log (3 | Log))");
> 
> // Fixed (explicit pipe separates Log from expression):
> TEST_SUCCESS_CASE("Exp 1", "1 | Log | (2 | Log | (3 | Log))");
> ```
> 
> The extra `|` after `Log` explicitly indicates that `(3 | Log)` is a separate pipeline element, not a parameter to `Log`.
> 
> ---
> 
> ### Impact Assessment
> 
> - Only the `"Exp 1"` test appears to be affected
> - Other tests like `"Exp 2"` and `"Exp 3"` work because they use fully parenthesized expressions
> - The breaking change affects the edge case pattern `Shard (expr)` where space precedes the paren
> - This is arguably a clarification of previously ambiguous syntax
> 
> ---
>  |

@claude makes sense, can you fix it and review formatter.rs as well cos that might be causing issues and most of all not sure its doing things in the right way. Our pr complicated it too

PR #1224, opened 2025-11-22
Add SFINAE-like auto-evaluation for shard parameters

sinkingsugar:
When a shard parameter expects a value but receives an executable (Shard/Func/Shards/Identifier resolving to shard), automatically wrap it in EvalExpr and retry. This executes the shard at compile-time and uses its output as the parameter value.

Before: Msg(NanoID) failed - Msg expects String, not Shard
After:  Msg(NanoID) works - auto-wraps like Msg((NanoID))

Key changes in eval.rs:
- can_eval_expr_wrap(): Check if value/result is wrappable
- wrap_in_eval_expr(): Wrap value in EvalExpr for compile-time eval
- set_shard_parameter(): Try fallback on parameter type mismatch

Added tests in hello.shs for:
- Identifier that resolves to shard (NanoID)
- Shards sequence as parameter
- Shard with parameters
- Chained shards in parameter

### DIFF
--- a/shards/core/runtime.cpp
+++ b/shards/core/runtime.cpp
@@ -1591,7 +1591,7 @@ bool validateSetParam(Shard *shard, int index, const SHVar &value) {
     }
   }
 #endif
-  SHLOG_ERROR("{}", err);
+
   return false;
 }
 
--- a/shards/lang/src/eval.rs
+++ b/shards/lang/src/eval.rs
@@ -2740,11 +2740,11 @@ fn create_shard_inner(
   Ok(s)
 }
 
-/// Check if a value can potentially be wrapped in an EvalExpr for auto-evaluation.
+/// Check if a value can potentially be wrapped in an Expr for auto-evaluation.
 /// This is used for SFINAE-like parameter setting: if setting a parameter fails,
-/// we try wrapping executable values in EvalExpr to get their result.
+/// we try wrapping executable values in Expr to get their result.
 /// Includes Identifier since identifiers can resolve to shard names (e.g., NanoID).
-fn can_eval_expr_wrap(value: &Value, var_value: &SVar) -> bool {
+fn can_expr_wrap(value: &Value, var_value: &SVar) -> bool {
   // Check if the result is a ShardRef - that's a strong indicator we should try wrapping
   if var_value.as_ref().valueType == SHType_ShardRef {
     return true;
@@ -2766,14 +2766,14 @@ fn can_eval_expr_wrap(value: &Value, var_value: &SVar) -> bool {
   )
 }
 
-/// Wrap a value in an EvalExpr for compile-time evaluation.
+/// Wrap a value in an Expr for evaluation.
 /// This allows `Msg(NanoID)` to work like `Msg((NanoID))` automatically.
 ///
 /// IMPORTANT: Only call this function for values that can_eval_expr_wrap() returns true for.
-fn wrap_in_eval_expr(value: &Value, line_info: LineInfo) -> Value {
+fn wrap_in_expr(value: &Value, line_info: LineInfo) -> Value {
   // For Shards sequences, the sequence IS the pipeline to execute directly
   if let Value::Shards(seq) = value {
-    return Value::EvalExpr(seq.clone());
+    return Value::Expr(seq.clone());
   }
 
   let content = match value {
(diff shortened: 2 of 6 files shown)


sinkingsugar:
> ## Code Review: Add SFINAE-like auto-evaluation for shard parameters
> ### Summary
> This PR adds a convenience feature that automatically wraps executable values (Shards, Funcs, Identifiers resolving to shards) in `EvalExpr` when passed as parameters, allowing `Msg(NanoID)` to work like `Msg((NanoID))`.
> 
> ### Code Quality
> **Positive aspects:**
> 
> * Clean separation of concerns with `can_eval_expr_wrap()` and `wrap_in_eval_expr()` helper functions
> * Good documentation comments explaining the SFINAE-like behavior
> * The fallback pattern is well-structured - try direct, then wrapped, then return original error
> 
> **Issues:**
> 
> 1. **Redundant `Value::Shards` match arm** in `wrap_in_eval_expr()` (eval.rs:2785):
>    ```rust
>    other => BlockContent::Const(other.clone()),
>    // Note: Value::Shards is handled above with early return
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    The `other` arm will never match `Value::Shards` because of the early return, but it adds confusion. Consider restructuring to avoid the catch-all when not all cases are actually possible.
> 2. **Inconsistency between AST and runtime checks** in `can_eval_expr_wrap()`:
>    
>    * The function checks both runtime `SVar` type (`SHType_ShardRef`) and AST `Value` type
>    * If an identifier resolves to something that's NOT a shard but the AST says `Value::Identifier`, it won't be caught by the match but could still return false from the ShardRef check
>    * This is probably fine for correctness but could lead to confusing behavior
> 3. **Missing `Value::Identifier` in AST match** in `can_eval_expr_wrap()`:
>    The function checks `Value::Shard`, `Value::Func`, `Value::Shards` in the AST but relies on runtime `SHType_ShardRef` to catch identifiers that resolve to shards. Consider adding `Value::Identifier` to the match for consistency:
>    ```rust
>    matches!(value, Value::Shard(_) | Value::Func(_) | Value::Shards(_) | Value::Identifier(_))
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Though this could cause unnecessary retry attempts for non-shard identifiers.
> 
> ### Potential Bugs
> 1. **Silent fallback may hide real errors**: When the wrapped evaluation also fails, the original error is returned. This is correct, but if the _wrapping_ succeeds but produces an incompatible type, the user gets the original "expected String, got Shard" error rather than a more helpful "auto-evaluation produced X but expected Y".
> 2. **No check for side effects**: If a shard has side effects, the auto-eval will execute those at compile-time. This is by design (same as explicit `((...))`) but worth considering if users might be surprised.
> 
> ### Performance Considerations
> * The fallback path creates a new `Value::EvalExpr` with cloned AST nodes - this is acceptable for the error path
> * `as_var()` is called twice on the fallback path - once for original, once for wrapped. This is unavoidable given the try-first approach
> * The overhead is only incurred when the first `set_parameter` fails, which is the minority case
> 
> ### Test Coverage
> The tests in `hello.shs` cover:
> 
> * ✅ Identifier resolving to shard (`NanoID`)
> * ✅ Shards sequence as parameter (`{1 | Math.Add(1) | ToString}`)
> * ✅ Shard with parameters (`Random.Name(3)`)
> * ✅ Multiple shards chained (`{["Result: " 42] | String.Format}`)
> 
> **Missing test cases:**
> 
> * ❌ No negative test verifying that non-shard identifiers (variables) still work correctly
> * ❌ No test for `Value::Func` case (only `Value::Shard` and `Value::Shards` tested)
> * ❌ No test verifying behavior when auto-eval fails (e.g., shard that requires runtime input)
> * ❌ No test for sequences with more than one element to verify only first element is checked
> 
> **Suggestion:** Add test cases that verify:
> 
> 1. A normal variable like `Msg(my-string-var)` still works
> 2. A failing auto-eval produces a sensible error message
> 3. An edge case like `Msg({Fail("test")})` behavior
> 
> ### Minor Issues
> 1. **Missing newline at end of file** - `hello.shs` ends without a trailing newline
> 2. **Comment says "outputs X"** but tests don't assert the actual output value (just that it doesn't error)
> 
> ### Overall Assessment
> This is a reasonable ergonomic improvement that makes the language more intuitive. The implementation is sound and follows existing patterns in the codebase. The main concerns are:
> 
> 1. Consider adding a few more test cases for edge cases and negative paths
> 2. The missing newline at EOF should be fixed
> 3. Consider whether the silent retry could mask errors in unexpected ways
> 
> **Verdict:** Approve with minor suggestions

@claude can you find out why Shards are Identifiers sometimes in AST? (Identifiers resolving to Shard) - I suspect it's some hack in the pest file? or read.rs? Can seek and report pls. Mentat focus.

PR #1272, opened 2026-05-08
feat(spreadsheet): add calamine-backed XLSX/XLS/ODS reader shards

sinkingsugar:
## Summary
- New \`shards-spreadsheet\` module: 4 shards (Load / Sheets / Read / ReadAll) reading Excel & OpenDocument workbooks into native Shards data via calamine 0.34. Pure Rust, in-memory bytes path supported — primary motivation is edge-talk iOS where xlsx files arrive without a writable filesystem path.
- Two output layouts via \`Layout\` enum: **Dense** (header-keyed \`Seq[Table]\`, LLM-friendly) and **Sparse** (\`Table[Int->Table[Int->Cell]]\`, preserves original cell coordinates, empty cells absent).
- Cell types map natively (Float/Int/Bool/String/None); DateTime cells render as ISO 8601 strings. Duplicate headers auto-suffix (\`Foo\`, \`Foo_2\`, …); empty headers become \`col_<idx>\`.
- **Bundled**: workaround for the Apple libc++ \`<complex>\` bug shipped in CLT 26.4.1 (\`__promote_t\` not included). Without it, native macOS Debug builds are broken on the latest OS. Force-include \`<__type_traits/promote.h>\` via \`CMAKE_CXX_FLAGS\` for Apple. Harmless once Apple fixes the SDK.
- **CLAUDE.md** gains a "Native macOS Build — Known Issues" section documenting the libc++ bug, the PCH-staleness pattern that masked it across Debug/Release configs, and the ASan self-deadlock in \`libclang_rt.asan_osx_dynamic\` on macOS 26.4.1.

## Shards
| Shard | Input | Output |
|---|---|---|
| \`Spreadsheet.Load\` | \`string\|bytes\` | \`Workbook\` (eager-parses every sheet) |
| \`Spreadsheet.Sheets\` | \`Workbook\` | \`[string]\` (preserves order) |
| \`Spreadsheet.Read\` | \`Workbook\` | \`Seq[Table]\` (Dense) or \`Table[Any]\` (Sparse) |
| \`Spreadsheet.ReadAll\` | \`string\|bytes\|Workbook\` | \`{sheetName: <as Read>}\` |

## Test plan
- [x] \`cargo check -p shards-spreadsheet\` clean
- [x] \`just build-rel\` succeeds end-to-end on macOS 26.4.1
- [x] \`build/Release/shards shards/tests/spreadsheet.shs\` passes — exercises Dense/Sparse, by-name and by-index sheet selection, header dedup, \`HasHeaders: false\`, and the bytes-input path
- [x] Fixture (\`shards/tests/data/spreadsheet_test.xlsx\`) regenerable from \`shards/modules/spreadsheet/build_fixture.py\` (uv shebang + openpyxl)
- [x] Registered in \`run-tests.sh\` so CI picks it up
- [ ] CI green


sinkingsugar:
### Code review

Found 2 issues:

1. Use-after-free in `range_to_dense` spillover-key branch. `Var::ephemeral_string` stores a raw pointer into the borrowed `str`; here `&format!("col_{}", i)` is a temporary whose `String` is dropped at the end of the `let key = ...;` statement, so by the time `insert_cell_into_table` runs on the next line, `key` holds a dangling pointer. The dense path with the literal-string header branch above (and the formula path at L229) correctly binds the `String` to a `let` first — only this else-branch does not. Triggered whenever a data row is wider than the header row.

https://github.com/fragcolor-xyz/shards/blob/ceb724ca6d69980ab46eeab18d50eb3ff18935ff/shards/modules/spreadsheet/src/lib.rs#L344-L356

2. New `shards/tests/spreadsheet.shs` is registered in `run-tests.sh` but not in any of the GitHub Actions workflows (`build-linux.yml`, `build-macos.yml`, `build-linux-valgrind.yml`). CI runs an explicit list of `./shards new ../shards/tests/*.shs` per workflow and does not invoke `run-tests.sh`, so the spreadsheet test will be silently skipped on every CI job. Precedent: the geo module's PR added the test to all three workflows alongside `run-tests.sh`.

https://github.com/fragcolor-xyz/shards/blob/ceb724ca6d69980ab46eeab18d50eb3ff18935ff/.github/workflows/build-linux.yml#L160-L170

sinkingsugar:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

sinkingsugar:
### Code review

Found 1 issue:

1. `compile-time-crc32` is declared in `Cargo.toml` but never used. The module uses the modern `#[shards::shard]` / `#[shards::shard_impl]` derive macros and does not implement `LegacyShard::hash()`, so there is no `crc32!()` call anywhere in `src/lib.rs`. Likely copy-paste residue from an older module template — safe to drop.

https://github.com/fragcolor-xyz/shards/blob/1d8b20b628d3ef3e2afe06568f944a6308304034/shards/modules/spreadsheet/Cargo.toml#L13-L17

---

_Edit: A previous version of this comment also flagged `SpreadsheetLoadShard::activate` for performing synchronous calamine file I/O directly in `activate()`. That was a false positive — the CLAUDE.md rule is about misusing async primitives (`block_on()`, `BlockingModel`, etc.), and calamine has no async API. Synchronous I/O is fine here; users can wrap calls in `Await(...)` if they want to push the work off the wire, same as with `CSV.Read` / `FS.Read`._

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1275, opened 2026-05-25
refactor(shell): extract shared shell-session crate for localshell + ssh

sinkingsugar:
## What

`localshell` and `ssh` were ~95% duplicated — both implement the same persistent interactive shell-session model (background reader thread filling a shared buffer, monotonic byte counter immune to truncation, sentinel `PS1` prompt detection) and differed only in their transport (PTY vs SSH channel).

This extracts a new internal crate **`shards-shell-common`** that owns everything transport-agnostic, behind a `ShellTransport` trait.

## Layout

**`shards-shell-common`** (new)
- `ShellTransport` trait — `read_into → ReadOutcome`, `write_all → TransportError`, `resize`, `close`, `poll_interval_ms`
- `ShellSession` state, unified `Drop`, and the generic reader thread
- all text helpers (prompt/interactive detection, ANSI cleaning, buffer truncation, VT100 rendering, escape interpretation)
- generic `Execute` / `SendInput` / `Read` / `Write` / `Resize` logic, exit-code capture, and the `Create`/`Connect` bootstrap helpers

**`localshell`** → `PtyTransport` (portable-pty) + `Create` + thin delegating shards
**`ssh`** → `SshTransport` (ssh2, preserving the session/channel lock ordering that avoids the reader/writer deadlock) + `Connect` + thin delegating shards

Each leaf wraps `ShellSession` in a local newtype so the `ref_counted_object_type_impl!` macro stays orphan-rule-legal. **Shard names, object type codes (`lshl`/`sshs`), registration entrypoints, and CMake wiring are unchanged** — purely an internal refactor.

## Verification

- `cargo check` clean on all three crates
- Full Release build links the rust union + `shards` binary
- `shards/tests/localshell.shs` passes end-to-end (Create, Execute, env/cwd persistence, interactive detection, SendInput, exit codes, PTY size/resize, raw Read/Write, `top` snapshot, auto-interrupt, incremental reads, truncation, StripAnsi, slow commands, heredoc). This exercises 100% of the shared code path.
- `SSH.Connect` registers + composes + enters its transport. The live SSH round-trip (`ssh-test.shs`) was **not** run — it needs the Docker SSH server on :2222 and Docker wasn't available in this environment. The SSH-specific transport (read/write/resize over a real channel) is therefore the only path not exercised end-to-end; its session logic is shared and validated via localshell.

## Behavioral note

One small change: previously a *poisoned writer/channel mutex* marked the session `Corrupted`; now `session_write` surfaces it as a plain error string without flipping session state (lock-poisoning of the buffer/interactive mutexes still marks `Corrupted` as before). Easy to restore the exact old behavior if preferred.

## Diffstat

The two leaf modules drop from ~1740/1771 lines to thin transports + shard scaffolding (**+1261 / −2979** across changed files), with the shared session logic now in one place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8450,6 +8450,7 @@ name = "shards-shell-common"
 version = "0.1.0"
 dependencies = [
  "lazy_static",
+ "regex",
  "shards",
  "strip-ansi-escapes",
  "vt100",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8350,8 +8350,7 @@ dependencies = [
  "lazy_static",
  "portable-pty",
  "shards",
- "strip-ansi-escapes",
- "vt100",
+ "shards-shell-common",
 ]
 
 [[package]]
@@ -8447,6 +8446,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "shards-shell-common"
+version = "0.1.0"
+dependencies = [
+ "lazy_static",
+ "shards",
+ "strip-ansi-escapes",
+ "vt100",
+]
+
+[[package]]
 name = "shards-spreadsheet"
 version = "0.1.0"
 dependencies = [
@@ -8463,10 +8472,9 @@ dependencies = [
  "compile-time-crc32",
  "lazy_static",
  "shards",
+ "shards-shell-common",
  "shellexpand",
  "ssh2",
- "strip-ansi-escapes",
- "vt100",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ members = [
     "shards/modules/candle",
     "shards/modules/pdf",
     "shards/modules/random",
+    "shards/modules/shell-common",
     "shards/modules/spreadsheet",
     "shards/modules/ssh",
     "shards/modules/svg",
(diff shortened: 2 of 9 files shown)


sinkingsugar:
Added a second commit on top of the refactor: **`LocalShell.WaitFor` / `SSH.WaitFor`** — an `expect`-style shard enabled by the now-shared session machinery.

It watches the session's output stream and blocks (cooperatively suspending the wire, like `Read`) until a regex matches, returning the matched text:

```shards
shell | LocalShell.Write("tail -f app.log\n" AppendNewline: false)
LocalShell.WaitFor(Session: shell Pattern: """Server started on :(\d+)""") >= line
```

- **Params:** `Session`, `Pattern` (regex), `Timeout` (ms, default 30000; `0` = wait indefinitely until the session ends), `StripAnsi` (default `true`)
- Returns the matched text; **errors on timeout** (catchable via `Maybe`)
- Shares the read cursor with `Read`/`Execute`, advancing past the match so subsequent reads continue after it
- Matching logic lives once in `shell-common::wait_for`

Verified via two new cases in `localshell.shs` (positive match against delayed background output + timeout). The positive test backgrounds a `READY_$((20+2))` echo whose *command line* can't match `READY_\d+`, so a match proves it waited for the real (delayed) output rather than the command echo. SSH's variant is identical and built/registered, but its live round-trip is unrun for lack of a Docker SSH server.

PR #1282, opened 2026-06-14
chore(deps): drop dead tracy-client v0.15.0 patch

sinkingsugar:
The root and union `[patch.crates-io]` pinned tracy-client to the old git `v0.15.0`, but the `profiling` fork and the gfx crate already depend on crates.io `0.17.4` — so the patch was never applied (cargo flagged it as unused). Removing it unifies tracy-client on 0.17.4 and stops fetching the stale git repo. No functional change.


sinkingsugar:
Superseded by #1286, which bundles this change into a single combined CI run. Closing to free the runner pool; the change is preserved and validated in #1286.

